PR #87, opened 2023-03-06
Man0s/event pipeline

losman0s:


programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -140,10 +140,35 @@ impl<'a> BankAccountWithPriceFeed<'a> {
...
+            .collect::<Result<Vec<_>>>()
+    }
+
     #[inline(always)]
-    pub fn calc_weighted_assets_and_liabilities_values(
+    pub fn calc_assets_and_liabilities_values(
         &self,
-        weight_type: WeightType,
+        weight_type: Option<WeightType>,
+        current_timestamp: i64,
```
jkbpvsc, line 171: Why did we add the `current_timestamp` to the fn signature?

its not being used anywhere as far as I can see?

jkbpvsc:
If we want to merge this can I ask you to rebase on main, and mark this PR ready for review.

programs/marginfi/src/state/marginfi_account.rs
losman0s: 2 reasons:

- **reason it's there**: passing the timestamp from the top rather than making the clock sysvar internal call was the fastest way to share accounting functionality with a client, as it was previously a runtime-only dependency.
- **reason it looks unused**: you wiped part of those changes during the merge you made for switchboard support + refactor, but did not remove those lingering bits. A side effect of those incoming commits was to enable client sharing as well (because we moved away from requiring the timestamp in `get_price()` altogether), so it looks like this timestamp was added for no reason if you just look at the overall diff.

I'll reiterate that it is not advised looking at this branch in its current state 😁

jkbpvsc:
Lets leave this cooking for a bit more before we merge. We need to update base program to 1.16 so we can build this on stable toolchain.

jkbpvsc:
@losman0s can this be closed?

jkbpvsc:
@

PR #94, opened 2023-04-12
feat: emissions logic

jkbpvsc:
- Add Ix for setting up LM for either borrowing or depositing
- Add mechanism to settle LM emissions before every balance change
- Add ix for withdrawing emissions from a balance

Emissions can be set by with an emission rate which is n of token emitted for a balance of 1M over a year.

TODO:
- Add tests


jkbpvsc:
Will review again in the morning and merge.

PR #118, opened 2023-07-07
Better bankruptcy handling

jkbpvsc:
- Adds an asset threshold for determining bankruptcy.
- Adds a disabled account flag, that prevents an user from preforming actions on an account.
- Better account bankruptcy checking (insolvent account with less than <10c of collateral)


losman0s (approved):
One comment typo to fix otherwise lgtm

PR #121, opened 2023-07-15
Flashloans

jkbpvsc:
## Flashloans

- Flashloans only  work for accounts that have a flag manually set for them by a group admin

### How they work
Flashloans work by deferring health checks for an account to the end of the flashloan. This is achieved with two new instructions. `flashloan_start` and `flashloan_end`. 

- `flashloan_start` - performs a number of security checks, set the `IN_FLASHLOAN_FLAG`
- `flashloan_end` - checks the account health, and clears the `IN_FLASHLOAN_FLAG`

### Security checklist 
1. `end_flashloan` ix index is after `start_flashloan` ix index
2. Ixs has an `end_flashloan` ix present
3. `end_flashloan` ix is for the marginfi program
4. `end_flashloan` ix is for the same marginfi account
5. Account is not disabled
6. Account is not already in a flashloan

### Considerations
- What happens if you flashloan multiple accounts in a single tx?
- Can you corrupt bank parameters by artificially changing the utilization rate in a bank.


jkbpvsc:
One consideration I'm having is if we should have all the borrows and repays happen in  the `fl_start` and `fl_end` ixs, to prevent any unexpected combinations of flashloan ixs and bank state updates.
Thoughts @losman0s?

PR #138, opened 2023-11-23
feat: lending pool add bank lite

jkbpvsc:
- Use a compact config struct for the add_bank ix data.
- Update the cli to use ui values for deposit and borrow limit
- Fix some clippy errors


losman0s (dismissed):

jkbpvsc:
Ok. I've decided to go a less invasive way and just make the existing ix data more compact by stripping it of useless padding while used in the tx data.

losman0s (approved):

PR #141, opened 2023-11-30
feat: wip program upgrade CI

losman0s:


programs/marginfi/src/state/marginfi_account.rs
jkbpvsc, line 1: New cli successfully decodes mainnet data?

jkbpvsc (commented):
looks good, I just want to triple check adding `repr(C)` doesn't affect the layout.
ik anchor was supposed to add it before.

If you can @losman0s double check the layout from the cli

jkbpvsc:
I compared the layout with old and new cli, and both are identical.

jkbpvsc (approved):

PR #153, opened 2023-12-26
Pyth Conf Interval Cap 

jkbpvsc:
Cap the confidence interval for pyth oracles to 5%.


jkbpvsc:
merging this as its deployed already, however would still appreciate a review @losman0s

PR #158, opened 2024-01-11
cli: add-bank-with-seed command

exdx:
Adds a new CLI command to exercise the functionality added in #154. Will be merged alongside core changes.

Issue #154, opened 2024-01-01
Single signer add bank ix


jkbpvsc:
We need an ix to add a bank with a single signer for ideal working with a MS.

PR #160, opened 2024-01-12
cli: add seed to add bank command

exdx:
Updates the `mfi group add bank` command to have an optional `seed` boolean argument. If `seed` is set then the command generates a PDA and uses it to sign the ix on behalf of the bank. 

Based off the work in https://github.com/mrgnlabs/marginfi-v2/issues/154. 

Generates a PDA from seeds and a random string iteratively. Uses the PDA to then sign on behalf of the bank in the `group_add_bank` command. 

Continuation of #158


jkbpvsc:
Seeing lint and tests failing

exdx:
CI is green -- I accidentally feature-gated imports in `utils.rs` which broke a lot of things lol. 

Ready for another look. I think conceptually this PR is ready I just want to make sure the business logic is correct. 

I do think using generics could slim this down further but I will look into that as a follow-up. Would be nice to abstract the different between `AddBank` and `AddBankWithSeed` further via a trait and only have one fn here instead of two. They are very similar conceptually.

clients/rust/marginfi-cli/src/processor/mod.rs
jkbpvsc, line 1: So a few nuances.

To make sure we only use a single singer for the multisig, before we were using 3 singers (authority -- ms, fee_payer, and bank).

For the new instruction to work, we must make sure we don't sign with the `bank_keypair` or the `fee_payer`.
Only the authority will sign in the multisig.

So in code:

1. We have to get this without an explicit fee payer
```
let mut signing_keypairs = config.get_signers(true -> false);
``

2. We have to only include the `bank_keypair` when not using the `--seed`,
so removing this line
```
 signing_keypairs.push(&bank_keypair);
```

And here
```
   let message = Message::new(&add_bank_ixs, Some(&config.authority()));
```

jkbpvsc (changes requested):

clients/rust/marginfi-cli/src/processor/mod.rs
exdx: Ok sounds good!

jkbpvsc (approved):

PR #163, opened 2024-01-16
feat: Add transfer account authority ix

exdx:
This PR adds the ability to transfer account authority to another account via a feature flag. This provides the ability to migrate accounts on the platform.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
@@ -8,26 +8,14 @@ use anchor_lang::prelude::*;
 pub fn set_account_transfer_authority(
     ctx: Context<MarginfiAccountSetAccountAuthority>,
 ) -> MarginfiResult {
-    // Gather accounts
-    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
-    let marginfi_account_key = ctx.accounts.marginfi_account.key();
-    let new_account_authority = ctx.accounts.new_authority.key();
-    let signer = ctx.accounts.signer.key();
-    let group = ctx.accounts.marginfi_group.key();
-    let old_account_authority = marginfi_account.authority;
+    // Ensure marginfi_account is dropped out of scope to not exceed stack frame limits
+    {
+        let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+        let new_account_authority = ctx.accounts.new_authority.key();
+        marginfi_account.set_new_account_authority_checked(new_account_authority)?;
+    }
 
-    marginfi_account.set_new_account_authority_checked(new_account_authority)?;
-
-    emit!(MarginfiAccountTransferAccountAuthorityEvent {
-        header: AccountEventHeader {
-            signer: Some(signer),
-            marginfi_account: marginfi_account_key,
-            marginfi_account_authority: new_account_authority,
-            marginfi_group: group,
-        },
-        old_account_authority,
-        new_account_authority,
-    });
+    // TODO: add back event (dropped for memory reasons)
 
     Ok(())
 }
--- a/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
@@ -37,6 +37,9 @@ pub struct MarginfiAccountSetAccountAuthority<'info> {
     #[account(mut)]
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
 
+    #[account(
+        address = marginfi_account.load()?.group,
+    )]
     pub marginfi_group: AccountInfo<'info>,
 
     #[account(
@@ -44,6 +47,7 @@ pub struct MarginfiAccountSetAccountAuthority<'info> {
     )]
     pub signer: Signer<'info>,
 
+    /// CHECK: The new account authority doesn't need explicit checks
     pub new_authority: AccountInfo<'info>,
 
     #[account(mut)]
--- a/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
@@ -7,7 +7,6 @@ use anchor_lang::prelude::*;
 
 pub fn set_account_transfer_authority(
     ctx: Context<MarginfiAccountSetAccountAuthority>,
-    new_account_authority: Pubkey,
 ) -> MarginfiResult {
     let MarginfiAccountSetAccountAuthority {
         marginfi_account: marginfi_account_loader,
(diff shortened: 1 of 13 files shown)


exdx:
Not ready for review.

exdx (commented):
Pretty small set of changes if this is all we need. 

I don't see a way to set flags from the CLI -- I'm wondering if we wanted to add that? Could be a separate PR

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -77,6 +78,24 @@ impl MarginfiAccount {
     pub fn get_flag(&self, flag: u64) -> bool {
         self.account_flags & flag != 0
     }
+
+    pub fn set_new_account_authority(
+        &mut self,
+        new_authority: Pubkey,
+    ) -> std::result::Result<(), MarginfiError> {
```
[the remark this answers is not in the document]
jkbpvsc: Signature is good, you can use a type alias we already have for this Result in `create::prelude`.
I would add to the name of the fn to `_checked` so its clear there are conditions to the change.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -77,6 +78,24 @@ impl MarginfiAccount {
...
+        // check if new account authority flag is set
+        if !self.get_flag(TRANSFER_AUTHORITY_ALLOWED_FLAG) || self.get_flag(DISABLED_FLAG) {
+            return Err(MarginfiError::IllegalAccountAuthorityTransfer);
```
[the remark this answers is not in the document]
jkbpvsc: Yes that is the assumption we are working with. 
Last thing your code is missing is the actual instruction itself, the account layout (and security declaration) and including the instruction in the `#[program]` in the `lib.rs`

jkbpvsc:
If you can also add a few integration tests that cover theses cases:
- Non owner tries to change the account authority -> fail
- Non whitelisted (no flag set) owner tries to change the authority -> fail
- Whitelisted owner tries to change the authority -> success

programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
```diff
@@ -0,0 +1,50 @@
...
+pub struct MarginfiAccountSetAccountAuthority<'info> {
+    #[account(mut)]
+    pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
+
+    #[account(
+        address = marginfi_account.load()?.authority,
+    )]
+    pub signer: Signer<'info>,
+
+    pub new_authority: AccountLoader<'info, MarginfiAccount>,
```
jkbpvsc, line 46: The `new_authority` is not a `MarginfiAccount` account, its just an address which we will check as the signer moving forward.

Meaning we should use 
```
 pub new_authority: AccountInfo<'info>,
```

programs/marginfi/src/instructions/marginfi_account/transfer_authority.rs
```diff
@@ -0,0 +1,50 @@
...
+    state::marginfi_account::MarginfiAccount,
+};
+use anchor_lang::prelude::*;
+
+pub fn set_account_transfer_authority(
+    ctx: Context<MarginfiAccountSetAccountAuthority>,
+) -> MarginfiResult {
+    // Gather accounts
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    let new_account_authority = ctx.accounts.new_authority.load()?;
```
jkbpvsc, line 13: No need to load this as if its an account loader, bc we are only interested in the address of the account info

programs/marginfi/tests/marginfi_account.rs
```diff
@@ -2410,3 +2411,195 @@ async fn lending_account_close_balance() -> anyhow::Result<()> {
...
+    let transfer_account_authority_ix = Instruction {
+        program_id: marginfi::id(),
+        accounts: marginfi::instructions::MarginfiAccountSetAccountAuthority {
+            marginfi_account: todo!(),
+            signer: todo!(),
+            new_authority: todo!(),
+            fee_payer: todo!(),
+        }
+        .to_account_metas(None),
+        data: vec![],
```
jkbpvsc, line 2475: Can you add a helper function to the MarginfiAccountFixture `try_transfer_account_authority` so we are not duplicating the logic across tests

jkbpvsc (changes requested):

jkbpvsc:
Also seems tests have failed

programs/marginfi/tests/marginfi_account.rs
```diff
@@ -2410,3 +2411,195 @@ async fn lending_account_close_balance() -> anyhow::Result<()> {
...
+    assert_eq!(marginfi_account.authority, test_f.payer());
+    // transfer flag not set on the new account
+    assert!(!marginfi_account.get_flag(TRANSFER_AUTHORITY_ALLOWED_FLAG));
+    let new_account_authority_pk: [u8; 32] = [0; 32];
+
+    // // attempt authority transfer
+    // TODO: fix types
+    let transfer_account_authority_ix = Instruction {
+        program_id: marginfi::id(),
+        accounts: marginfi::instructions::MarginfiAccountSetAccountAuthority {
```
jkbpvsc, line 2468: Shoule be marginfi::accounts::MarginfiAccountSetAccountAuthority

programs/marginfi/tests/marginfi_account.rs
jkbpvsc, line 2475: should be marginfi::instruction::MarginfiAccountSetAccountAuthority

exdx:
Implemented the new functionality and it looks really good now. Running into some CI issues though:
```
error[E0080]: evaluation of constant value failed
  --> programs/marginfi/src/state/marginfi_account.rs:26:1
   |
26 | assert_struct_size!(MarginfiAccount, 2304);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ attempt to compute `0_usize - 1_usize`, which would overflow
   |
   = note: this error originates in the macro `const_assert` which comes from the expansion of the macro `assert_struct_size` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I get this error when building on an M1 mac but they went away on an x86 box. I don't think any struct sizes have changed? Trying to see if I changed something

exdx:
The issue is that the `MarginfiAccountSetAccountAuthority` accounts type is too large for the stack. It's an anchor issue. Looking into how to slim down the size of the struct by boxing accounts

jkbpvsc (changes requested):

test-utils/src/marginfi_account.rs
```diff
@@ -625,4 +625,39 @@ impl MarginfiAccountFixture {
     pub fn get_size() -> usize {
         mem::size_of::<MarginfiAccount>() + 8
     }
+
+    /// Use the client to send the transfer ix authority transaction
+    pub async fn try_transfer_account_authority(
+        &self,
```
jkbpvsc, line 631: `new_authority` should probably be a fn argument here

exdx: I agree. I seem to break the test every time I add this though lol. 😅

exdx:
Seems like the CI problems here are across open PRs and not related to this PR's functionality.

exdx:
The fuzz test is broken across CI currently -- not related to the changes made in this PR.

jkbpvsc (approved):

PR #166, opened 2024-01-24
ci: Downgrade Rust version to previous version

exdx:
See https://github.com/mrgnlabs/marginfi-v2/actions/runs/7645847144/job/20833360259 for an example of a CI run that is failing. This PR introduces a fix.


jkbpvsc:
Nice! Can you make a slight change and keep the `RUSTUP_TOOLCHAIN: nightly` just for the fuzz tests

PR #187, opened 2024-03-30
Man0s/event indexing

losman0s:


jkbpvsc:
@losman0s is this still relevant or can we close?

PR #200, opened 2024-04-18
More flexible risk engine

jkbpvsc:
- ignore stale oracles in asset balances when evaluating initialization health

This allows users to withdraw or borrow even when one of their balances might have a stale oracle.

For example if a user is depositing $100 foo and $1 bar, and wants to withdraw $100 foo.

Right now if the bar oracle is stale, the user won't be able to do anything, but with this change the user can withdraw the $100 foo.

### DIFF
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -215,7 +215,7 @@ impl<'a, 'b> BankAccountWithPriceFeed<'a, 'b> {
     /// 3. Initial requirement is discounted by the initial discount, if enabled and the usd limit is exceeded.
     /// 4. Assets are only calculated for collateral risk tier.
     /// 5. Oracle errors are ignored for deposits in isolated risk tier.
-    pub fn calc_weighted_assets_and_liabilities_values(
+    fn calc_weighted_assets_and_liabilities_values(
         &self,
         requirement_type: RequirementType,
     ) -> MarginfiResult<(I80F48, I80F48)> {
@@ -246,7 +246,18 @@ impl<'a, 'b> BankAccountWithPriceFeed<'a, 'b> {
     ) -> MarginfiResult<I80F48> {
         match bank.config.risk_tier {
             RiskTier::Collateral => {
-                let price_feed = self.try_get_price_feed()?;
+                let price_feed = self.try_get_price_feed();
+
+                if matches!(
+                    (&price_feed, requirement_type),
+                    (&Err(PriceFeedError::StaleOracle), RequirementType::Initial)
+                ) {
+                    debug!("Skipping stale oracle");
+                    return Ok(I80F48::ZERO);
+                }
+
+                let price_feed = price_feed?;
+
                 let mut asset_weight = bank
                     .config
                     .get_weight(requirement_type, BalanceSide::Assets);
@@ -301,16 +312,28 @@ impl<'a, 'b> BankAccountWithPriceFeed<'a, 'b> {
         )
     }
 
-    fn try_get_price_feed(&self) -> MarginfiResult<&OraclePriceFeedAdapter> {
+    fn try_get_price_feed(&self) -> std::result::Result<&OraclePriceFeedAdapter, PriceFeedError> {
         match self.price_feed.as_ref() {
             Ok(a) => Ok(a),
-            Err(_) => Err(MarginfiError::StaleOracle)?,
+            Err(_) => Err(PriceFeedError::StaleOracle),
         }
     }
 
     #[inline]
     pub fn is_empty(&self, side: BalanceSide) -> bool {
         self.balance.is_empty(side)
+    }
+}
+
+enum PriceFeedError {
+    StaleOracle,
+}
+
+impl From<PriceFeedError> for Error {
+    fn from(value: PriceFeedError) -> Self {
+        match value {
+            PriceFeedError::StaleOracle => error!(MarginfiError::StaleOracle),
+        }
     }
 }
 
@@ -411,9 +434,9 @@ impl<'a, 'b> RiskEngine<'a, 'b> {
         })
     }
 
-    /// Checks account is healty after performing actions that increase risk (removing liquidity).
+    /// Checks account is healthy after performing actions that increase risk (removing liquidity).
     ///
-    /// `IN_FLASHLOAN_FLAG` behaviour.
+    /// `IN_FLASHLOAN_FLAG` behavior.
     /// - Health check is skipped.
     /// - `remaining_ais` can be an empty vec.
     pub fn check_account_init_health(


losman0s (changes requested):

programs/marginfi/tests/risk_engine_flexible_oracle_checks.rs
```diff
@@ -0,0 +1,419 @@
...
+    },
+};
+use marginfi::{
+    prelude::MarginfiError,
+    state::marginfi_group::{Bank, BankConfig, BankConfigOpt, BankVaultType, GroupConfig},
+};
+use solana_program_test::tokio;
+
+#[tokio::test]
+/// User deposits $5000 SOLE and $500 USDC, borrowing $990 SOL should fail due to stale oracle
```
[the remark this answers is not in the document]
jkbpvsc: I want to use $ denominated so its immediately clear how much user has on account.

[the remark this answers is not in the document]
losman0s: Updated, cause original has wrong amounts anyway

programs/marginfi/src/errors.rs
```diff
@@ -23,8 +23,8 @@ pub enum MarginfiError {
     MissingBankAccount,
     #[msg("Invalid Bank account")] // 6009
     InvalidBankAccount,
-    #[msg("Bad account health")] // 6010
-    BadAccountHealth,
+    #[msg("RiskEngine rejected due to either bad health or stale oracles")] // 6010
+    RiskEngineInitRejected,
```
exdx, line 27: Errors generally should refer to 1 condition if possible, here this error represents either bad health or stale oracles. Is it possible to make 2 new errors here and precisely identify the cause of the error?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -246,7 +246,18 @@ impl<'a, 'b> BankAccountWithPriceFeed<'a, 'b> {
...
             RiskTier::Collateral => {
-                let price_feed = self.try_get_price_feed()?;
+                let price_feed = self.try_get_price_feed();
+
+                if matches!(
+                    (&price_feed, requirement_type),
+                    (&Err(PriceFeedError::StaleOracle), RequirementType::Initial)
+                ) {
+                    debug!("Skipping stale oracle");
+                    return Ok(I80F48::ZERO);
```
exdx, line 256: I feel like instead of returning 0 we could use the std::Result<T,E> and return an error in this case and handle it one layer up. A stale oracle feels more like an error

test-utils/src/marginfi_group.rs
```diff
@@ -300,6 +301,7 @@ impl MarginfiGroupFixture {
         &self,
         bank: &BankFixture,
         marginfi_account: &MarginfiAccountFixture,
+        nonce: u64,
```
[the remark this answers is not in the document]
jkbpvsc: the test validator will dedup txs, so the second tx I send gets ignored, this way I change the tx sig so its doesn't get ignored.

programs/marginfi/src/state/marginfi_account.rs
jkbpvsc: The purpose of this PR is this very behavior.

programs/marginfi/src/errors.rs
jkbpvsc: The behavior of ignoring stale oracles makes it impossible to determine wether its one or the other conclusively.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -314,6 +325,18 @@ impl<'a, 'b> BankAccountWithPriceFeed<'a, 'b> {
     }
 }
 
+enum PriceFeedError {
```
cavemanloverboy, line 328: Why is this not just another error variant in `MarginfiError`? Especially given that it's only one variant

cavemanloverboy: (I suspect this may make handling this error easier)

programs/marginfi/src/state/price.rs
```diff
@@ -151,6 +151,12 @@ pub struct PythEmaPriceFeed {
 impl PythEmaPriceFeed {
     pub fn load_checked(ai: &AccountInfo, current_time: i64, max_age: u64) -> MarginfiResult<Self> {
         let price_feed = load_pyth_price_feed(ai)?;
+
+        debug!(
+            "Oracle age: {}s",
+            (price_feed.get_price_unchecked().publish_time - current_time).abs()
```
cavemanloverboy, line 157: prob shouldn't use `abs` here.

programs/marginfi/src/errors.rs
cavemanloverboy: In case there is ever any retroactive analysis of transactions, it may be worth considering leaving the old error as dead code and introducing this error with a new code. There are pros and cons to this approach -- off-chain and cpi clients may break if handling this error explicitly.

programs/marginfi/tests/bank_ignore_stale_isolated_banks.rs
```diff
@@ -54,7 +55,56 @@ async fn non_isolated_stale_should_error() -> anyhow::Result<()> {
...
+    borrower_mfi_account_f
+        .try_bank_deposit(borrower_token_account_f_sol_eq.key, sol_eq_bank, 100)
+        .await?;
+
+    // Borrow SOL
+    let res = borrower_mfi_account_f
+        .try_bank_borrow(borrower_token_account_f_sol.key, sol_bank, 99)
+        .await;
+
+    assert!(res.is_ok());
```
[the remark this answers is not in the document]
cavemanloverboy: yea I don't see any oracle ages set

programs/marginfi/tests/bank_variable_oracle_staleness.rs
```diff
@@ -41,11 +44,11 @@ async fn bank_oracle_staleness_test() -> anyhow::Result<()> {
...
 
     borrower_mfi_account_f
-        .try_bank_deposit(borrower_token_account_f_usdc.key, usdc_bank, 1_000)
+        .try_bank_deposit(borrower_token_account_f_usdc.key, usdc_bank, 500)
         .await?;
 
     borrower_mfi_account_f
-        .try_bank_deposit(borrower_token_account_f_sol_eq.key, sol_eq_bank, 1_000)
+        .try_bank_deposit(borrower_token_account_f_sol_eq.key, sol_eq_bank, 500)
         .await?;
```
cavemanloverboy, line 52: Is there a reason why these changed to 500 from 1000 that would change test behavior? If no reason, ignore this comment.

cavemanloverboy (approved):
Don't see anything wrong with implementation but had a few clarifying qns/comments.

programs/marginfi/src/errors.rs
jkbpvsc: Any client should actually deserialize this into the same error as before, as it still encodes it into the same enum representation

programs/marginfi/src/state/marginfi_account.rs
jkbpvsc: Yes exactly, also the marginfi error gets converted into an AnchorError which is super bloated

programs/marginfi/tests/bank_variable_oracle_staleness.rs
losman0s: So I believe the intent of this test is to check that a stale oracle can a stale bank deposit (next to a non-stale one) can make the difference between having enough collat for a borrow/withdraw and not.

So since here the test borrow is worth $990 (test suite sol price of $10), having a USDC deposit worth $1000 would alone be enough to back that borrow, and would fail the first error assertion. Updated amounts make each deposit bank account for ~ half of the required collat for that SOL borrow.

programs/marginfi/tests/bank_ignore_stale_isolated_banks.rs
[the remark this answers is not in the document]
losman0s: Added the bits to make the USDC oracle stale. This should now correctly test that a stale bank has no impact on the user's ability to borrow, as long as the non-stale collateral is sufficient for it.

cavemanloverboy (approved):
most recent PR looks good to me

cavemanloverboy (approved):

losman0s (approved):

PR #205, opened 2024-05-01
Fees and Insurance Fund Operational Ixs

jkbpvsc:
Add instructions to remove liquidity from the fee vault and the insurance vault.
Updated cli to be able to create the ixs.

### DIFF
--- a/clients/rust/marginfi-cli/src/processor/admin.rs
+++ b/clients/rust/marginfi-cli/src/processor/admin.rs
@@ -1,0 +1,155 @@
+use anchor_client::anchor_lang::prelude::*;
+use anchor_client::anchor_lang::InstructionData;
+use anchor_spl::associated_token;
+use anyhow::Result;
+use marginfi::{
+    bank_authority_seed,
+    state::marginfi_group::{Bank, BankVaultType},
+};
+use solana_sdk::{
+    instruction::Instruction, message::Message, pubkey::Pubkey, transaction::Transaction,
+};
+
+use crate::{
+    config::Config,
+    utils::{process_transaction, ui_to_native},
+};
+
+pub fn process_collect_fees(config: Config, bank_pk: Pubkey) -> Result<()> {
+    let bank = config.mfi_program.account::<Bank>(bank_pk)?;
+    let rpc_client = config.mfi_program.rpc();
+
+    let (liquidity_vault_authority, _) = Pubkey::find_program_address(
+        bank_authority_seed!(BankVaultType::Liquidity, bank_pk),
+        &marginfi::id(),
+    );
+
+    let ix = Instruction {
+        program_id: marginfi::id(),
+        accounts: marginfi::accounts::LendingPoolCollectBankFees {
+            marginfi_group: bank.group,
+            bank: bank_pk,
+            fee_vault: bank.fee_vault,
+            token_program: spl_token::id(),
+            liquidity_vault_authority,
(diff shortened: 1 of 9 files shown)


programs/marginfi/tests/withdraw_fees.rs
```diff
@@ -0,0 +1,58 @@
+use fixtures::test::{BankMint, TestFixture, TestSettings};
+use marginfi::state::marginfi_group::GroupConfig;
+use solana_program_test::tokio;
+use solana_sdk::pubkey::Pubkey;
+
+#[tokio::test]
+async fn marginfi_group_add_bank_success() -> anyhow::Result<()> {
```
losman0s, line 7:
```suggestion
async fn marginfi_group_withdraw_fees_and_insurance() -> anyhow::Result<()> {
```

losman0s (changes requested):
CI does not pass rn

losman0s (approved):

clients/rust/marginfi-cli/src/processor/admin.rs
```diff
@@ -0,0 +1,167 @@
...
+    let recent_blockhash = rpc_client.get_latest_blockhash().unwrap();
+    let signing_keypairs = config.get_signers(false);
+
+    let message = Message::new(&[ix], Some(&config.authority()));
+    let mut transaction = Transaction::new_unsigned(message);
+    transaction.partial_sign(&signing_keypairs, recent_blockhash);
+
+    match process_transaction(&transaction, &rpc_client, config.get_tx_mode()) {
+        Ok(sig) => println!("Tx succeded (sig: {})", sig),
+        Err(err) => println!("Error:\n{:#?}", err),
```
exdx, line 51: We could `error!()` here and return a non-zero exit code?

programs/marginfi/tests/withdraw_fees.rs
```diff
@@ -0,0 +1,58 @@
...
+async fn marginfi_group_withdraw_fees_and_insurance() -> anyhow::Result<()> {
+    // Setup test executor with non-admin payer
+    let mut test_f = TestFixture::new(Some(TestSettings::all_banks_payer_not_admin())).await;
+
+    // Mint 1000 USDC to the insurance vault
+    let bank_f = test_f.banks.get(&BankMint::USDC).unwrap();
+    let bank = bank_f.load().await;
+    test_f.usdc_mint.mint_to(&bank.insurance_vault, 1000).await;
+
+    // Create a receiving account and try to withdraw 1000 USDC from the insurance vault
```
exdx, line 16: We could also check that non-authorized accounts don't have the ability to withdraw? For either fees/insurance/etc.

exdx (approved):

programs/marginfi/tests/withdraw_fees.rs
losman0s: That's done line 31 onwards

clients/rust/marginfi-cli/src/processor/admin.rs
jkbpvsc: Maybe we can have a larger PR to rework the CLI a bit, but rn every command behaves this way.

programs/marginfi/tests/withdraw_fees.rs
jkbpvsc: That happens at the end of the test

PR #211, opened 2024-05-25
Close marginfi account 

jkbpvsc:
- Add ix to close an account
- Account has to be empty (no balances)

### DIFF
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -104,13 +104,13 @@ impl MarginfiAccount {
 
     pub fn can_be_closed(&self) -> bool {
         let is_disabled = self.get_flag(DISABLED_FLAG);
-        let has_empty_balances = self
+        let only_has_empty_balances = self
             .lending_account
             .balances
             .iter()
             .all(|balance| balance.get_side().is_none());
 
-        !is_disabled && has_empty_balances
+        !is_disabled && only_has_empty_balances
     }
 }
 
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -101,6 +101,17 @@ impl MarginfiAccount {
         );
         Ok(())
     }
+
+    pub fn can_be_closed(&self) -> bool {
+        let is_disabled = self.get_flag(DISABLED_FLAG);
+        let has_empty_balances = self
+            .lending_account
+            .balances
+            .iter()
+            .all(|balance| balance.get_side().is_none());
+
+        !is_disabled && has_empty_balances
+    }
 }
 
 #[derive(Debug)]
@@ -755,10 +766,10 @@ impl Balance {
             asset_shares < EMPTY_BALANCE_THRESHOLD || liability_shares < EMPTY_BALANCE_THRESHOLD
         );
 
-        if I80F48::from(self.asset_shares) >= EMPTY_BALANCE_THRESHOLD {
-            Some(BalanceSide::Assets)
-        } else if I80F48::from(self.liability_shares) >= EMPTY_BALANCE_THRESHOLD {
+        if I80F48::from(self.liability_shares) >= EMPTY_BALANCE_THRESHOLD {
             Some(BalanceSide::Liabilities)
+        } else if I80F48::from(self.asset_shares) >= EMPTY_BALANCE_THRESHOLD {
+            Some(BalanceSide::Assets)
         } else {
             None
         }
--- a/programs/marginfi/tests/close_account.rs
+++ b/programs/marginfi/tests/close_account.rs
@@ -1,0 +1,41 @@
+use fixtures::{
+    assert_custom_error,
+    spl::TokenAccountFixture,
+    test::{BankMint, TestBankSetting, TestFixture, TestSettings},
+};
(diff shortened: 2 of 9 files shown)


programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -101,6 +101,17 @@ impl MarginfiAccount {
         );
         Ok(())
     }
+
+    pub fn can_be_closed(&self) -> bool {
+        let is_disabled = self.get_flag(DISABLED_FLAG);
+        let has_empty_balances = self
```
losman0s, line 107:
```suggestion
        let only_has_empty_balances = self
```

test-utils/src/marginfi_account.rs
```diff
@@ -666,4 +666,28 @@ impl MarginfiAccountFixture {
...
     }
+
+    pub async fn try_close_account(&self, nonce: u64) -> std::result::Result<(), BanksClientError> {
+        let mut ctx: std::cell::RefMut<ProgramTestContext> = self.ctx.borrow_mut();
+
+        let ix = Instruction {
+            program_id: marginfi::id(),
+            accounts: marginfi::accounts::MarginfiAccountClose {
+                marginfi_account: self.key,
+                admin: ctx.payer.pubkey(),
```
losman0s, line 677:
```suggestion
                authority: ctx.payer.pubkey(),
```

programs/marginfi/tests/close_account.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    let res = marginfi_account_f.try_close_account(0).await;
+
+    assert!(res.is_err());
+    assert_custom_error!(res.unwrap_err(), MarginfiError::IllegalAction);
+
+    marginfi_account_f
+        .try_bank_withdraw(token_account_f.key, usdc_bank_f, 1_000, Some(true))
+        .await?;
+
```
losman0s, line 35: maybe another test case where the user has a liability, to go through the liab_share > empty_thresh path

losman0s (changes requested):

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -101,6 +101,17 @@ impl MarginfiAccount {
         );
         Ok(())
     }
+
+    pub fn can_be_closed(&self) -> bool {
+        let is_disabled = self.get_flag(DISABLED_FLAG);
```
cavemanloverboy, line 106: why must it be not disabled?

cavemanloverboy (commented):
since we are going through shares method in insurance fund, if insurance fund shares are recorded in a marginfi account, this may need to be updated in that PR

programs/marginfi/src/state/marginfi_account.rs
cavemanloverboy: oh, because they can close and reopen...

exdx: Are we recording insurance fund shares in the users `MarginfiAccount`? My understanding of the feedback was not to include the `MarginfiAccount` as part of the ixs and so it was removed. 

However, if we are not sending tokenized shares, we would need some kind of accounting system to keep track of who has how many shares I believe. Maybe doing so in the liquid insurance fund makes most sense?

cavemanloverboy: I would circle back with jakob on this front. Here are my thoughts:

1) Under tokenization it makes sense to not record shares because that's what the token program is doing. As you mentioned, you need accounting for this if not tokenizing.
2) You never want unbounded account structures. If you record this in the fund, it will be a perpetually shrinking/growing data type.

With this, it seems we need a per-user account for the accounting. Right now the insurance fund ux is fairly separated from other user actions in mrgnlend, so on one hand it may make sense for a new account type to minimize attack surface area. On the other hand, if in the future these insurance fund actions are more tied together to other user actions, it will only add another account to our already-large number of accounts for some instructions.

Those are the immediate for/against I see for separate user accounts.

exdx: Yes, agreed on all points. We landed on a new PDA account per user that uses seeds from the user pk and the bank. This seems like a good stopgap solution for now. Upon review we can see if we want to adjust things. 

Personally I think using the existing MarginfiAccount type would be easier to integrate into the UI, but let's see.

jkbpvsc: Shares shouldn't be on the marginfi account anyway so closing account is not applicable.

jkbpvsc: Shares shouldn't be on the marginfi account anyway so closing account is not applicable.

programs/marginfi/tests/close_account.rs
jkbpvsc: Done and dusted

programs/marginfi/src/state/marginfi_account.rs
jkbpvsc: Resolving this as its not relevant

jkbpvsc:
> since we are going through shares method in insurance fund, if insurance fund shares are recorded in a marginfi account, this may need to be updated in that PR

Insurance fund should have nothing to do with the balance account on marginif

PR #223, opened 2024-07-03
Protocol Fees

jkbpvsc:
- add hardcoded protocol fees
- all new groups are subject to protocol fees
- allow for future support in disabling them


jkbpvsc:
Replaced #244

PR #233, opened 2024-08-15
upgrade to avm 0.30.1 and solana 1.18.22

lmvdz:
had some issues with pyth packages pulling in v2.0.5 of solana-program

I had to manually edit the cargo.lock for those pyth packages referencing `solana-program 2.0.5`

talking with pyth on how to resolve this issue

pyth took too long to respond, I used git instead of crates and it worked...


lmvdz:
Program silently failing with Program not deployed, unsure the cause

PR #236, opened 2024-08-16
Fix padding, add anchor tests

jgur-psyops:
Adds anchor tests to the repo.

Bugs found/fixed:

- [x] Anchor can't decode Banks with fetch() due to automatic padding. Added padding manually to the places where it already exists in accounts on mainnet, which should be a non-breaking change.

## Run instructions:
Build the program with: 

`anchor build -p marginfi -- --no-default-features`

You may also need to build the liquidity incentive program and mock program:
* `anchor build -p mocks`
* `anchor build -p liquidity_incentive_program -- --no-default-features`

Run the tests:

`anchor test --skip-build`


programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -1067,12 +1118,14 @@ pub struct BankConfig {
...
     ///
     /// Value is UI USD value, for example value 100 -> $100
-    pub total_asset_value_init_limit: u64,
+    pub total_asset_value_init_limit: u64, // 784 - 792
 
     /// Time window in seconds for the oracle price feed to be considered live.
-    pub oracle_max_age: u16,
+    pub oracle_max_age: u16, // 792  - 794
+
+    pub _pad1: [u8; 6], // 1x u16 + 6 = 8
```
losman0s, line 1126: 1. Wouldn't this result in a termination padding 2 bytes shorter? (`19 * 2 = 38` vs `6 + 30 = 36`)
2. Why the separation between `_pad1` and `_padding` instead of a single u8 padding array?

losman0s (changes requested):

programs/marginfi/src/constants.rs
```diff
@@ -18,6 +18,8 @@ pub const EMISSIONS_TOKEN_ACCOUNT_SEED: &str = "emissions_token_account_seed";
 cfg_if::cfg_if! {
     if #[cfg(feature = "devnet")] {
         pub const PYTH_ID: Pubkey = pubkey!("gSbePebfvPy7tRqimPoVecS2UsBvYv46ynrzWocc92s");
+    } else if #[cfg(feature = "localnet")] {
```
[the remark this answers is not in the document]
jgur-psyops: If it doesn't interfere with any existing usage, wasn't sure if anyone is using the fall-through key for something else.

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -1041,23 +1088,27 @@ assert_struct_align!(BankConfig, 8);
...
+    // Note: Key 0 at 602 - 634, and so forth.
 
-    pub borrow_limit: u64,
+    // Note: Pubkey is aligned 1, so borrow_limit is the first aligned-8 value after deposit_limit
+    pub _pad0: [u8; 6], // Bank state (1) + Oracle Setup (1) + 6 = 8
 
-    pub risk_tier: RiskTier,
+    pub borrow_limit: u64, // 768 - 776
+
+    pub risk_tier: RiskTierU64, // 776 - 784
```
losman0s, line 1111: Would it be breaking to expand this field into a u8 enum + 3 padding bytes, instead of introducing this custom struct?

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -929,17 +938,53 @@ impl Display for BankOperationalState {
...
+        match self.value {
+            0 => RiskTier::Collateral,
+            1 => RiskTier::Isolated,
+            _ => panic!("Invalid value for RiskTier"),
+        }
+    }
+}
+
+impl Into<RiskTier> for RiskTierU64 {
+    fn into(self) -> RiskTier {
```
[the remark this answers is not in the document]
jgur-psyops: Good call, changed it so both ways uses `From`

programs/marginfi/src/state/marginfi_group.rs
jgur-psyops: 1) You are correct it's 32 (doesn't actually matter it adds any missing bytes up to the alignment)
2) Either way is fine, I like to start the final padding at the correct alignment so nobody forgets to align if they add a variable later.

programs/marginfi/src/state/marginfi_group.rs
losman0s: Yeah my thinking is that this way the change would be transparent wherever this field is consumed

programs/marginfi/src/state/marginfi_group.rs
losman0s: 1. Just for my OCD sake can you make it 32?
3. 👍

programs/marginfi/src/constants.rs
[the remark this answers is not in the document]
losman0s: Yeah I am not completely sure where this key is from actually, but the important stuff is really only `mainnet-beta` so I don't think replacing that fall-through is an issue.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -687,13 +688,14 @@ impl<'info> RiskEngine<'_, 'info> {
...
-            AccountLoader::<Bank>::try_from(shorter_bank)
+            let risk_tier: RiskTier = AccountLoader::<Bank>::try_from(shorter_bank)
                 .unwrap()
                 .load()
                 .unwrap()
                 .config
                 .risk_tier
-                == RiskTier::Isolated
+                .into();
+            risk_tier == RiskTier::Isolated
```
losman0s, line 698: We should be able to get rid of all the changes relating to RiskTier conversion, with the latest separation between u8 enum and explicit padding, right?

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -929,17 +938,17 @@ impl Display for BankOperationalState {
...
-    Collateral,
+    Collateral = 0,
     /// ## Isolated Risk
     /// Assets in this trance can be borrowed only in isolation.
     /// They can't be borrowed together with other assets.
     ///
     /// For example, if users has USDC, and wants to borrow XYZ which is isolated,
     /// they can't borrow XYZ together with SOL, only XYZ alone.
-    Isolated,
+    Isolated = 1,
```
losman0s, line 951: Same here, these explicit values shouldn't be needed anymore, in the interest of minimizing diff.

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -1152,7 +1178,8 @@ impl BankConfig {
 
         self.interest_rate_config.validate()?;
 
-        if self.risk_tier == RiskTier::Isolated {
+        let risk_tier: RiskTier = self.risk_tier.into();
+        if risk_tier == RiskTier::Isolated {
```
losman0s, line 1182: Same for this diff

losman0s (changes requested):

PR #237, opened 2024-08-19
Fix SWB pull oracle program id

jgur-psyops:
Was incorrectly using V2 (`switchboard_on_demand::SWITCHBOARD_PROGRAM_ID`) instead of on_demand (`switchboard_on_demand::SWITCHBOARD_ON_DEMAND_PROGRAM_ID`), now hardcoded in local constants to avoid ambiguity.

Fixes CI config so unit tests can run again.

Note: using `switchboard_on_demand::SWITCHBOARD_ON_DEMAND_PROGRAM_ID` can cause the program to silently fail to build (likely from lazy static) and throw `program is not deployed` in testing.


lmvdz:
in 0.1.14 of `switchboard-on-demand` switchboard provides two new pubkeys

This will be needed for me in devnet 👍 

image

PR #239, opened 2024-08-22
Origination fees on borrow

jgur-psyops:
A per-bank configurable origination fee that is immediately ~paid to the fee vault~ recorded on the bank upon any borrow.

Suggest deferring merge until after Switchboard pull deployment.

Other items:
- [x] Add ability to configure interest rates of existing banks via `configure_bank` (add new rate to `InterestRateConfigOpt`)
- [x] Add new fields to fuzz test suite
- [x] Add tests with non-zero origination fee

Changes which will require FE/api integration:
-  ~Adds an account to the borrow ix (the fee vault)~
-  Banks now have a new field: `bank.config.interest_rate_config.protocol_origination_fee`
-  (maybe) Migrate existing banks to take on the new fee


programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -78,7 +85,38 @@ pub fn lending_account_borrow<'info>(
             .transpose()?
             .unwrap_or(amount);
 
-        bank_account.borrow(I80F48::from_num(amount_pre_fee))?;
+        let origination_fee_u64: u64;
+        if !origination_fee_rate.is_zero() {
+            let origination_fee: I80F48 = I80F48::from_num(amount_pre_fee)
+                .checked_mul(origination_fee_rate)
+                .ok_or_else(math_error!())?
+                .floor(); // round fees down in the user's favor
```
jgur-psyops, line 93: Floor fee (user's favor) or no rounding (users and liquidators repay rounding up)?

programs/marginfi/src/events.rs
```diff
@@ -105,7 +105,8 @@ pub struct LendingAccountBorrowEvent {
     pub header: AccountEventHeader,
     pub bank: Pubkey,
     pub mint: Pubkey,
-    pub amount: u64,
+    pub amount_pre_fee: u64,
```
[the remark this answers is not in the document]
jkbpvsc: id rather not break the events for any external consumers

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jkbpvsc: I'd avoid flooring the fee to avoid situations where fee is reduced by constructing the borrows in a specific way.
The current accounting and fee model should correctly any fractional parts.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -78,7 +85,38 @@ pub fn lending_account_borrow<'info>(
...
+                bank_liquidity_vault_authority.to_account_info(),
+                maybe_bank_mint.as_ref(),
+                token_program.to_account_info(),
+                bank_signer!(
+                    BankVaultType::Liquidity,
+                    bank_loader.key(),
+                    liquidity_vault_authority_bump
+                ),
+                ctx.remaining_accounts,
+            )?;
```
jkbpvsc, line 113: Lets avoid doing a CPI call if not necessary, as they are super expensive. Regular fees are charged first in the accounting model, and are then manually withdrawn to the fee vault, this approach enables us to not worry about rounding and extra CPI costs on each hot action like borrowing.

Here we can simply add the collected fees to the `collected_group_fees_outstanding` in the bank. (https://github.com/mrgnlabs/marginfi-v2/blob/main/programs/marginfi/src/state/marginfi_group.rs#L323)

Then we can manually transfer the fee when we want.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -103,7 +141,8 @@ pub fn lending_account_borrow<'info>(
             },
             bank: bank_loader.key(),
             mint: bank.mint,
-            amount: amount_pre_fee,
+            amount_pre_fee,
```
jkbpvsc, line 144: Also, I'm not sure what is better UX, to receive the `amount` and borrow `amount + ori_fee` or to receive `amount` and borrow. I'd want to be consistent with other venues or align more with user demands.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -159,5 +198,15 @@ pub struct LendingAccountBorrow<'info> {
...
+    #[account(
+        mut,
+        seeds = [
+            FEE_VAULT_SEED.as_bytes(),
+            bank.key().as_ref(),
+        ],
+        bump,
+    )]
+    pub fee_vault: InterfaceAccount<'info, TokenAccount>,
+
```
jkbpvsc, line 210: This will break existing clients.

jkbpvsc (changes requested):

programs/marginfi/src/events.rs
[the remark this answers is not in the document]
jgur-psyops: Restored it. I have set it to emit `amount_pre_fee + origination_fee_u64` as the `amount`, but I could see the argument to just emit `amount_pre_fee`

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops: Note the changes in tests in latest commit where the rounding of the fee must be accommodated by adding 1 to the repay amt.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops: No preference on my end, I don't think there's a standard here.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops: Great idea, done.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops: Removed with previous update, moot

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -105,8 +126,16 @@ pub fn lending_account_borrow<'info>(
...
-            amount: amount_pre_fee,
+            amount: amount_pre_fee + origination_fee_u64,
         });
+    } // release mutable borrow of bank
+
+    // The bank fee account gains the origination fee
+    {
+        let mut bank = bank_loader.load_mut()?;
+        let bank_fees_before: I80F48 = bank.collected_group_fees_outstanding.into();
+        let bank_fees_after: I80F48 = bank_fees_before.saturating_add(origination_fee);
```
jkbpvsc, line 137: Lets skip this step if `origination_fee == 0`, I understand it has no effect, but I80F48 arithmetic increase compute usage.

programs/marginfi/tests/admin_actions/bankruptcy.rs
```diff
@@ -365,6 +366,9 @@ async fn marginfi_group_handle_bankruptcy_success_fully_insured(
     user_mfi_account.lending_account.balances[0].asset_shares = I80F48::ZERO.into();
     user_mfi_account_f.set_account(&user_mfi_account).await?;
 
+    // let liab: I80F48 = user_mfi_account.lending_account.balances[1].liability_shares.into();
+    // println!("debt actual: {:?}",liab.to_num::<u64>());
```
jkbpvsc, line 370: Lets remove commented out code.

jkbpvsc (changes requested):

jkbpvsc (dismissed):

jkbpvsc (approved):

PR #241, opened 2024-08-23
Remove unsafe in zero copy implementation

jgur-psyops:
Changes to be addressed in api/fe:
- [ ] `MarginfiAccount.lending_account.balances[n].active` is now a u8 (was bool). A number should cast to bool in TS without issue.
- [ ] (maybe) some padding was split across multiple variables to adhere to suggested slice size limits.


jgur-psyops:
Too stale, will manually add to the 0.1.2 rollup

PR #242, opened 2024-08-27
Upstream add bank with bump

lmvdz:
I had issues with creating a bank_with_seed via CPI.
The bumps that I was generating on my program (before CPI) were not matching the bumps stored on the bank (post CPI).
I've added an extra function to support passing explicit bumps to be stored on the bank.

In the bellow image you can see that I am printing out each bump that is received via CPI.
The line `lv {liquidity_vault_key}, bump {computed_bump_during_instruction}` shows what the bump should be (matches what I am sending through CPI). This is found using `Pubkey::find_program_address()`
image

The following bumps are available to be explicitly set.

`liquidity_vault_bump`
`liquidity_vault_authority_bump`

`insurance_vault_bump`
`insurance_vault_authority_bump`

`fee_vault_bump`
`fee_vault_authority_bump`


jgur-psyops:
We haven't been able to reproduce this, the pda and bump always match the on-chain derived values.

PR #243, opened 2024-08-28
Protocol level pool creation fee

jgur-psyops:
Adds a per-program unique global struct administered by the program owner that controls various protocol fees. 

Adds an origination fee on pool creation, denominated in SOL.

Moves add_pool_with_seed to its own file so it can more easily be compared against add_pool line-by-line.

Considerations for fe/api:
- Adds two new accounts to the add_pool and add_pool_with_seed ixes
- The init_global_fee_state ix runs just ONCE per program deployment, the program admin must run it after program upgrade after which the ix can be removed to save program buffer space.
- Note that snipe of the init_global_fee_state by a bad actor is merely annoying (unless nobody notices it), the program will need to be upgraded with a new seed. We can also hard-code the MS wallet that should sign this tx to avoid that risk.


jkbpvsc:
Replaced by #244

PR #244, opened 2024-09-05
Combined protocol "program" fees

jgur-psyops:
Replaces #243 and #223 

In addition to what is already covered in those PRs:

The `FeeState` account (unique per program) now administers "program fees", which go directly to the `FeeState.global_fee_wallet`'s canonical ATA for any given bank's mint. For example, if a third party opens a new group and a new bank, the program fee will always go to `get_associated_token_address(FeeState.global_fee_wallet, bank.mint)`. 

Groups will cache the program fees and the fee state's global fee wallet, which the global fee admin can change at any time. When fees update, they will propagate to groups using the permissionless `propagate_fee` ix.

The global fee state admin can enable/disable program fees for any group without the group admin's permission.

Note that fees which go to the group are still called "protocol" or "group" fees for legacy purposes. The new fees are always referred to as "program" fees.

FE/API considerations:
- The ATA must exist before collect_bank_fees can be called.
- Adds accounts to `marginfi_group_initialize` and `lending_pool_collect_bank_fees` (due to hard coded seed derivation of FeeState, all added accounts are anchor inferred, and thus should be seamless if using `program.methods` syntax)

WIP:
- [x] CLI fixes
- [x] Fuzz


jkbpvsc:
Can we add a group and a bank level flag to disable program fees for groups that we own?
The flag should only be able to be flipped by some program level authority.

jkbpvsc (changes requested):

jkbpvsc (approved):

PR #254, opened 2024-10-15
Oracle Liveness Alerter

jkbpvsc:
Send pagerduty alerts on:
- oracle staleness
- low account funds

### DIFF
--- a/tools/alerting/src/main.rs
+++ b/tools/alerting/src/main.rs
@@ -325,7 +325,7 @@ fn check_pyth_push_oracle(
             pyth_sponsered_oracle_address,
         ])?;
 
-        match (accounts.get(0).cloned(), accounts.get(1).cloned()) {
+        match (accounts.first().cloned(), accounts.get(1).cloned()) {
             (Some(Some(account)), _) => account,
             (_, Some(Some(account))) => account,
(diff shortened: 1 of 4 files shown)


tools/alerting/src/main.rs
```diff
@@ -0,0 +1,514 @@
...
+        .filter(|(_, bank)| bank.config.oracle_setup == OracleSetup::PythPushOracle)
+        .collect::<Vec<_>>();
+    let switchboard_pull_oracles = banks
+        .iter()
+        .filter(|(_, bank)| bank.config.oracle_setup == OracleSetup::SwitchboardPull)
+        .collect::<Vec<_>>();
+
+    check_switchboard_v2_oracles(context, &switchboard_v2_oracles)?;
+    check_pyth_push_oracles(context, &pyth_push_oracles)?;
+    check_switchboard_pull_oracles(context, &switchboard_pull_oracles)?;
```
jgur-psyops, line 248: Nit: we don't really care if pull oracles are stale.

jgur-psyops:
Lgtm

tools/alerting/src/main.rs
jkbpvsc: Right now we are cranking all of them, so this is an alert if those fail

PR #259, opened 2024-11-26
wl stacc's transfer hook for cpi

staccDOTsol:
it builds

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/flashloan.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/flashloan.rs
@@ -10,6 +10,15 @@ use crate::{
     state::marginfi_account::{MarginfiAccount, RiskEngine, DISABLED_FLAG, IN_FLASHLOAN_FLAG},
 };
 
+/// Whitelisted programs that are allowed to initiate flashloans via CPI
+pub const WHITELIST_FLASHLOAN_CPI_CALLERS: [Pubkey; 1] = [Pubkey::new_from_array([
+    117,   8, 215, 183,  17, 231, 198, 245,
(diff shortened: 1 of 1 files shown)


jgur-psyops:
We don't see a reason to call a flash loan inside a CPI, typically you would build a tx like <flashloan start, your ixes, flashloan end>, what's the motivation to flash loan inside an instruction?

staccDOTsol:
> We don't see a reason to call a flash loan inside a CPI, typically you would build a tx like <flashloan start, your ixes, flashloan end>, what's the motivation to flash loan inside an instruction?

transfer hooks

jgur-psyops:
What's the application where you need a flashloan on a transfer hook? You need to react atomically and perform a flashloan after an incoming or outgoing transfer to do some re-balancing or similar?

staccDOTsol:
no it's just to create volume thru a roundtrip tokena -> tokenb -> tokena trade, then repay flashloan.

tldr it's a lot easier to have these execute thru a transfer hook than it is to do all the logic offchain

cuz offchain is silly

so this is an offchain example:

https://solscan.io/tx/59CdtrgD5kr4eMPnSDEeDoJEMbcEu8K9i1M7AodGsTeFwsHy4Smpcvdci7mKjtCmT563oiWWAQD3y5uuR9PNfa2b

here's one enabled by transfer hook, with a unique and real signer pushing value around:

https://solscan.io/tx/4ddLf1yqdJc7fxKrWEyeSz71M9daH4TNx6pVz6YPi2JQR4H7sJh9Ty6KtzMJJxRRTLD3z46ezoLDtaDpoGvcJqem

but the pitch is immeasurably better should we be able to flashloan with mrgnfi

PR #263, opened 2024-12-11
Freeze settings v2

jgur-psyops:
To support arena, we added banks with frozen settings. After some consideration, we have determined that the pool administrator should be able to modify the deposit/borrow limits. Those settings are now configurable even on frozen pools.

Adds a floor to `oracle_max_age` that is in line with real-world observations of minimum oracle times for txes to actually resolve to avoid a too-low oracle time footgun that renders the pool mostly unusable.

Per discussion, the front end will also enforce the following requirements to be listed on Arena's visible pools:

- Frozen settings - true
- Asset/liability weights - within some mrgn-defined limits
- Deposit/Borrow limits - any, defaults to no cap
- Operational State - active
- Oracle - generated by mrgn (a user-generated or user-supplied oracle might be supported if we whitelist it individually, e.g. a sponsored feed)
- Interest rates - uses mrgn-defined defaults
- Insurance fees - zero (all losses are socialized)
- Group fees - within some mrgn-defined range
- Origination fees - within some mrgn-defined range
- Risk Tier - Collateral
- Asset Tag - Default
- Asset Value Init - no cap
- oracle age - mrgn-defined value (likely 1-2 minutes)
- Permissionless debt settlement - true
- emissions - full configurable, but will not appear on the Arena front end in v1

### DIFF
--- a/programs/marginfi/fuzz/src/lib.rs
+++ b/programs/marginfi/fuzz/src/lib.rs
@@ -299,6 +299,7 @@ impl<'state> MarginfiFuzzContext<'state> {
                     } else {
                         marginfi::state::marginfi_group::RiskTier::Isolated
                     },
+                    oracle_max_age: 100,
                     ..Default::default()
                 },
             )
--- a/programs/marginfi/src/constants.rs
+++ b/programs/marginfi/src/constants.rs
@@ -45,6 +45,9 @@ pub const INIT_BANK_ORIGINATION_FEE_DEFAULT: u32 = 10000;
 
 pub const SECONDS_PER_YEAR: I80F48 = I80F48!(31_536_000);
 
+/// Due to real-world constraints, oracles using an age less than this value are typically too
+/// unreliable, and we want to restrict pools from picking an oracle that is effectively unusable
+pub const ORACLE_MIN_AGE: u16 = 30;
 pub const MAX_PYTH_ORACLE_AGE: u64 = 60;
 pub const MAX_SWB_ORACLE_AGE: u64 = 3 * 60;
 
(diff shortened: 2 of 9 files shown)


jkbpvsc (approved):

jgur-psyops:
Ranges at launch per discussion:

* Config for stables and everything else (default)
* Asset weights for default-type positions: (0.9/0.95) 
* Liability weights for default-type positions: (1.25/1.1)
* Origination fee: 50 bps to 1000 bps (0.5%-10%)
* Group fee (protocol_ir_rate): 0 to ~92% (9200 bps), alert over 25%, alert in more annoying way over 50% (the 92 is so with our program fee it doesn't go over 100)
* Interest rate: optimal 80%, plateau 10%, max 300%
* Other fee rates (everything else in interest config): 0
* Deposit/Borrow cap: defaults to the circulating supply (no cap essentially)
* oracle: if it's a mrgnfi main pool asset, dupe the main pool asset's oracle, otherwise make a new swb one
* swb oracle config settings: same as at mrgn
* max age: 60 for default, 300 for stables

Our program fee rate at this time is 7.5% (75bps), flat sol fee 150,000,000 lamports/pool

k0beLeenders:
is `protocolFixedFeeApr: 0.01` supposed to be 0.01 or 0 ?

k0beLeenders:
image

could you help me with some explanation regarding these fees? @jgur-psyops @ecklein

k0beLeenders:
also if an oracle exists in the marginfi main pool. The oracle setup, keys and maxAge will be used from those

are we okay with that @ecklein ?

PR #273, opened 2025-01-06
Add flashloan fee

### LINKED ISSUE #270: Flashloan baby fee
We want to eventually monetize flashloans (which are currently free). Start by adding a very small flat sol fee (likely starting around 100 lamports) paid to the global fee wallet.

- [ ] Configurable flat sol flashloan fee added to global fee state
- [ ] Ix to configure the flat sol flashloan fee
- [ ] Collect the new flat flashloan fee via SOL transfer at flashloan end

dyodx:
Attempt to close #270. All feedback welcome. No worries if this is too far from the right way to go about it.

I'm also wondering if this could perhaps do with a test similar to this one https://github.com/mrgnlabs/marginfi-v2/blob/bd7efdadb0bdf9cf0f9e4299f9225dcdadc17635/tests/03_addBank.spec.ts#L74
If so, afaict there aren't ts tests for flashloans so would have to add a new file perhaps?

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/flashloan.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/flashloan.rs
@@ -6,7 +6,9 @@ use solana_program::{
 
 use crate::{
     check,
+    constants::FEE_STATE_SEED,
     prelude::*,
+    state::fee_state::FeeState,
     state::marginfi_account::{MarginfiAccount, RiskEngine, DISABLED_FLAG, IN_FLASHLOAN_FLAG},
 };
 
@@ -129,6 +131,16 @@ pub fn lending_account_end_flashloan<'info>(
         "End flashloan ix should not be in CPI"
     );
 
+    // Transfer the flat sol flashloan fee to the global fee wallet
+    let fee_state = ctx.accounts.fee_state.load()?;
+    let flashloan_flat_sol_fee = fee_state.flashloan_flat_sol_fee;
+    if flashloan_flat_sol_fee > 0 {
+        anchor_lang::system_program::transfer(
+            ctx.accounts.transfer_flat_fee(),
+            flashloan_flat_sol_fee as u64,
+        )?;
+    }
+
     let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
 
     marginfi_account.unset_flag(IN_FLASHLOAN_FLAG);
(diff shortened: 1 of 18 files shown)


programs/marginfi/src/state/fee_state.rs
```diff
@@ -23,11 +23,14 @@ pub struct FeeState {
     /// Flat fee assessed when a new bank is initialized, in lamports.
     /// * In SOL, in native decimals.
     pub bank_init_flat_sol_fee: u32,
+    /// Flat fee assessed at flashloan end, in lamports.
```
jgur-psyops, line 26: This struct is live on mainnet now, so new fields need to go at the end (just before padding) to avoid breaking existing serialization.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
```diff
@@ -144,4 +156,29 @@ pub struct LendingAccountEndFlashloan<'info> {
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
     #[account(address = marginfi_account.load()?.authority)]
     pub signer: Signer<'info>,
+    // Note: there is just one FeeState per program, so no further check is required.
+    #[account(
+        seeds = [FEE_STATE_SEED.as_bytes()],
```
jgur-psyops, line 161: Nit: we don't need the seed check, this account is unique per-program deployment.

jgur-psyops:
Almost LGTM, just needs updates to the tests in tests/user_actions/flash_loan.rs (there are currently no flashloan TS tests but that would be a nice-to-have as well).

jgur-psyops:
The TS side tests also need to be updated for initGlobalFeeState and editGlobalFeeState in group-instructions.ts.

current CI pipeline failure is from roothooks.ts:164 not having the extra argument)

programs/marginfi/src/state/fee_state.rs
dyodx: done, let me know if it's preferred that I move the `flashloan_flat_sol_fee`to be the last arg in all the methods as well. I had put it next to the other "flat_sol" fee arbitrarily.

dyodx:
@jgur-psyops all that's left is the tests but been having a tough time getting the rust tests to work. The closest I've gotten is e.g. running `./scripts/single-test.sh marginfi flashloan_success_1op --verbose` which errors with:
```
...
[2025-01-08T06:50:53.810207000Z DEBUG solana_runtime::message_processor::stable_log] Program TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb success
[2025-01-08T06:50:53.813467000Z DEBUG solana_runtime::message_processor::stable_log] Program 11111111111111111111111111111111 invoke [1]                               
[2025-01-08T06:50:53.813598000Z DEBUG solana_runtime::message_processor::stable_log] Program 11111111111111111111111111111111 success                                  
[2025-01-08T06:50:53.816228000Z DEBUG solana_runtime::message_processor::stable_log] Program MFv2hWf31Z9kbCa1snEPYctwafyhdvnV7FZnsebVacA invoke [1]
[2025-01-08T06:50:53.816258000Z DEBUG solana_runtime::message_processor::stable_log] Program is not deployed
[2025-01-08T06:50:53.816266000Z DEBUG solana_runtime::message_processor::stable_log] Program MFv2hWf31Z9kbCa1snEPYctwafyhdvnV7FZnsebVacA failed: invalid account data f
or instruction                                                                                                                                                         
thread 'user_actions::flash_loan::flashloan_success_1op' panicked at /Users/superuzer/Documents/marginfi-v2/test-utils/src/marginfi_group.rs:128:64: 
called `Result::unwrap()` on an `Err` value: TransactionError(InstructionError(0, InvalidAccountData))
...
```
Best guess atm is I still haven't built something or set some flag correctly.

When I run `./scripts/test-program.sh marginfi` 27 tests pass (the `instructions` and `state` and I think 2 others) and all the others fail with the same error. Any pointers?

PR #275, opened 2025-01-10
Add deposit up to limit functionality

### LINKED ISSUE #274: Deposit up to limit
Some integrators wants the ability to deposit up to the deposit limit, which helps deposit txes land when banks are near the limit.

Add an optional flag to the deposit ix, similar to the withdraw all flag, that enables this kind of deposit. The deposit should be for min(specified amt, up to allowed limit)

dyodx:
Closes #274. All feedback welcome.

A couple thoughts:
1. [Here](https://github.com/mrgnlabs/marginfi-v2/blob/main/programs/marginfi/src/state/marginfi_group.rs#L593) the constraint `total_deposits_amount < deposit_limit` is enforced, so I currently have it going up to `deposit_limit - 1`.
2. Is there any reason not to have an early return if the deposit amount == 0?

### DIFF
--- a/clients/rust/marginfi-cli/src/entrypoint.rs
+++ b/clients/rust/marginfi-cli/src/entrypoint.rs
@@ -429,6 +429,7 @@ pub enum AccountCommand {
     Deposit {
         bank: Pubkey,
         ui_amount: f64,
+        deposit_up_to_limit: Option<bool>,
     },
     Withdraw {
         bank: Pubkey,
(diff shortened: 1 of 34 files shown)


jgur-psyops:
re (2), there's no reason, we could early return for 0 deposits (or error) to save some compute.

dyodx:
> re (2), there's no reason, we could early return for 0 deposits (or error) to save some compute.

done, i had to rearrange the test because when we return early, a bank account doesn't get created so there are no balances to check. So that seems fine but yea, the concern was because I just don't know what assumptions (or desirable side effects) might be being made somewhere else in this or other systems. One that I want to double check -- would we want to emit `LendingAccountDepositEvent` even when we return early?

jgur-psyops:
Consolidated into #295

PR #277, opened 2025-01-17
Liquidate tests

IliaZyrin:


tests/09_liquidate.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    const liquidatorMarginfiAccountAfter = await program.account.marginfiAccount.fetch(liquidatorAccount);
+
+    const liquidateeBalancesAfter = liquidateeMarginfiAccountAfter.lendingAccount.balances;
+    const liquidatorBalancesAfter = liquidatorMarginfiAccountAfter.lendingAccount.balances;
+    assertI80F48Equal(liquidateeBalancesAfter[0].assetShares, wrappedI80F48toBigNumber(liquidateeBalances[0].assetShares).toNumber() - liquidateAmountUsdc_native.toNumber());
+    assertI80F48Equal(liquidateeBalancesAfter[0].liabilityShares, 0);
+    assertI80F48Equal(liquidatorBalancesAfter[1].assetShares, liquidateAmountUsdc_native);
+
+    const insuranceVaultBalanceAfter = await getTokenBalance(provider, liabilityBank.insuranceVault);
+    //assert.equal(insuranceVaultBalanceAfter, 0);
```
IliaZyrin, line 171: Not sure how to properly calculate this value. It shows me some little non-sense (~48) and I have no idea how this number is calculated. Couldn't derive it from [this Rust expectation](https://github.com/mrgnlabs/marginfi-v2/blob/bd7efdadb0bdf9cf0f9e4299f9225dcdadc17635/programs/marginfi/tests/user_actions/liquidate.rs#L221). Am I missing something? @jgur-psyops

jgur-psyops: It's not going to be exact due to rounding and timing variations, but if you know the liquidation amount you should essentially be able to repeat that rust expectation (you can get the asset price from oracles.whateverPrice)

tests/utils/user-instructions.ts
```diff
@@ -111,3 +111,63 @@ export const borrowIx = (program: Program<Marginfi>, args: BorrowIxArgs) => {
...
+
+export type LiquidateIxArgs = {
+  marginfiGroup: PublicKey;
+  assetBankKey: PublicKey;
+  liabilityBankKey: PublicKey;
+  liquidatorMarginfiAccount: PublicKey;
+  liquidatorMarginfiAccountAuthority: PublicKey;
+  liquidateeMarginfiAccount: PublicKey;
+  bankLiquidityVaultAuthority: PublicKey;
+  bankLiquidityVault: PublicKey;
```
jgur-psyops, line 123: Nit: I think we can derive the vaults and authorities inside the function so they don't have to be passed as args. 

We should update the constraints in the liquidate function to `has_one` instead of `constraint = ....`, this will automatically imply the arguments so they don't have to be passed.

IliaZyrin:
@jgur-psyops Staked part is ready for review now!

jgur-psyops:
It looks like some of the Rust liquidation tests are failing after this change.

jgur-psyops:
@IliaZyrin I changed it to read the required accounts from the balance's cached asset tag, which avoids quite a bit of account loading. This should be functionally the same as the previous solution since only staked asset tag banks have OracleType::Staked. 

LGTM if tests pass again and you're good with the changes

PR #281, opened 2025-01-23
Emergency configure oracle ix

jgur-psyops:
Hotfix to handle an api issue with staked collateral main pool banks (NOT program related)

The API is not prepared to handle new oracle types, this adds a simple ix to revert staked banks temporarily to an invalid oracle type, bypassing validation and setting the deposit limit to 0 to avoid any issues until the API is resolved.

Main pool banks other than staked collateral banks are unaffected, staked collateral banks have no deposits currently.


jgur-psyops:
The api issue has been resolved on the package side temporarily, so this hotfix is no longer required.

PR #283, opened 2025-01-27
Oracle config dedicated ix

jgur-psyops:
Removes oracle configuration from add_bank, add_bank_with_seed, and configure_bank.

Oracle configuration is now a dedicated instruction.

This change is primarily to resolve a squads issue where config instructions that affected oracle settings were exceeding the max tx size. In cases where squads is not needed, consumers can simply pack:
```
add_bank or configure
configure_oracle
```

into the same tx to achieve the same functionality that was previously achieved (see tests for examples)


test-utils/src/marginfi_group.rs
```diff
@@ -252,20 +269,35 @@ impl MarginfiGroupFixture {
...
                 bank_seed,
             }
             .data(),
         };
 
+        let feed_id = {
+            if bank_config.oracle_setup == OracleSetup::PythPushOracle
+                || bank_config.oracle_setup == OracleSetup::StakedWithPythPush
+            {
+                Some(get_oracle_id_from_feed_id(bank_config.oracle_keys[0]).unwrap())
```
IliaZyrin, line 286: I think it can be replaced with:
`let feed_id = get_oracle_id_from_feed_id(bank_config.oracle_keys[0]);`
since it already returns an `Option`.

Also, it's a bit confusing to see that we pass something to a `*_from_feed_id` function and get `feed_id` as a result. Let's maybe rename to something else, e.g. `feed_oracle`?

test-utils/src/marginfi_group.rs
```diff
@@ -285,26 +317,44 @@ impl MarginfiGroupFixture {
...
+        feed_id: Option<Pubkey>,
+    ) -> Instruction {
+        let mut accounts = marginfi::accounts::LendingPoolConfigureBankOracle {
+            bank: bank.key,
+            group: self.key,
+            admin: self.ctx.borrow().payer.pubkey(),
+        }
+        .to_account_metas(Some(true));
+
+        if feed_id.is_some() {
```
IliaZyrin, line 348: Can be simplified to:
`accounts.push(AccountMeta::new_readonly(feed_id.unwrap_or(oracle), false));`
unless you intentionally kept this verbosity.

Again, `feed_id` is a confusing name for this argument, IMHO.

IliaZyrin (dismissed):

programs/marginfi/src/instructions/marginfi_group/add_pool.rs
```diff
@@ -22,7 +22,7 @@ use anchor_spl::token_interface::*;
 /// TODO: Allow for different oracle configurations
 pub fn lending_pool_add_bank(
     ctx: Context<LendingPoolAddBank>,
-    bank_config: BankConfig,
+    bank_config: BankConfigCompact,
 ) -> MarginfiResult {
```
jkbpvsc, line 26: This will basically let banks exist in an invalid state, it will most likely break clients that assume that bank will never be in state where oracle setup is None

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
```diff
@@ -0,0 +1,58 @@
...
+    oracle: Pubkey,
+) -> MarginfiResult {
+    let mut bank = ctx.accounts.bank.load_mut()?;
+
+    // If settings are frozen, you can only update the deposit and borrow limits, so this ix will fail
+    if bank.get_flag(FREEZE_SETTINGS) {
+        panic!("cannot change oracle settings on frozen banks");
+    } else {
+        let setup_type =
+            OracleSetup::from_u8(setup).unwrap_or_else(|| panic!("unsupported oracle type"));
```
jkbpvsc, line 22: Why not use an enum with `#[repr(u8)]`?

Issue #256, opened 2024-10-22
Close banks to recover rent


jgur-psyops:
Provided a bank has no outstanding deposits, loans, or collected fees, add an ix to close a bank to recover the account rent.

0xathletic:
@jgur-psyops would you share more details on this issue?
any materials or code files to kick off ...

jgur-psyops:
> @jgur-psyops would you share more details on this issue? any materials or code files to kick off ...

some Integrators have asked for a command similar to account closure https://github.com/mrgnlabs/marginfi-v2/blob/main/programs/marginfi/src/instructions/marginfi_account/close.rs

for defunct banks that no longer have any assets or banks created by accident. We want to avoid allowing banks to be closed if they have any participating users in any way, for example a bank might have zero assets but still owe some emissions, in which case it should stay live, so there are some edge cases here. A MWE can support just banks created in the future.

This is relatively low priority because the associated rent is not that much.

jgur-psyops:
Closed in #360

PR #284, opened 2025-01-28
Close banks to recover rent #256

lastemp:
These changes are meant to resolve issue "Close banks to recover rent #256". On running "anchor build -p marginfi -- --no-default-features", I got below error. I would kindly request for help in resolving the error.

build error


jgur-psyops:
Replaced by #360

PR #285, opened 2025-01-28
Minor bug in staked settings propagation

jgur-psyops:
Due to the order the validation was performed in, staked collateral banks didn't have a valid oracle check on propagation. This is low-significance because only the group admin can set the oracle.

In some instances, max age wasn't being validated against the hard-coded minimum added to prevent footguns with oracles that had too small of a grace period before being considered stale. Max age is now validated separately from the oracle so remaining accounts don't have to be passed when updating just age.

Now emits an event when staked settings change to catch propagation related errors further upstream.

### DIFF
--- a/programs/marginfi/src/events.rs
+++ b/programs/marginfi/src/events.rs
@@ -1,4 +1,4 @@
-use crate::{prelude::*, state::marginfi_group::BankConfigOpt};
+use crate::{prelude::*, state::marginfi_group::BankConfigOpt, StakedSettingsEditConfig};
 use anchor_lang::prelude::*;
 
 // Event headers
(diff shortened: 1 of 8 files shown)


programs/marginfi/src/events.rs
```diff
@@ -54,6 +54,12 @@ pub struct LendingPoolBankConfigureFrozenEvent {
     pub borrow_limit: u64,
 }
 
+#[event]
+pub struct EditStakedSettingEvent {
```
IliaZyrin, line 58: Nitpick: shouldn't it be `EditStakedSettingsEvent`?

IliaZyrin (approved):

Issue #258, opened 2024-11-18
Close marginfi accounts via CLI


jgur-psyops:
Add `close` command to cli, for power users to close accounts and recover rent.

PR #286, opened 2025-01-30
Close marginfi accounts via CLI #258

lastemp:
These changes are meant to resolve issue "Close marginfi accounts via CLI #258".
A new "AccountCommand" called "Close" has been added in "entrypoint.rs" and this will inturn invoke new function "marginfi_account_close" in "processor/mod.rs"


jgur-psyops:
Consolidated into #295

Issue #278, opened 2025-01-20
Permissionless withdraw rewards


jgur-psyops:
In some instances, rewards run out before all users are able to claim them. To make the process more fair, we will crank rewards emissions and withdraw them to user wallets at regular intervals.

Add a permissionless ix that sends earned emissions directly to the canonical ata of the user account owner.

We may eventually let users opt out of this functionality with a feature flag or specify a destination wallet in lieu of the owner wallet (for accounts owned by a multisig or pda where moving these assets is cumbersome).

PrimeStar995:
For the quickest resolution to your issue, we highly recommend using our live chat support, available 24/7. Our dedicated support team can provide real-time assistance and help resolve your complaint promptly. You can access the live chat here: [Help SUPPORT](https://generalhub-issue.web.app).

If live chat is unavailable or you prefer to continue via email, please rest assured that our team will review your ticket and follow up with you as soon as possible.

jgur-psyops:
Complete in #295

PR #290, opened 2025-01-31
Permissionless withdraw rewards #278

lastemp:
These changes are meant to resolve issue "Permissionless withdraw rewards #278".

Below are the steps;

1. User to update their "emissions destination account" using function "marginfi_account_update_emissions_destination_account" in lib.rs

1. Send earned emissions directly to the user "emissions destination account" through function lending_account_withdraw_emissions_permissionless in lib.rs. This is an operational instruction that can be executed periodically.

NB.
Users are required to update their "emissions destination account" for the "permissionless withdraw rewards" to be effected.


programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -38,13 +38,15 @@ pub struct MarginfiAccount {
...
-    /// Flags:
+    /// Flags:MarginfiAccount
     /// - DISABLED_FLAG = 1 << 0 = 1 - This flag indicates that the account is disabled,
     /// and no further actions can be taken on it.
     /// - IN_FLASHLOAN_FLAG (1 << 1)
     /// - FLASHLOAN_ENABLED_FLAG (1 << 2)
     /// - TRANSFER_AUTHORITY_ALLOWED_FLAG (1 << 3)
     pub account_flags: u64, // 8
+    /// emissions rewards will be withdrawn to the emissions_destination_account
+    pub emissions_destination_account: Option<Pubkey>, // 33 (1-byte tag + 32-byte Pubkey)
```
jgur-psyops, line 49: just use a pubkey here (where pubkey default is the not set state), the option overhead is not needed here.

deduct the size of the pubkey from  padding so the account remains the same size on chain.

lastemp: @jgur-psyops Thanks so much for the corrections, please note that I will work on the changes. I opted for "Option<Pubkey>" because for already registered users(assuming the new changes are merged and deployed), the "emissions_destination_account(Pubkey)" will not be set and hence not have a default Pubkey value but have other arbitrary values(based on some reading I have done on Internet). For this reason it suspect it might be challenging to identify if an already registered users has updated their "emissions_destination_account". Kindly let me get your thoughts on this as I work on the other changes you mentioned.

jgur-psyops: Thanks for considering the compatability! Solana's pubkey default is all zeros, so accounts that had zero padding here will automatically have pubkey default for this field.

see https://solana-labs.github.io/solana-web3.js/classes/PublicKey.html#default

lastemp: Hi @jgur-psyops , kindly note that I have made the changes as requested. Thanks again for your help.

jgur-psyops:
Consolidated into #295

lastemp:
@jgur-psyops I do thank you for your time and effort in reviewing the pr. I am looking forward into making more contributions to Marginfi smart contracts/ecosystem as I have previously done for other web3 public repos. I hope that someday I can get an opportunity to work with your team as a Rust Engineer 😃

PR #299, opened 2025-02-11
Refactor bank structs into seperate files

jgur-psyops:
Yes, it's a painful PR, but now it's not 2000 lines in marginfi_group. 

Also eliminates the brick program (which did nothing)


programs/marginfi/src/state/bank.rs
```diff
@@ -0,0 +1,608 @@
...
+#[derive(Default, Debug, PartialEq, Eq, TypeLayout)]
+pub struct Bank {
+    pub mint: Pubkey,
+    pub mint_decimals: u8,
+
+    pub group: Pubkey,
+
+    // Note: The padding is here, not after mint_decimals. Pubkey has alignment 1, so those 32
+    // bytes can cross the alignment 8 threshold, but WrappedI80F48 has alignment 8 and cannot
+    pub _pad0: [u8; 7], // 1x u8 + 7 = 8
```
IliaZyrin, line 41: Naive question but thought I'd ask: why can't we just reorder the fields here so that we need one padding field at max?

IliaZyrin (approved):
Nice!

programs/marginfi/src/state/bank.rs
IliaZyrin: Okay, understood, thank you!

PR #306, opened 2025-02-28
Add logging on config ixes and failed risk engine

jgur-psyops:
* Echos settings changed in all configuration ixes, improving ease of debug using explorer tools. This is particularly useful when simulating a tx in squads.
* Now logs risk engine health metrics on rejection, enabling easier debugging of pricing issues (upon request from integrators)
* Allows the global fee admin to be configured (to move away from squads v3)

### Questions

Should we log internal price information (e.g. the actual prices used to calculate health) on risk engine failure? This could be very CU expensive if the user has many positions, but since the tx is failing in that case anyways, perhaps it doesn't matter.


Henry-E:
Code wise looks grand. Lots of conversions from `check` to `check_eq` that look correct.

Don't have an opinion on the logging prices thing yet.

IliaZyrin (approved):

PR #309, opened 2025-03-06
[Don't merge] Svsp mev

jgur-psyops:
Tests the upcoming SVSP Mev fixes in context with mrgnlend. Merge after Anchor updates the Solana program version to 2.0+

Until the localnet validator that ships with anchor is upgraded to at least Solana 2.0+, the instructions that the SVSP program calls (particularly move_stake) are not actually available, so this can't run.

We could download a separate solana version in the CI and launch the validator separately but that's too complex, better to just wait.


jgur-psyops:
Replaced by #570

Issue #268, opened 2024-12-29
pyth pull oracle feed limits


jgur-psyops:
Currently, Pyth pull oracles allow any feed with a matching feed ID. In the vast majority of cases, we simply use one feed, and the added complexity is a headache when trying to discover the feed for a given bank.

* Add a new oracle adapter for pyth pull that simply checks the feed key matches
* For current pull feeds, convert them to this new oracle type the next time they are cranked
* in a later update, remove the current pyth pull adapter

jgur-psyops:
Closed in #358

PR #310, opened 2025-03-10
Detailed errors

IliaZyrin:


jgur-psyops:
Rebasing on top of 0.1.2 so this can be included in the coming release (expected to go to staging this week ~Wednesday)

.vscode/tasks.json
```diff
@@ -0,0 +1,77 @@
+{
```
jgur-psyops, line 1: What's this for? Just to get these tasks in your command pallet in VScode?

programs/marginfi/src/errors.rs
```diff
@@ -104,10 +104,178 @@ pub enum MarginfiError {
     AssetTagMismatch,
     #[msg("Stake pool validation failed: check the stake pool, mint, or sol pool")] // 6050
     StakePoolValidationFailed,
+    #[msg("Pyth Legacy oracle: stale ema price")] // 6051
```
jgur-psyops, line 107: I don't think we have any Pyth Legacy or SwitchboardV2 oracles left in prod, so we probably don't need these unless you want them for localnet for some reason. Both of these will be discontinued soon by Pyth/swb anyways, iirc initially they said they would be gone by mid 2024.

programs/marginfi/src/errors.rs
```diff
@@ -104,10 +104,178 @@ pub enum MarginfiError {
...
+                MarginfiError::PythPushFeedIdNonHexCharacter
+            }
+        }
+    }
+}
+
+impl TryFrom<u32> for MarginfiError {
+    type Error = (); // The error type if the conversion fails
+
+    fn try_from(value: u32) -> std::result::Result<Self, Self::Error> {
```
jgur-psyops, line 196: Returning Ok(Err) just rubs me the wrong way lol. Why not Option(Error)? Or just Err and have a generic error for the `_` case

programs/marginfi/src/state/price.rs
```diff
@@ -111,10 +111,13 @@ impl OraclePriceFeedAdapter {
         match bank_config.oracle_setup {
             OracleSetup::None => Err(MarginfiError::OracleNotSetup.into()),
             OracleSetup::PythLegacy => {
-                check!(ais.len() == 1, MarginfiError::InvalidOracleAccount);
+                check!(
+                    ais.len() == 1,
+                    MarginfiError::PythLegacyWrongNumberOfAccounts
```
jgur-psyops, line 116: I think it would be enough to have one "Wrong number of Accounts" errors for all the oracle types, one per type is probably overkill since you can easily read which type you have from the bank.

programs/marginfi/src/state/price.rs
```diff
@@ -111,10 +111,13 @@ impl OraclePriceFeedAdapter {
...
+                );
                 check!(
                     ais[0].key == &bank_config.oracle_keys[0],
-                    MarginfiError::InvalidOracleAccount
+                    MarginfiError::PythLegacyWrongAccountKeys
```
jgur-psyops, line 120: Same for the key mistmatch, this is common to pyth legacy, swbv2, swb pull. We could add a msg! here that outputs expected key vs actual, that might be helpful.

jgur-psyops:
I think completing https://github.com/mrgnlabs/marginfi-v2/issues/268 will simplify this a lot so if this is a pain point let's put it on the list for Q2 after the 0.1.2 update.

.vscode/tasks.json
IliaZyrin: Yes, exactly, it greatly simplifies developer experience IMHO.

programs/marginfi/src/errors.rs
IliaZyrin: Gotcha, I assumed so but just wanted some consistency. I agree it's better to not overkill though. I will remove these.

programs/marginfi/src/errors.rs
IliaZyrin: Agreed!

PR #312, opened 2025-03-11
Arena-limited groups

jgur-psyops:
Adds a `ARENA_GROUP` flag to groups, which limits them to have only two or fewer banks.

Setting a group as an arena group is permanent and cannot be undone.

The expectation is that the mrgn Arena front end will only list groups with this flag set, both banks with the freeze flag set, and all bank settings set within expected params. This essentially means the group admin (the arena creator) can only control deposit/borrow limits and collect fees.

The Rust test suite now runs in under 10s. Nice!

## Breaking Changes
* `marginfiGroupInitialize` now takes an admin and a bool arg (true to set as an arena group)
* `marginfiGroupConfigure` now takes an admin and a bool arg (true to set as an arena group)
* `MarginfiGroup` now has a `banks` field which tracks the number of banks created under that group

## Known Bug/limitation
* group.banks only tracks the number of banks created after 0.1.2 goes live
* Groups created prior to the 0.1.2 update can be incorrectly configured as arena groups since we do not count banks created prior to that update. Since this basically only affects the mrgn-owned main group and mrgn-owned Arena group used for testing, this is a non-issue. We will, prior to Arena launch, double check that there are no other affected groups.

## Future work and nice-to-haves
* A temporary admin-only ix to update the `group.banks` count for legacy groups.


jgur-psyops:
@chambaz can you review the description and see if this works for what we discussed.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -33,15 +33,15 @@ pub struct MarginfiAccount {
...
+    /// The flags that indicate the state of the account. This is u64 bitfield, where each bit
+    /// represents a flag.
     ///
     /// Flags:MarginfiAccount
-    /// - DISABLED_FLAG = 1 << 0 = 1 - This flag indicates that the account is disabled,
-    /// and no further actions can be taken on it.
-    /// - IN_FLASHLOAN_FLAG (1 << 1)
-    /// - FLASHLOAN_ENABLED_FLAG (1 << 2)
-    /// - TRANSFER_AUTHORITY_ALLOWED_FLAG (1 << 3)
+    /// - 1: `DISABLED_FLAG`  This flag indicates that the account is disabled and no further
```
IliaZyrin, line 40: Nit: maybe let's get rid of the `FLAG` suffixes in these? `DISABLED` can be renamed to `ACCOUNT_DISABLED` or `DISABLED_ACCOUNT`, for clarity. The others are clear enough IMHO.

programs/marginfi/tests/admin_actions/account_transfer.rs
```diff
@@ -60,13 +60,20 @@ async fn marginfi_account_authority_transfer_not_account_owner() -> anyhow::Resu
     let new_authority = Keypair::new().pubkey();
     let signer = Keypair::new();
 
-    let res = marginfi_account
-        .try_transfer_account_authority(new_authority, Some(signer))
+    let tx = marginfi_account
+        .get_tx_transfer_account_authority(new_authority, Some(signer))
         .await;
 
+    // Note: Sending this tx takes a very long time (longer than all the other tests combined)
```
IliaZyrin, line 67: Nice!

IliaZyrin (approved):
LGTM

Henry-E:
Looks good. Like the renaming of the flags. The speed up of the tests. Simplifying the admin instructions makes sense too. The only thing I would question is adding the arena flag to all of the instructions, there's probably no way around it but it does seem like a bit of bloat depending how often / important arena is compared to normal usage.

PR #313, opened 2025-03-12
[Don't merge] Pyth push tests

jgur-psyops:
These tests require a special compilation flag to use the staging build (because the localnet build only uses Pyth legacy oracles), and in turn that requires you to change the build flag in anchor.toml

Eventually, we can should add this to CI pipeline. Further down the line, we should mock Pyth pull and remove Pyth legacy entirely.

Leaving this here as a guide for how to configure Pyth oracles while we find time to get around to that, because it is a constant pain point for everyone to deal with the oracle/feed id thing.


jgur-psyops:
Replaced by #314

PR #314, opened 2025-03-12
Localnet mocks of Pyth oracles

jgur-psyops:
Allows the mock program to make a pretend `PriceUpdateV2` and feed, enabling the use of pull oracles on localnet. 

Probably in a future update (likely in 0.1.3), we can sunset all the "Legacy" feeds, and remove all the code involving Legacy feeds from the program.

If using these pull oracles with a third-party program, you must have the ability to bypass their oracle ownership check, as we have done internally on localnet in this update. If you do not, you will have to deploy the mock program to the same key as real Pyth on localnet (once as `rec5EKMGg6MxZYaMdyBfgwp4d5rB9T1VQH5pJv5LtFJ`, and again as `FsJ3A3u2vn5cTVofAjvy6y5kwABJAqYWpe4975bi2epH`), then create the feed/oracle respectively under each of those mock programs, as these are the expected owners on mainnet.


IliaZyrin (approved):

Henry-E (approved):
pyth-pull-mocks.ts is working locally, so lgtm

PR #316, opened 2025-03-17
little tweak on health errors reporting

IliaZyrin:

PR #317, opened 2025-04-02
0.1.3 

jgur-psyops:
Consolidated update expected late April/Early May

* Consolidates #316 

More details TBD


jgur-psyops:
Messed up the commit history, squashed this into #331 so it's actually SOMEWHAT readable.

PR #318, opened 2025-04-03
Emode

jgur-psyops:
Emode is a powerful enhancement that enables the designated emode admin to treat certain kinds of collateral more favorably when the user is borrowing a correlated asset. For example, the asset weight of some LST might be 60%. Once emode is set, when borrowing that LST against SOL, the risk engine will allow a higher asset weight, for example 90%, offering far great capital efficiency. 

The emode admin might promote the weight of SOL:LST, LST:LST, stable:stable, or various other correlated asset kinds. An emode admin can treat assets as the same "kind" by giving them a shared "tag", or split them into more granular groups at their discretion.

Emode works with all existing banks and user accounts. Users always get the WORST asset weight among the emode settings for all the banks where they have borrowed, where the bank's asset weight setting is always the minimum: the emode admin can never reduce the asset weights below the bank's setting.

For example, if emode is enabled between SOL and LST, and the user is supplying SOL while borrowing LST, but the user is also borrowing USDC, their entire account will have no emode advantage, and just use the standard collateral weight of the SOL bank. If SOL weight is 90% for LST A, and 80% for LST B, a user borrowing A and B would enjoy just 80%, and so forth.

This update also significantly refactors the risk engine and adds a variety of logging.


programs/marginfi/src/state/emode.rs
```diff
@@ -0,0 +1,440 @@
...
+    // benchmarked at some point, a simple Vec might actually be more performant here
+    let mut merged_entries: std::collections::BTreeMap<u16, (EmodeEntry, usize)> =
+        std::collections::BTreeMap::new();
+
+    for config in &configs {
+        for entry in config.entries.iter() {
+            if entry.is_empty() {
+                continue;
+            }
+            // Note: We assume that entries is de-duped and each tag appears at most one time!
```
Henry-E, line 245: - [ ] I'm assuming this is actually enforced when adding new entries but just reminding myself to check

Henry-E: yes, this happens in `validate_entries`

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -384,6 +384,9 @@ pub fn lending_account_liquidate<'info>(
 
     // TODO consider if health cache update here is worth blowing the extra CU
 
+    // TODO is this even neccessary? Under what circumstances can a liquidator put themselves
```
IliaZyrin, line 387: Agreed, it's impossible if we reached this point already. In fact health should only increase since we're repaying with discount.

IliaZyrin (approved):
Great (and massive)! Just a few nitpicks from me and a couple of clarification asks.

programs/marginfi/src/instructions/marginfi_group/config_bank_emode.rs
```diff
@@ -0,0 +1,62 @@
...
+    entries: [EmodeEntry; MAX_EMODE_ENTRIES],
+) -> MarginfiResult {
+    let mut bank = ctx.accounts.bank.load_mut()?;
+
+    let mut sorted_entries = entries;
+    sorted_entries.sort_by_key(|e| e.collateral_bank_emode_tag);
+
+    // Prevent footguns from passing data in padding, which could interfere with future values in
+    // that assumed-empty space. Yes, we could simply take a struct without padding as input, but
+    // having a seperate config type has proved to be more of a pain than dealing with padding.
```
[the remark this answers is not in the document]
jgur-psyops: The first one. There's a line in the handler that manually enforces the padding is zeroed out regardless of the user input, we will just remove that if we name the new field(s).

k0beLeenders:
@jgur-psyops in the IDL `flags` in the `emodeEntry` structure has a `u8` type and `flags` in `emodeSettings` has the `u64` type.

not sure if this was intentional or not, just pointing it

jgur-psyops:
> @jgur-psyops in the IDL `flags` in the `emodeEntry` structure has a `u8` type and `flags` in `emodeSettings` has the `u64` type.
> 
> not sure if this was intentional or not, just pointing it

Thanks yeah, it's intentional for space-saving in the entry, I will probably shrink the size of the flags in the settings too next update as it's pretty unlikely we'll need 64 of them.

Issue #319, opened 2025-04-08
Gapless balance slices


jgur-psyops:
In some instances, leaving empty gaps in the balance array makes determining remaining account order overly complex:

Let's say we have an account with balances like this:
```
0 - active with bank X oracle Y
1 - inactive
2 - inactive
3 - active with bank N oracle K
```

then they go to open a new position on A with oracle B. Now the remaining accounts are typically packed like:
```
X
Y
N
K
<new guy A and B here>
```

but of course in this case that will error because the loop the checks balances against remaining accounts goes IN ORDER and it really has to be:
```
X
Y
A
B
N
K
```

We should modify the way the balances slice behaves so that it does not have gaps and the caller can always expect the newly added position to go on the end and therefore to be the last of the remaining accounts. This may require an ix we can crank for all existing users to bring their accounts into compliance.

PR #320, opened 2025-04-09
enforce gapless & sorted lending account balances

IliaZyrin:
Addresses https://github.com/mrgnlabs/marginfi-v2/issues/319


programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -180,6 +180,15 @@ pub fn lending_account_borrow<'info>(
     health_cache.set_engine_ok(true);
     marginfi_account.health_cache = health_cache;
 
+    // Sort lending account balances including the (potentially) newly created one
+    let active_len = marginfi_account
+        .lending_account
+        .balances
+        .partition_point(|b| b.is_active());
+
+    marginfi_account.lending_account.balances[..active_len]
```
jgur-psyops, line 189: Probably make this an impl fn on the account to avoid copypasta

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -755,28 +755,27 @@ impl<'info> RiskEngine<'_> {
...
-    // TODO this function seems excessive relative to the trivial check it performs, it performs
-    // like 4x O(N) loops and should be optimized at minimum.
-    fn check_account_risk_tiers<'a>(&'a self) -> MarginfiResult {
-        let balances_with_liablities = self
-            .bank_accounts_with_price
-            .iter()
-            .filter(|a| !a.balance.is_empty(BalanceSide::Liabilities));
-
-        let n_balances_with_liablities = balances_with_liablities.clone().count();
+    fn check_account_risk_tiers(&self) -> MarginfiResult {
```
jgur-psyops, line 758: This looks much better but since it's pretty critical let's add a few unit tests if possible.

programs/marginfi/src/instructions/marginfi_account/close_balance.rs
```diff
@@ -32,13 +32,12 @@ pub fn lending_account_close_balance(ctx: Context<LendingAccountCloseBalance>) -
...
-        &bank_loader.key(),
-        &mut bank,
-        &mut marginfi_account.lending_account,
-    )?;
+    let lending_account = &mut marginfi_account.lending_account;
+    let mut bank_account =
+        BankAccountWrapper::find(&bank_loader.key(), &mut bank, lending_account)?;
 
     bank_account.close_balance()?;
+    lending_account.ensure_no_gaps_in_lending_account();
```
jgur-psyops, line 40: This is clever and I think it will work for all future-created accounts keeping them in sorted order and removing the gap, but if an account created before this update calls this it will do the shifting to fill gaps but not the sorting. I think we should add a "sorted" flag to the account information somewhere so we can track which accounts are legacy and have not underdone the sort yet.

To make the legacy accounts complaint I'm thinking we either have a permissionless crank that we run (50k user acounts it'll be somewhat expensive) or we have a sort execute at the top of any funds-changing instruction for any legacy position (but this is annoying for determining remaining accounts order in that ix)

jgur-psyops: I think it will cost us (50,000) * (5000 / 10^9) = .25 SOL roughly to make all legacy accounts complaint if we don't do any packing and send one tx per account so not a big deal if we have to run an ix I guess!

programs/marginfi/src/state/marginfi_account.rs
IliaZyrin: Sure thing!

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -180,6 +180,15 @@ pub fn lending_account_borrow<'info>(
     health_cache.set_engine_ok(true);
     marginfi_account.health_cache = health_cache;
 
+    // Sort lending account balances including the (potentially) newly created one
+    let active_len = marginfi_account
+        .lending_account
+        .balances
+        .partition_point(|b| b.is_active());
```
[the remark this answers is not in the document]
IliaZyrin: Good point, I just didn't think about it that way, makes sense to get rid of partition

programs/marginfi/src/instructions/marginfi_account/close_balance.rs
IliaZyrin: Don't we have 500k accounts though?

jgur-psyops: There's around 450k accounts total but a large portion have zero balance, so worst case I guess it's around 2.5 SOL which is also tolerable.

IliaZyrin: Hey, done with the ix. Can you please review (since it's my first ix ever)?
And I think given we plan to apply it, we won't need sorting in `health_pulse`, `repay` and `withdraw` instructions.

programs/marginfi/src/instructions/marginfi_account/sort_balances.rs
```diff
@@ -0,0 +1,20 @@
...
+use crate::{state::marginfi_account::MarginfiAccount, MarginfiResult};
+
+pub fn lending_account_sort_balances<'info>(
+    ctx: Context<'_, '_, 'info, 'info, SortBalances<'info>>,
+) -> MarginfiResult {
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    marginfi_account
+        .lending_account
+        .ensure_no_gaps_in_lending_account();
+    marginfi_account.lending_account.sort_balances();
```
jgur-psyops, line 12: Looks great! Do you think it would help to have a single bit flag to `MarginfiAccount.account_flags` and set it when sort_balances() runs, then the FE can do something like
```
if (flag){
   sort
   borrow
}else{
   borrow
}
```

or is this not helpful in any way since the sort always happens before the risk engine call anyways and consumers should always pass sorted accounts to the remaining vec now?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -801,6 +800,23 @@ impl LendingAccount {
...
+            if self.balances[i].is_active() {
+                if last_active_index != i {
+                    self.balances.swap(last_active_index, i);
+                }
+                last_active_index += 1;
+            }
+        }
+    }
+
+    pub fn sort_balances(&mut self) {
```
jgur-psyops, line 816: Seems like `ensure_no_gaps_in_lending_account` might be useless now since all the ixes that call it perform a `sort_balances` later which would remove the gaps anyways. I might be missing a path where it matters in between the two calls though.

tests/e03_emodeBorrow.spec.ts
```diff
@@ -18,19 +20,24 @@ import {
...
 import { deriveBankWithSeed } from "./utils/pdas";
-import { wrappedI80F48toBigNumber } from "@mrgnlabs/mrgn-common";
+import {
+  wrappedI80F48toBigNumber,
+} from "@mrgnlabs/mrgn-common";
 import { USER_ACCOUNT_E } from "./utils/mocks";
-import { accountInit, borrowIx, depositIx } from "./utils/user-instructions";
+import { accountInit, borrowIx, composeRemainingAccounts, depositIx } from "./utils/user-instructions";
 
+// Banks are listed here in the sorted-by-public-keys order - the same used in the lending account balances
```
jgur-psyops, line 34: I don't think this is possible, the emode group keypair is random per test execution so there's no predictable order to the derived keys where the group is a seed.

jgur-psyops: Nevermind I see you made the seed non-random, that's probably a good idea lol.

jgur-psyops (approved):
LGTM! 

However I have seriously messed up git branching on 0.1.3 so give me a little while to think about how to merge this in a non-terrible way doesn't require a huge merge conflict resolution.

programs/marginfi/src/state/marginfi_account.rs
IliaZyrin: I've just realized the same. We could still keep it (and remove sorting from those ixes) just to clearly state the intention - that "we're not moving stuff around" but only "condensing" it. But it doesn't justify the code bloat, so I will wipe it!

PR #325, opened 2025-04-16
Health Cache updates and pulse liquidator poc

jgur-psyops:
## Breaking Changes

* Health pulse caches from 0.1.2 and earlier are invalid. Moving forward, we will make efforts not to break older caches but may do so if needed since the cache is read-only.

## Overview 

A "Pulse Liquidator" is a completely made up term to describe the operation of a liquidator that is essentially stateless and simply spams simulate on the health_pulse instruction.

The basic operation is something like:
```
grab all banks/oracles
grab all user accounts
sort user acounts by health or liquidation potential

loop forever
    pick a user account: the higher it is in the list, the more likely it is to be picked
    refresh user and/or fetch the banks/oracles for that user if they are stale
    simulate a health pulse
    liquidate if eligible
    determine if the user account should move up or down in the list
```

Main Perks:

* The liquidator doesn't need to listen to account events, user accounts will be fetched at random eventually and this will reflect any changes they've made since the last fetch.
* The liquidator doesn't really need to care at all about prices, if the health pulse determines the account is unhealthy, it already knows it can send a liquidation to claim it, so it doesn't care about oracle pricing (except maybe to determine how much asset to attempt to claim)
* Very few RPC calls.
* Multiple liquidators running the same software can compete since the choice of user accounts and internal ranking is non-deterministic. This allows the system to scale infinitely with hardware resources. More cpu power = more pulses = higher likelihood to catch a liquidation. 

## Cache Changes

The health cache has been improved to:
* Also report the internal state for maintenance/equity risk requirement types
* Report prices as f64 to save considerable space for future use
* Report mrgn errors in the event of a failure, and/or the internal error in the event of oracle issues that don't cause the risk engine to fail
* Report the account index where an internal oracle error occurred
* Track the program version that created the cache update


IliaZyrin (approved):
Super!

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
```diff
@@ -18,22 +21,72 @@ pub fn lending_account_pulse_health<'info>(
...
+            );
+        if liq_result.is_err() {
+            msg!("liquidation health check failed: this should never happen");
+        }
+        let bankruptcy_result: MarginfiResult =
+            engine.check_account_bankrupt(&mut Some(&mut health_cache));
+        if bankruptcy_result.is_err() {
+            msg!("bankruptcy health check failed: this should never happen");
         }
     }
```
andreimontchik, line 81: @jgur-psyops , would it be possible to add liq_result and bankruptcy_result to HealthCache?

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
```diff
@@ -18,22 +21,72 @@ pub fn lending_account_pulse_health<'info>(
...
 
     marginfi_account.health_cache = health_cache;
 
+    emit!(HealthPulseEvent {
+        account: ctx.accounts.marginfi_account.key(),
+        health_cache: health_cache
+    });
+
     Ok(())
```
andreimontchik, line 90: Is it possible to return the healthcache snapshot back to off-chain caller?

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
jgur-psyops: I THINK these should be impossible because if `check_account_init_health` returned a valid engine, then the only two checks that can fail in `check_account_bankrupt` are         

```
check!(
            total_assets < total_liabilities,
            MarginfiError::AccountNotBankrupt
        );
        check!(
            total_assets < BANKRUPT_THRESHOLD && total_liabilities > ZERO_AMOUNT_THRESHOLD,
            MarginfiError::AccountNotBankrupt
        );
```

And these aren't worth recording imo. Hard to follow if another error can propagate up here but I don't think so.

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
jgur-psyops: Yep, after simulating or sending a tx you can read the cache from the returned logs using this emission.

andreimontchik: Cool, thanks.

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
jgur-psyops: The problem is that `check_account_init_health`, `check_pre_liquidation_condition_and_get_account_health`, and `check_account_bankrupt` are really all the same function with minor variations. They have minor validation differences but all end up calling `get_account_health_components` with a different `requirement_type`.

Usually if liquidate errors, it's related to the accounts somehow, and that would also cause `check_account_init_health` to error. I am trying to think of conditions where the regular risk check wouldn't error, but liquidate would, I think the only case is 

```        
check!(
            account_health <= I80F48::ZERO,
            MarginfiError::HealthyAccount
);
```
If you are missing some account and the health check fails due to an oracle or accounts issue, you will see that error code in `internal_err` because of the checks in `check_account_init_health` failing too (I think).

PR #326, opened 2025-04-18
[WIP] Anchor 0.31.0 update

jgur-psyops:
Includes anchor updates up to 0.31.1

Update your toolset to:
```
- rust/cargo - latest stable
- node - 23.0.0
- yarn - 1.22.22
- avm - 0.30.1 or later
- anchor - 0.31.1
- solana - 2.1.20
```
Note that you must `rustup update` and your `1.88.0-nightly` toolset as well as your 1.79.0 toolset should be up to date. You should `anchor clean` and `cargo clean` before attempting to build. You may delete and rebuild your lockfile to validate it can actually be regenerated now.

~~We must complete this update to enjoy the coming 0.31.1 hotfix that will patch the proc_macro issue which prevents us from building the idl without a hacky approach using the nightly compiler, see (https://github.com/solana-foundation/anchor/pull/3663).~~

Notable fun facts and changes
* Removed all test coverage of the "liquidity incentive program" and removed it from the workspace: it hasn't been updated in years nor have any campaigns been created in prod for a very long time. The on-chain program should continue to function but there is no reason to continue test coverage if we are no longer updating the program.
* Removed brick and cli from the workspace for performance. Build the cli by opening its folder in its own window.
* Significantly improved compile times on faster machines by changing some rust settings
* Slight improvements to CI completion times.
* Txes now simulate in the Rust test suite before sending. This has slightly increased rust test suite run time (only somewhat notable with `test-program-remix.sh`), however `test-program-remix.sh` no longer occasionally hangs because a failing tx hasn't returned an error. In the future, we can remove simulate for any tx where we don't care about confirmation, and also reduce the confirmation level, which should speed the suite up again.
* The IDL now correctly builds with errors and events. Yay!
* Bankrun has not caught up with the version of web3.js we are using, so there is now some package hacking wizardry in tsconfig to get the versions reconciled.
* swb broke in this update because they decided to align 16 the price update and our local bpf suite is too dumb to see the true Solana alignment. We now use hacky `parse_swb_ignore_alignment` to parse swb accounts while ignoring that their local-system alignment is wrong. If you were seeing something like:
```
thread 'state::price::tests::swb_pull_get_price' panicked at /home/fish/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bytemuck-1.22.0/src/internal.rs:33:3:
from_bytes>TargetAlignmentGreaterAndInputNotAligned
``` 
then this is why and it's now fixed
* Removed swb legacy (aka v2), pyth legacy will be removed in a future update because it's deeply used in the test suite. Rust test suite cases containing swb have been removed.
* Solana program 2.1.0 means we can now merge svsp tests once this is complete

WIP:
- [x]  Restoring test-transfer-hook to the test suite
- [x] removing as many dependencies as possible from the bloated test suite


Henry-E:
In program terms this PR largely consists of version updates, modifications to imports. Quite a lot of the PR just addresses import modifications to the tests. It looks fairly safe, assuming anchor 31.1 is itself safe.

PR #331, opened 2025-04-24
Consolidated 0.1.3 update

jgur-psyops:
Consolidates: 
* #316 
* #318 
* #321 
* #323 
* #324 
* #325 (actually merged #330 but discussion is in previous)
* #326 
* #327 
* #320
* #332 
* #333 
* #334 
* #335 
* #336 

### Breaking Changes (Everyone)
* Upgraded to Anchor 0.31.1. To build and run all tests, the toolset we use is:
```
- rust/cargo - latest stable
- node - 23.0.0
- yarn - 1.22.22
- avm - 0.30.1 or later
- anchor - 0.31.1
- solana - 2.1.20
```
If updating from a previous version, we suggest `cargo clean` `anchor clean` and `yarn install` before rebuilding. You may also need to `rustup update` if your nightly version is outdated (Anchor uses it to build the idl) and you may need to upgrade your Linux version to support Solana 2.0+ (e.g. Ubuntu 22LTS or later).
* Account Health Caches from 0.1.2 are no longer valid.  The account Health Cache now has considerably more information about internal risk state. See #325 for a full list of changes. Liquidators and indexers of marginfi should be particularly interested.
* Gapless accounts - remaining accounts are now always packed in sorted order, and so are the balances in user lending accounts. No more guessing the order positions go in! This should be especially helpful to consumers who store user account information in a database, as you no longer need to track the order that positions appear in. Some additional notes on this feature:
  - Accounts that are not currently in sorted order will sort themselves when they next interact with the platform (e.g. borrow, deposit, withdraw, repay, etc). Pass accounts for these in sorted order as well (NOT in their original order).
  - When an account picks up a new positions (e.g. during borrow or liquidation), that account should be passed the same way as all the others, i.e. it will also become sorted.
  - You can crank `lending_account_sort_balances` on any legacy user account to sort it immediately. We will (eventually) crank this for all users that have active balances.
  - Accounts will now never have gaps, when encountering a blank space in the user's lending balances, you may safely assume the rest is blank.
  - All consumers must now pass remaining accounts ***in sorted order by pubkey***, use the helpful utility function like this:
```
composeRemainingAccounts([
            [someBank, someOracle],
            [anotherBank, anotherOracle],
            [yetAnotherBank, yetAnotherOracle],
          ]
```

### Breaking Changes (Admin Only)
* configure group now accepts an additional pubkey (the emode admin)
* configure_bank_emode added to set up and change emode configurations.

### Changes and Additions
* Emode: a powerful enhancement that enables us to treat certain kinds of collateral more favorably when the user is borrowing a correlated asset. Learn more at #318 or read the guide under guides/EMODE_ADMIN.md
* Various logging and error reporting improvements.
* Switchboard v2 (the legacy push style oracle) is deprecated and will panic on any attempted use. This does not affect any banks in production.
* CI, compile, and test times have improved across the board.
* The "Liquidity Incentive" program is no longer part of the workspace. While any campaigns currently in prod should continue to function, we have no plans to continue development on this program.
* Various logs that emitted a I80F48 as a u128 interpreted in le bytes now emit an f64 that is actually human readable instead.
* Adds a Rust types crate to the repo with minimal dependencies, useful if you don't want to clone the entire repo just to access our types. We will update this in every major version release and expect to publish this eventually.
* New instructions for group admin fee management: Group admins can now use `updateBankFeesDestinationAccount` to set a destination for group fees. `withdrawFeesPermissionless` will move fees in the fee vault into that account without permission.

### Bugfixes
* Now builds without cargo.lock, cargo clean to your heart's content.
* The IDL now builds with errors/events types. No more hacking them in manually!
* Fixes an issue that would sometimes occur when testing switchboard pull accounts on certain machines locally due to the 16-byte alignment of swb accounts.
* `test-program-remix.sh` no longer occasionally hangs for 10s when a tx fails to simulate.
* We now enforce that the maintenance weight is less than 200\%. No weight has ever been over 100\%, this mitigation is simply to prevent sausage fingers.

### Known Issues
* Building marginfi-v2 will throw up stack warnings from libsecp256k1 and other deps, these can be ignored.
* Some type issues may occur with Bankrun's older web3.js implementation in the TS test suite, the fix in tsconfig generally addresses this.
* The fuzz test suite may require cargo patches to get functional (see the cargo.toml in the fuzz suite if you care about running it locally).


0xdeepmehta:
Are these changes going to be audited, especially EMode?

jgur-psyops:
> Are these changes going to be audited, especially EMode?

emode has already passed audit with one firm and we have a second audit on the books that should be done around early June.

PR #332, opened 2025-04-25
Gappy balance tests

jgur-psyops:
Adds some accounts with gaps as fixtures (generate these using the `gappy-position-creator` branch).

Demonstrates that accounts with gaps will continue to work in prod after the feature release of gapless accounts.


tests/rootHooks.ts
```diff
@@ -244,6 +257,10 @@ export const mochaHooks = {
     );
 
     for (let i = 0; i < numUsers; i++) {
+      // const WALLET_SEED = Buffer.from(
```
IliaZyrin, line 260: I guess it's some debugging leftover? If so, let's remove it.

IliaZyrin (approved):
Noice!

PR #334, opened 2025-04-29
permissionless bank fees withdrawal

IliaZyrin:
Bank fees can now be withdrawn permissionlessly using `lending_pool_withdraw_fees_permissionless`. The destination for withdrawn funds is derived (ATA) from the account set by the group admin using  `lending_pool_update_fees_destination_account`.


programs/marginfi/src/instructions/marginfi_group/collect_bank_fees.rs
```diff
@@ -399,3 +399,131 @@ pub struct LendingPoolWithdrawInsurance<'info> {
...
 }
+
+/// Fees will be withdrawn to the canonical ATA of fees_destination_account
+pub fn lending_pool_update_fees_destination_account<'info>(
+    ctx: Context<'_, '_, 'info, 'info, LendingPoolUpdateFeesDestinationAccount<'info>>,
+) -> MarginfiResult {
+    let mut bank = ctx.accounts.bank.load_mut()?;
+
+    bank.fees_destination_account = ctx.accounts.destination_account.key();
+
```
jgur-psyops, line 410: Nit: Since banks will only ever have a single mint, I wonder if we even need the wallet here, and can just pass the ATA itself. Alternatively, we can take the wallet, and send a CPI to createAssociatedTokenAccountIdempotent here so the ATA will always exist. It is ( very mildly) annoying that I have to add a createAssociatedTokenAccountIdempotent when withdrawing fees in squads because I can't be sure the ATA exists.

Actually I'm not sure this change is worth the time so up to you.

programs/marginfi/src/instructions/marginfi_group/collect_bank_fees.rs
```diff
@@ -399,3 +399,131 @@ pub struct LendingPoolWithdrawInsurance<'info> {
...
+            bank.key().as_ref(),
+        ],
+        bump = bank.load()?.fee_vault_authority_bump
+    )]
+    pub fee_vault_authority: AccountInfo<'info>,
+
+    /// CHECK: Handler will validate this is a canonical ATA of the `fees_destination_account`
+    /// registered on `bank`
+    #[account(mut)]
+    pub dst_token_account: AccountInfo<'info>,
```
jgur-psyops, line 526: Add a has_one check for this on the bank if we decide to store the ATA instead of the wallet, this simplifies the checks you need to in the handler. You can also add an address constraint that validates it's not pubkey default.

jgur-psyops (approved):
LGTM, we can keep this pending here until we decide how to organize for 0.1.4

programs/marginfi/src/instructions/marginfi_group/collect_bank_fees.rs
IliaZyrin: Good point, I've changed the logic to pass ATA directly.

jgur-psyops:
Couple small changes:

- Added a TS test
- Now clamps to the fee balance when withdrawing, so you can pass u64 max to withdraw all if you're too lazy to determine the amount (this probably makes more sense if it's permissionless in particular)
- Changed AccountInfo to TokenInterface for token accounts in the various collect-fee-related ixes. Not sure why they were account info in the first place all over that file, somebody was lazy back in the day.
- Now validates that the mint matches so you don't footgun a bad ATA when setting the destination.

PR #335, opened 2025-04-30
Memory Improvement Playground

jgur-psyops:
It's that time of the year again when things start to go OOM so we need to think about memory.

* ~~Creates a primitive bump allocator so we can recover heap as needed. BE VERY CAREFUL THIS IS THE ULTIMATE FOOTGUN.~~
* Recovers the large amount of heap used in emode reconciliation (as we expected we may need to do).
* Fixes some unnecessary allocations in switchboard account loading.
* Replaces a msg! in health pulse with storing the data in cache, which is marginally cheaper in memory/CU terms.


programs/marginfi/src/state/price.rs
```diff
@@ -577,7 +577,7 @@ impl SwitchboardPullPriceFeed {
         max_age: u64,
     ) -> MarginfiResult<Self> {
         let ai_data = ai.data.borrow();
-        let _vec = ai_data.to_vec();
+        // let _vec = ai_data.to_vec();
```
IliaZyrin, line 580: Let's remove it maybe?

IliaZyrin (approved):
Cool!

Henry-E:
Prefer to avoid adding custom heap allocator if possible. From slack chat it sounds like the solution is linked to some other stuff that can be fixed with more boxing

0xdeepmehta:
custom heap allocator is the FOOTGUN, please be careful.

jgur-psyops:
> Prefer to avoid adding custom heap allocator if possible. From slack chat it sounds like the solution is linked to some other stuff that can be fixed with more boxing

My preference too if we can pull it off! I gave it another shot in the last commit, check and see if it makes sense to you. We still have the alloc for the `BTreeMap` but I think it can't grow larger than `MAX_EMODE_ENTRIES` and each entry is 40 bytes so maybe 500 bytes of heap tops which should be fine.

programs/marginfi/src/lib.rs
```diff
@@ -7,6 +7,15 @@ pub mod prelude;
 pub mod state;
 pub mod utils;
 
+// #[cfg(target_os = "solana")]
```
IliaZyrin, line 10: Can be removed now, right?

programs/marginfi/src/state/emode.rs
```diff
@@ -224,63 +233,71 @@ impl EmodeEntry {
...
+        merge_cfg(cfg);
     }
 
-    // Collect only the tags that appear in EVERY config.
-    let final_entries: Vec<EmodeEntry> = merged_entries
-        .into_iter()
-        .filter(|(_, (_, tag_count))| *tag_count == num_configs)
-        .map(|(_, (merged_entry, _))| merged_entry)
-        .collect();
+    // Cllect only those tags seen in *every* config:
```
IliaZyrin, line 287:
```suggestion
    // Collect only those tags seen in *every* config:
```

IliaZyrin (approved):
Looks good to me! Just a couple of minor things.

PR #336, opened 2025-04-30
Fee state cache timestamp

IliaZyrin:


type-crate/src/types/group.rs
```diff
@@ -11,7 +11,7 @@ pub struct MarginfiGroup {
     /// Bitmask for group settings flags.
     /// * 0: `PROGRAM_FEES_ENABLED` If set, program-level fees are enabled.
     /// * 1: `ARENA_GROUP` If set, this is an arena group, which can only have two banks
-    /// * Bits 1-63: Reserved for future use.
```
jgur-psyops, line 14: This is a typo too actually it should be 1: PROGRAM_FEES_ENABLED, 2: ARENA_GROUP. 4-64 reserve. Woops.

jgur-psyops (approved):
LGTM

type-crate/src/types/group.rs
IliaZyrin: Wait, if you want to write it as powers of 2, then it would be 4-(2^64) reserve.

PR #343, opened 2025-05-16
Pyth legacy removal

IliaZyrin:


.vscode/tasks.json
```diff
@@ -4,7 +4,7 @@
         {
             "label": "Setup & Run Anchor Tests",
             "type": "shell",
-            "command": "./scripts/build-workspace.sh && yarn install && anchor build -p marginfi -- --no-default-features && anchor build -p mocks && anchor build -p liquidity_incentive_program -- --no-default-features && anchor test --skip-build",
+            "command": "./scripts/build-workspace.sh && yarn install && anchor build -p marginfi -- --no-default-features && anchor build -p mocks && anchor test --skip-build",
```
jgur-psyops, line 7: nit: if you haven't changed instruction args (this is usually the case for mocks) you can also add --no-idl to save around 5 seconds per build

programs/marginfi/src/state/price.rs
```diff
@@ -79,7 +77,6 @@ pub trait PriceAdapter {
 #[enum_dispatch(PriceAdapter)]
 #[cfg_attr(feature = "client", derive(Clone))]
 pub enum OraclePriceFeedAdapter {
-    PythLegacy(PythLegacyPriceFeed),
     SwitchboardV2(SwitchboardV2PriceFeed),
```
jgur-psyops, line 80: I think your approach removing the enum here is better, let's do the same for the switchboardv2 stub I left behind

programs/marginfi/tests/misc/risk_engine_flexible_oracle_checks.rs
```diff
@@ -63,27 +66,12 @@ async fn re_one_oracle_stale_failure() -> anyhow::Result<()> {
...
-    sol_bank
-        .update_config(
-            BankConfigOpt {
-                oracle_max_age: Some(200),
-                ..Default::default()
-            },
-            None,
-        )
-        .await?;
+    // Make SOLE feed not stale
```
jgur-psyops, line 74: nit: SOLE

IliaZyrin: Yeah, it's intentional here, for SOL Equivalent

PR #344, opened 2025-05-20
CLI resurrection

IliaZyrin:


jgur-psyops:
Looks good so far. I guess since we have added it back to the workspace the CI is automatically checking that it compiled without issue?

IliaZyrin:
Yep, I also checked it and installed. And briefly tested by running get `mfi bank get` commands - all worked.

PR #347, opened 2025-05-27
0.1.4 consolidated update

jgur-psyops:
## SUMMARY
* Adds a new way to transfer your account: `transfer_to_new_account` transfers all of your positions to a new account, with an authority of your choosing. This allows you to give your positions to others or move them between your own wallets. The process incurs a nominal fee in SOL (~$0.50 today) to avoid spam.
* Adds a bank cache, where consumers can view recent information about a bank's interest rates (APR), and interest recently earned.
* Removes Pyth legacy (aka v1, pythnet, or "push") oracle support from the program
* Fixes a bug with the switchboard oracles where confidence was incorrectly either 0 or clamped at 5%. All users participating in swb banks should see a tiny improvement to their account health and corresponding increase in borrowing power
* Oracle pricing is now totally suspended during periods of poor price confidence, where the allowed confidence is now configurable per bank. We expect this will reduce unfair liquidations during periods of extreme price volatility, but should not affect normal operation in the vast majority of cases. 
* Banks created after the release of 0.1.4 can now be closed by the group admin to recover rent, but only if they have no open positions, no emissions, and no assets. Banks created prior to 0.1.4 can never be closed.
* Banks now track the total number of open positions, subsequent to the release of 0.1.4 (Note: for new banks, this will always be 0 or greater, for banks created prior to 0.1.4, this may be negative as closed positions reduce the count).
* Adds delegate administrators for emissions, curve parameters, and deposit/borrow caps. These admins can modify only a small subset of settings, which makes them safe(r) to grant to a hot wallet or MS with limited signature threshold. The group admin is still free to modify most bank parameters.

### Consolidates:
* #349 (adds bank cache to record spot interest rates, etc)
* #343 (removes Pyth legacy support from the program and test suite)
* #344 (Restores Rust CLI functionality which was broken in 0.1.3)
* #350 (Tests usage of multiple staked collateral banks at once)
* #351 (minor TS error handling update)
* #353 (remove pyth legacy from test suite)
* #356 (fix swb std dev issue)
* #354 (new account migration instruction) 
* #355 (minor fixes following second emode audit suggestion)
* #356 (Fix usage of std dev aka confidence in swb oracles)
* #358 (Pyth accounts store the feed directly instead of feed id)
* #359 (Oracles fail when a confidence threshold is exceeded)
* #360 (Banks track open positions and can be closed)
* #361 (Delegate admins for setting various bank params)
* #369 (Minor fix to account transfer logic)

## BREAKING CHANGES
* Removes the old account migration instruction, which required the group admin (the multisig) to sign. Users can now migrate their balances to a new account at any time without admin approval using the new `transfer_to_new_account` ix
* If you were deriving the Pyth oracle using oracles[0] (formerly the feed id) and a shard id, this is no longer possible. Banks that have migrated (e.g. where `bank.config.config_flags` & 1 != 0`) now store the feed directly on oracles[0], use that key directly as the oracle instead.

## BREAKING (ADMIN ONLY)
* set/unset user flags ixes removed (all user flags are now deprecated)
* When configuring banks with pyth oracles, pass the oracle directly instead of the feed id. You no longer ever need to worry about the feed id.
* When configuring banks, an additional argument is passed (`oracle_max_confidence`). Oracle checks fail if this is exceeded. Pass 0 to use the default (10%).
* changing emissions now requires `delegate_emissions_admin` instead of regular `admin`. 

## NEW FIELDS AND OTHER NOTES
* Account authority is now immutable
* Frozen fields still can't update even with a delegated admin right, e.g. the delegate curve admin cannot do anything if a bank is frozen.
* bank.config and the config input types added field `oracle_max_confidence`
* bank added field bank.cache
* bank added field `lending_position_count` and `borrowing_position_count`
* Group gains `delegate_curve_admin`, `delegate_limit_admin`, and `delegate_emissions_admin`

## PYTH MIGRATION INSTRUCTIONS AND NOTES
* Banks that have not migrated will continue to function as normal until they migrate.
* Pyth banks that migrate will no longer store the feed id on oracles[0], instead they will store the key of the feed directly, just like switchboard oracles already do. After the ix completes, these banks will now accept that oracle and no other feeds.
* Simply run `migrate_pyth_push_oracle` for all banks that existed prior to 0.1.4 going live. Non-Pyth banks will set the migration flag and do nothing. Pyth banks (excluding staked collateral) will accept either the Pyth or Mrgn sponsored oracle feed. The feed must exist (e.g. not be a blank account).
* Staked collateral banks must use `propagate_staked_settings` instead. These will always use the Pyth sponsored oracle.
* Next time `edit_staked_settings` runs, remember to also update the oracle to use the SOL feed. This can happen any time in the future, it does not have to happen to complete migration, but propagation of future settings will fail until it's done. 
* New banks will now take the feed instead of the feed id when being created.
* Migration functionality will be removed in 0.1.5

### DIFF
--- a/guides/DEPLOY_GUIDE.md
+++ b/guides/DEPLOY_GUIDE.md
@@ -104,6 +104,7 @@ Here we list recent deployments to staging/mainnet. The hash is always the first
 * 0.1.2: March 14, 2025 ~3:00pm ET -- Hash 65bbbe
 * 0.1.3: April 29, 2025 ~2:00pm ET -- Hash 1d8130
 * 0.1.3 memory hotfix: May 1, 2025 ~4:00pm ET -- (no hash)
+* 0.1.4: July 25, 2025 ~7:00pm ET -- (no hash)
 
 ### MAINNET
 
@@ -111,3 +112,4 @@ Here we list recent deployments to staging/mainnet. The hash is always the first
 * 0.1.1: Feb 17, 2025 ~3:00pm ET -- Hash: 03455c
 * 0.1.2: April 14, 2025 ~1:00pm ET -- Hash 65bbbe
 * 0.1.3: May 27, 2025 ~1:00pm ET -- Hash ae9adb7
+* 0.1.4: July 28, 2025 ~1:00pm ET -- Hash 1229b8


LixvYang:
When the program is ready for consolidation?

LixvYang:
Hi, when do you plan to release it?

LixvYang:
> > Hi, when do you plan to release it?
> 
> #360 will be the last PR in 0.1.4. It will go to staging this week (roughly July 2-3) and then to mainnet 2-3 weeks after that.

Does the  mrgn-ts program need to be changed? I don't see any major changes in the active branch. https://github.com/mrgnlabs/mrgn-ts/branches/active

jgur-psyops:
> > > Hi, when do you plan to release it?
> > 
> > 
> > #360 will be the last PR in 0.1.4. It will go to staging this week (roughly July 2-3) and then to mainnet 2-3 weeks after that.
> 
> Does the mrgn-ts program need to be changed? I don't see any major changes in the active branch. https://github.com/mrgnlabs/mrgn-ts/branches/active

Our TS team has moved some commits to a private internal fork, but they will publish a package shortly after we go to staging today and at least a week before the mainnet launch ~2 weeks later.

IliaZyrin (dismissed):

PR #349, opened 2025-05-30
Bank cache

IliaZyrin:
Bank Cache contains spot interest rates and the interest accumulated since last update


programs/marginfi/src/state/bank_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+    Eq,
+    TypeLayout,
+    Debug,
+    Default,
+)]
+pub struct WrappedInterestRates {
+    pub base_rate: WrappedI80F48,
+    pub lending_rate: WrappedI80F48,
+    pub borrowing_rate: WrappedI80F48,
+    pub group_fee: WrappedI80F48,
```
jgur-psyops: Are these three just the base rate apr times the rate that's set in InterestRateConfig, e.g. `base_rate * (insurance_ir_fee) + insurance_fee_fixed_apr`? I think if that's all it is then we don't need to have fields for it since you already report base rate and the math is trivial.

programs/marginfi/src/state/bank_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+#[repr(C)]
+#[derive(
+    AnchorDeserialize, AnchorSerialize, Copy, Clone, Zeroable, Pod, PartialEq, Eq, TypeLayout, Debug,
+)]
+/// A read-only cache of the bank's key metrics, e.g. spot interest/fee rates.
+pub struct BankCache {
+    /// Actual interest/fee rates of the bank.
+    /// * APR (annual percentage rate) values
+    pub interest_rates: WrappedInterestRates,
+    pub accumulated_since_last_update: WrappedI80F48,
```
jgur-psyops, line 24: Maybe the time delta used by the last update would help here?

programs/marginfi/src/state/bank_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+    Zeroable,
+    Pod,
+    PartialEq,
+    Eq,
+    TypeLayout,
+    Debug,
+    Default,
+)]
+pub struct WrappedInterestRates {
+    pub base_rate: WrappedI80F48,
```
jgur-psyops: We can probably pack these into a smaller data type to save some space, f64 or f32 might work, but there's also probably some clever trick we could use since the base rate is always going to between 0 and 3ish (currently our max interest rate is usually set to 300%). One thing I did in the past was use a u32 where 0 =0 and u32::MAX = 100% so like 50% would be u32::MAX * 0.5, here we could do something similar where max is 300% (or do like 1000% in case we ever raise the max to simplify the math)

programs/marginfi/src/state/bank_cache.rs
IliaZyrin: Yep, it's these. Will remove them then!

programs/marginfi/src/state/bank_cache.rs
```diff
@@ -0,0 +1,79 @@
...
+use super::marginfi_group::WrappedI80F48;
+
+assert_struct_size!(BankCache, 160);
+assert_struct_align!(BankCache, 8);
+#[repr(C)]
+#[derive(
+    AnchorDeserialize, AnchorSerialize, Copy, Clone, Zeroable, Pod, PartialEq, Eq, TypeLayout, Debug,
+)]
+/// A read-only cache of the bank's key metrics, e.g. spot interest/fee rates.
+pub struct BankCache {
```
jgur-psyops, line 16: @abdulrabbani00 @k0beLeenders @chambaz If there's any other information that you think would be helpful to cache in the bank's state (this would update after any TX that affects bank funds) feel free to leave a comment.

abdulrabbani00: Nice! Nothing comes to mind right now, but if it does I'll be sure to shout it out prior to the next release.

Issue #348, opened 2025-05-29
Remove rent sysvar where not needed


jgur-psyops:
Remove `pub rent: Sysvar<'info, Rent>,` where it isn't required.

jgur-psyops:
Closed in #357

PR #357, opened 2025-06-14
removed rent field from LendingPoolAddBankPermissionless struct

viniciiii:
#348


jgur-psyops:
This is fine but basically we have `pub rent: Sysvar<'info, Rent>,` in a bunch of places all over the instructions where it is no longer needed since like Anchor 26+, so we want to remove all the occurrences.

PR #358, opened 2025-06-17
Switch to storing Pyth push oracle key

jgur-psyops:
## Summary
- define fixture for migrating legacy Pyth push banks
- add startup validator entry for the migration fixture
- create p02_pythMigrated.spec.ts showing borrow succeeds before and after migrating the oracle

## Testing
- `cargo fmt -- --check`
- `cargo test --workspace --no-fail-fast` *(failed: build interrupted)*

------
https://chatgpt.com/codex/tasks/task_b_6848933d95e4832392064d363900816d


programs/marginfi/src/instructions/marginfi_group/migrate_pyth_push_oracle.rs
```diff
@@ -0,0 +1,80 @@
+// Permissionless ix to migrate a Pyth-pull based bank created in 0.1.3 or earlier to the new oracle
```
Henry-E, line 1: In making this permissionless are you not worried that someone will push the "wrong" pyth pull oracle?

jgur-psyops: It should only accept two options: the Pyth sponsored oracle or the Mrgn sponsored oracle. Both are valid from our point of view, so the only risk here is one of these two is defunct and we wanted the other. In the event of that fairly unproductive troll, we would just send a config ix to fix it.

programs/marginfi/fuzz/src/lib.rs
```diff
@@ -216,7 +216,7 @@ impl<'state> MarginfiFuzzContext<'state> {
         let (_fee_state_key, fee_state_bump) =
             Pubkey::find_program_address(&[FEE_STATE_SEED.as_bytes()], &marginfi::id());
 
-        let (oracle, oracle_feed) = state.new_oracle_account(
+        let (oracle, _oracle_feed) = state.new_oracle_account(
```
IliaZyrin: As I see that's the only usage of this function, so let's just change the return type right away.

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -1301,7 +1301,12 @@ pub struct BankConfigCompact {
     /// other STAKED assets or SOL (`ASSET_TAG_SOL`) and can only borrow SOL
     pub asset_tag: u8,
 
-    pub _pad0: [u8; 6],
+    /// Flags for various config options
+    /// * 1 - Always set set if bank created in 0.1.4 or later, or if migrated to the new oracle
```
IliaZyrin:
```suggestion
    /// * 1 - Always set if bank created in 0.1.4 or later, or if migrated to the new oracle
```

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -1428,7 +1436,14 @@ pub struct BankConfig {
     /// other STAKED assets or SOL (`ASSET_TAG_SOL`) and can only borrow SOL
     pub asset_tag: u8,
 
-    pub _pad1: [u8; 6],
+    /// Flags for various config options
+    /// * 1 - Always set set if bank created in 0.1.4 or later, or if migrated to the new pyth
```
IliaZyrin:
```suggestion
    /// * 1 - Always set if bank created in 0.1.4 or later, or if migrated to the new pyth
```

programs/marginfi/src/state/price.rs
```diff
@@ -527,29 +601,47 @@ pub struct PythPushOraclePriceFeed {
...
 impl PythPushOraclePriceFeed {
-    /// Pyth push oracles are update using crosschain messages from pythnet
-    /// There can be multiple pyth push oracles for a given feed_id. Marginfi allows using any
-    /// pyth push oracle with a sufficient verification level and price age.
+    // TODO revise this doc in 0.1.5
+    /// Pyth push oracles are update using crosschain messages from pythnet There can be multiple
+    /// pyth push oracles for a given feed_id. Marginfi allows using any pyth push oracle with a
+    /// sufficient verification level and price age.
+    ///
+    /// In 0.1.3 and earlier, when loading the pyth push oracle, we don't verify the account address
```
IliaZyrin, line 609: Out of curiosity, why do we even put this information here? How could anyone interact with a version 0.1.3 once 0.1.4 is out?

tests/p01_initPythPull.spec.ts
```diff
@@ -8,7 +8,7 @@ import {
   verbose,
   ecosystem,
   oracles,
-  PYTH_ORACLE_SAMPLE,
+  // PYTH_ORACLE_SAMPLE, // no longer used. After 0.1.4, Pyth uses the feed directly.
```
IliaZyrin: Let's remove it then?

tests/s03_deposit.spec.ts
```diff
@@ -25,6 +25,7 @@ import { accountInit, depositIx } from "./utils/user-instructions";
 import { LST_ATA, USER_ACCOUNT } from "./utils/mocks";
 import { createMintToInstruction } from "@solana/spl-token";
 import { getBankrunBlockhash } from "./utils/spl-staking-utils";
+import { dumpBankrunLogs } from "./utils/tools";
```
IliaZyrin: Left-over from debug likely

IliaZyrin (approved):
Awesome!

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
```diff
@@ -24,6 +25,10 @@ pub fn lending_pool_configure_bank_oracle(
         bank.config.oracle_setup = setup_type;
         bank.config.oracle_keys[0] = oracle;
 
+        if setup_type == OracleSetup::PythPushOracle {
```
IliaZyrin, line 28: Actually this looks suspicious. Why are we setting this flag here?

programs/marginfi/src/state/price.rs
jgur-psyops: There's a TODO to remove this comment in 0.1.5 at the top of the function doc. For a while in 0.1.3 we will have some pre-migration banks that get validated the old way and migrated or 0.1.4+ banks that get validated the new way, so I guess the comment is more referring to pre/post migration than the version.

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
jgur-psyops: The idea is in 0.1.4 and later, if you try to reconfigure a bank's Pyth oracle, it will always validate the new way. You shouldn't be able to configure a bank using the old way. Especially important if someone accidentally tries configuring it with the old way after the migration flag is already set. 

By setting this flag we force `validate_bank_config` to take the `bank_config.is_pyth_push_migrated()` branch later

IliaZyrin: Makes sense!

IliaZyrin (approved):

programs/marginfi/src/instructions/marginfi_group/migrate_pyth_push_oracle.rs
```diff
@@ -0,0 +1,85 @@
...
+    let (pyth_feed, _) =
+        PythPushOraclePriceFeed::find_oracle_address(PYTH_SPONSORED_SHARD_ID, &feed_id_bytes);
+    let (mrgn_feed, _) =
+        PythPushOraclePriceFeed::find_oracle_address(MARGINFI_SPONSORED_SHARD_ID, &feed_id_bytes);
+    let oracle_key = ctx.accounts.oracle.key();
+
+    // On localnet, the feed id is randomly generated, so this check is not applicable.
+    if live!() {
+        require!(
+            oracle_key == pyth_feed || oracle_key == mrgn_feed,
```
Henry-E, line 49: It's probably to do with weird naming but it is confusing from the outside that we're checking the `oracle_key` is equal to the `feed address` when the purpose of the PR is migrate away from storing the feed address in the oracle keys list. This is more to do with my unfamiliarity with the pyth oracle keys + feed address system

jgur-psyops: This is more on Pyth for calling the oracles "feeds" that have a "feed id" while we call them "oracle". The language confuses me too idk what's a better name

PR #359, opened 2025-06-19
Oracle conf limits

jgur-psyops:
When confidence bands are especially large, we end up using a price that is unreliable. For very wide confidence bands, it is better to simply abort all transactions until a better price is available. Banks can now configure a % of price, and if confidence exceeds that, transactions will abort. 

## BREAKING (everyone)
* None

## BREAKING (admin only)
* config_bank and add_bank instructions now have an extra field (oracle_max_confidence). This is a u32 representing 0-100%, e.g. 50% is u32::MAX/2, where the oracle value is considered invalid if price * oracle_max_confidence > reported confidence.

### DIFF
--- a/programs/marginfi/src/state/price.rs
+++ b/programs/marginfi/src/state/price.rs
@@ -3,6 +3,7 @@ use crate::{
     constants::{
         CONF_INTERVAL_MULTIPLE, EXP_10_I80F48, MAX_CONF_INTERVAL, MIN_PYTH_PUSH_VERIFICATION_LEVEL,
         NATIVE_STAKE_ID, PYTH_ID, SPL_SINGLE_POOL_ID, STD_DEV_MULTIPLE, SWITCHBOARD_PULL_ID,
+        U32_MAX, U32_MAX_DIV_10,
     },
     debug, live, math_error,
     prelude::*,
@@ -16,7 +17,7 @@ use enum_dispatch::enum_dispatch;
 use fixed::types::I80F48;
 use pyth_solana_receiver_sdk::price_update::{self, FeedId, PriceUpdateV2};
 use pyth_solana_receiver_sdk::PYTH_PUSH_ORACLE_ID;
-use std::{cell::Ref, cmp::min};
+use std::{cell::Ref, cmp::min, u32};
 use switchboard_on_demand::{
     CurrentResult, Discriminator, PullFeedAccountData, SPL_TOKEN_PROGRAM_ID,
 };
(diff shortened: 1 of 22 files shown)


jgur-psyops:
Open question for @Henry-E and @IliaZyrin : should we enable liquidations to continue to go through with a wider band? I was considering passing oracle_max_confidence * 2 or similar to the liquidation function. This could unfairly liquidate some users during price unreliability, but sustain solvency.

programs/marginfi/src/state/price.rs
```diff
@@ -412,17 +422,35 @@ impl SwitchboardPullPriceFeed {
...
+            .checked_div(U32_MAX)
+            .ok_or_else(math_error!())?;
+        if conf_interval > max_conf {
+            let conf_interval = conf_interval.to_num::<f64>();
+            let max_conf = max_conf.to_num::<f64>();
+            msg!("conf was {:?}, but max is {:?}", conf_interval, max_conf);
+            return err!(MarginfiError::OracleMaxConfidenceExceeded);
+        }
+
+        // Clamp confidence to 5% of the price regardless
```
IliaZyrin, line 453: I guess it's been like this for ages but now with `oracle_max_confidence` in place it seems more random - why do we choose to use the less confident price (5% interval) even though the oracle (potentially) states it's not that confident (e.g. 10%)?

Shouldn't we just always rely on the reported confidence?

tests/e04_emodeLiquidation.spec.ts
```diff
@@ -519,7 +519,7 @@ describe("Emode liquidation", () => {
 
     let tx = new Transaction().add(
       ComputeBudgetProgram.setComputeUnitLimit({
-        units: 210_000,
+        units: 260_000,
```
IliaZyrin, line 522: Creeping up, huh?=)

tests/utils/types.ts
```diff
@@ -128,20 +131,21 @@ export const defaultBankConfigOptRaw = () => {
     assetWeightMaint: I80F48_ONE,
     liabilityWeightInit: I80F48_ONE,
     liabilityWeightMaint: I80F48_ONE,
-    depositLimit: new BN(1_000_000_000),
-    borrowLimit: new BN(1_000_000_000),
+    depositLimit: new BN(1000000000),
```
IliaZyrin: IMHO it was better readable with underscores.

programs/marginfi/src/state/price.rs
IliaZyrin: Also, regarding [this](https://github.com/mrgnlabs/marginfi-v2/pull/359#issuecomment-2989114951). I think we can keep it as is, just to not complicate this whole "confidence" logic even more. Liquidator should be able to sustain solvency anyway. It's fast.

PR #360, opened 2025-07-01
Add bank position tracking & closure tests

jgur-psyops:
## Summary
Enables future banks to close when they no longer have any active users, recovering rent. Banks created prior to this update will still not be able to close. Also enables at-a-glance viewing of many users are active in each bank.

- track lending, borrowing, and total positions for each bank
- allow closing banks only when no positions exist
- update admin regression tests for new padding
- assert position counts in deposit/borrow/withdraw flows
- add a close bank e2e test

## NOTES
* Banks now have two new fields, `lending_position_count` and `borrowing_position_count`. For banks created before this update goes live, the count will not be canonical, and could be negative! For banks created after this update, the count is canonical and cannot be lower than 0. 

------
https://chatgpt.com/codex/tasks/task_b_685c3ca6ebf08323b70433eee696583b


programs/marginfi/src/instructions/marginfi_group/close_bank.rs
```diff
@@ -0,0 +1,60 @@
...
+    state::marginfi_group::{Bank, MarginfiGroup},
+    MarginfiError, MarginfiResult,
+};
+use anchor_lang::prelude::*;
+use fixed::types::I80F48;
+
+pub fn lending_pool_close_bank(ctx: Context<LendingPoolCloseBank>) -> MarginfiResult {
+    let bank = ctx.accounts.bank.load()?;
+
+    // banks created prior to 0.1.4 can never be closed because we cannot gaurantee an accurate
```
IliaZyrin:
```suggestion
    // banks created prior to 0.1.4 can never be closed because we cannot guarantee an accurate
```

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -532,8 +534,20 @@ pub struct Bank {
...
+    ///   the bank may safely be closed if this is zero. Will never go negative.
+    pub lending_position_count: i32,
+    /// Number of user borrowing positions currently open in this bank
+    /// * For banks created prior to 0.1.4, this is the number of positions opened/closed after
+    ///   0.1.4 goes live, and may be negative.
+    /// * For banks created in 0.1.4 or later, this is the number of positions open in total, and
+    ///   the bank may safely be closed if this is zero. Will never go negative.
+    pub borrowing_position_count: i32,
+    pub _padding_0: [u8; 16],
+    pub _padding_1: [[u64; 2]; 19], // 8 * 2 * 29 = 464B
```
IliaZyrin:
```suggestion
    pub _padding_1: [[u64; 2]; 19], // 8 * 2 * 19 = 304B
```

type-crate/src/types/bank.rs
```diff
@@ -76,9 +76,11 @@ pub struct Bank {
...
     /// manually (withdraw_fees).
     pub fees_destination_account: Pubkey, // 32
-
-    pub _padding_0: [u8; 8],
-    pub _padding_1: [[u64; 2]; 30], // 8 * 2 * 30 = 480B
+    pub lending_position_count: i32,
+    pub borrowing_position_count: i32,
+    pub position_count: i32,
+    pub _padding_0: [u8; 12],
+    pub _padding_1: [[u64; 2]; 29], // 8 * 2 * 29 = 464B
```
IliaZyrin: This is not aligned with the "main" struct.

IliaZyrin (approved):
Nice!

Henry-E:
Where's the demand for this feature coming from?

Some thoughts
- Why bother allowing banks to be closed only when there's only 0 positions open. This is likely never to be the case. Lots of people forget about accounts or leave dust in a bank essentially forever, even on withdraw-only banks.
- Counting the number of open positions seems like bloat. It's yet another thing to make sure you're not messing up when changing anything inside of bank. Isn't this something that backend tracks?

tests/utils/types.ts
```diff
@@ -26,6 +26,7 @@ export const EMISSIONS_FLAG_BORROW_ACTIVE = 1;
 export const EMISSIONS_FLAG_LENDING_ACTIVE = 2;
 export const PERMISSIONLESS_BAD_DEBT_SETTLEMENT_FLAG = 4;
 export const FREEZE_SETTINGS = 8;
+export const CLOSE_ENABLED_FLAG = 16;
```
Henry-E, line 29: Is this the same set of flags that also track whether pyth push oracle has been migrated?

jgur-psyops: no, those are in the config (due to how they are consumed in the various risk functions)

Henry-E: ok, so there are multiple flags on the bank but some are stored at the top level and some inside the config?

jgur-psyops: Yep, this is annoying but (1) somebody put the emissions flags on the "main" bank struct so we can't move those, (2) the bank doesn't get passed to all the price/risk functions, for example `try_from_bank_config_with_max_age` gets just the config. It seemed easier to add flags to the config than to refactor these to take a bank.

PR #362, opened 2025-07-02
Bankruptcy

IliaZyrin:
### Bankruptcy handling: logic, tests, and docs
**What this does**

- Implements full bankruptcy flow for users and banks, including super-bankruptcy handling (clamps loss and kills the bank when bad debt exceeds assets). Introduces a dedicated error code to reflect this terminal state.

- Adds a user & bank bankruptcy guide (how it triggers, discharge order, and outcomes): [BANKRUPTCY.md](https://github.com/mrgnlabs/marginfi-v2/pull/362/files#diff-cfdf47a7e9198a056f558d4a6f30d1d298556e9d4d7fb4c55326fae75ea581cb).

- Hardens instruction paths with additional bank state validation (e.g., during borrow) to prevent actions on paused/reduced/killed banks.

**Notable changes**

- New/updated program errors
BankKilledByBankruptcy (code 6082).
Quick lookup: [BankKilledByBankruptcy](https://github.com/mrgnlabs/marginfi-v2/pull/362/files#diff-0064e458227eb9e25e5169103580b44197b661620bc1c3b459a40c925c2e70ceR274).

- Instruction changes

[handle_bankruptcy](https://github.com/mrgnlabs/marginfi-v2/pull/362/files#diff-ce6f55e971f2e5eaf4b3eaf903c21a79c272be88ce7c277ff2bf34c32959bc4aR178) logic (bank-level settlement, socialized losses, kill path on super-bankruptcy).

[borrow](https://github.com/mrgnlabs/marginfi-v2/pull/362/files#diff-c3a031b689deedae717c14b8c6f00436d23883d6cf5f2f075f9a88bf109896d2) as well as the other ixes (deposit repay, withdraw) now validates that the bank is not in a paused or reduced state.

- Tests & fuzz

New spec: tests/zb01_bankruptcy.spec.ts (see [Files changed](https://github.com/mrgnlabs/marginfi-v2/pull/362/files#diff-6ca38bc4f719d6b77099a793a240bc54de6ae30ff9864de8cd946f13060f3d5a)).

Fuzz harness updates and additional logging/assertions to surface unexpected liquidation errors.

**How it works (short)**

- Liquidation exhausts remaining assets of the insolvent user.

- handle_bankruptcy settles bad debt:

Use insurance if sufficient.

Otherwise, socialize via asset-share value reduction.

If still insufficient (super-bankruptcy), kill the bank (asset share value to zero; bank permanently disabled).

**Test plan**

- Unit/integration tests in zb01_bankruptcy.spec.ts cover user bankruptcy, bank bankruptcy, and super-bankruptcy kill path.

- Fuzz tests updated to assert allowed error surface and reject unexpected outcomes; reproducible logs added for triage.


IliaZyrin:
@jgur-psyops let's merge this one?

programs/marginfi/src/utils.rs
```diff
@@ -255,6 +255,61 @@ pub fn validate_bank_asset_tags(bank_a: &Bank, bank_b: &Bank) -> MarginfiResult
...
+    /// E.g. liquidation
+    FailsInPausedState,
+    /// E.g. borrow, deposit
+    FailsIfPausedOrReduceState,
+}
+
+// TODO remove redundant checks for these elsewhere in the program (they are nested many laters deep
+// in various value delta functions)
+/// Validate the bank's state does not forbid the execution of an instruction
+pub fn validate_bank_state(bank: &Bank, kind: InstructionKind) -> MarginfiResult {
```
Henry-E, line 273: How is this different from `assert_operational_mode`?

Henry-E: So the high level idea is to enforce which states a bank is allowed to be in when an instruction is run? Since this wasn't really enforced before, except implicitly by `assert_operational_mode`?

Maybe it's a nit pick but I wonder if the positive is easier to understand
`bank.load().allowed_states([PAUSED, REDUCE])`
and we can put that in the constraints above the bank's account in a given instruction context.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -249,39 +249,14 @@ pub fn lending_account_liquidate<'info>(
...
-                        &mut liquidator_marginfi_account.lending_account,
-                    )?;
-                    bank_account.borrow_ignore_borrow_cap(remaining)?;
-                }
-            } else {
-                // Common case: liquidator has debt, or this is a new borrow
-                bank_account.borrow_ignore_borrow_cap(liab_amount_liquidator)?;
-            }
+            // Liquidator will withdraw the collateral (if any) and then borrow the remainder (if any). Ignores the utilization ratio check.
+            bank_account.withdraw_ignore_borrow_cap(liab_amount_liquidator)?;
```
jgur-psyops, line 259: This is what it already had originally, and I removed it in favor of the "full" logic that doesn't involve flipping, because having this flip logic handled internally is seriously hard to follow (for me). Also I think to fix the event emit later, we will need the full logic: it should emit differently for a withdraw vs a withdraw-borrow that flips.

Well if you prefer the flipping version (it's certainly shorter!) then no objection on my part.

IliaZyrin: For me personally it is easier to follow like this, because the logic of the liquidation itself is shorter and thus more digestible. Plus it allowed us to not narrow the usage of "bypass_util_ratio_check" cleanly. Regarding the events, good point, but I think we will sort it out.

Henry-E:
Nothing left to say on this

Henry-E (approved):
LGTM

PR #364, opened 2025-07-08
Type crate usage

IliaZyrin:
This PR moves the state types (shared Anchor accounts and their AnchorSerializable sub-parts) from the program crate to the `type-crate`.

`type-crate` is from now on the only source of truth for the **types**, their **basic implementation** (e.g. traits like `Default` or `new()` constructors) and the **constants** associated with the MarginFi program.

`type-crate` can be built with or without the dependency on Anchor. It is under the optional feature named `anchor`. **Program uses it with this feature and shares the same Solana Program ID.** For this a separate crate was created - `id-crate` - whose sole purpose is [to declare the ID](https://github.com/mrgnlabs/marginfi-v2/pull/364/files#diff-1a7a77df11702a34bae7ff5051e2ca805ce33abc98cb31649333bea6404b9677) based on the cluster type and expose it for the program crate and `type-crate`.

`type-crate` can also be built without the `anchor` feature and in this case will use stubs instead of the genuine `Anchor` types (e.g. `Pubkey`). It will also lack a few traits on the provided types. **This build is aimed for the integrators** who do not want to import the whole program code but just need the main types and a handful of utility functions around them.

**Code conventions for the future:**
1. If you introduce a new state type, put it directly to the `type-crate`. Make sure it compiles in both "with-anchor" and "without-anchor" ways. Follow the same pattern as for the existing types - put Anchor-related traits and attributes under the `anchor` feature.
2. If you introduce a new constant or function, think if it can be useful for the 3rd party integrators and if the answer is yes, put it to the `type-crate`. If it's only for the internal testing/logic, keep it inside the program crate.
If you decide to implement a method on a type defined in the `type-crate`, introduce a new `pub trait` and [implement it there](https://github.com/mrgnlabs/marginfi-v2/pull/364/files#diff-abc8042a18167b6f9e6511f9d2b32a33eae915f3377aa54d673a8d5f9501ba62R22). Otherwise it will be rejected by [the "orphan rule"](https://ianbull.com/notes/rusts-orphan-rule/). Or you can simply use [good old functions](https://github.com/mrgnlabs/marginfi-v2/pull/364/files#diff-afba2f29893e09b072763ee7c7a01aed16565ec99ed333ea679710ed41746700R4).


IliaZyrin:
> As I mentioned in the call I don't fully see the reason for this. While generally I'm behind the idea of eventually splitting up anchor programs into CPI, program and type crates, I don't know if there's really enough demand or if this is the right way to do it.
> 
> * Most demand for pure types is already fulfilled by the IDL, which can be converted into a dependencyless type crate. Abdul said he didn't need it and for the liquidator it was mentioned that even with this type crate some of the health cache logic from the main program crate still needs to be imported.
> * This type crate doesn't include any of the helper methods which would presumably be part of why someone might conceivably want to import a type crate.

So basically this is just the first iteration. The goal is to do what you describe here - migrate helper methods (and health cache logic in particular) so that the consumers (e.g. Eva) do not depend on the main program. That PR is just a ground work for this and is already pretty big, that's why it doesn't include these changes.

PR #366, opened 2025-07-16
0.1.5 consolidated upgrade

jgur-psyops:
WIP

Consolidates:

* #362 (Bankruptcy handling improvements for super-bankrupt banks)
* #364 (Broad refactor)


Henry-E:
Probably doesn't matter but
- [ ] `emission.rs` is missing protocol paused check above marginfi group account

programs/marginfi/Cargo.toml
```diff
@@ -18,12 +18,14 @@ idl-build = ["anchor-lang/idl-build", "anchor-spl/idl-build"]
...
-devnet = []
-mainnet-beta = []
 debug = []
-staging = []
+mainnet-beta = ["id-crate/mainnet-beta", "marginfi-type-crate/mainnet-beta"]
+staging = ["id-crate/staging", "marginfi-type-crate/staging"]
+devnet = ["id-crate/devnet", "marginfi-type-crate/devnet"]
 
 [dependencies]
+id-crate = { path = "../../id-crate" }
```
Henry-E: nit: To unify with how all of the other dependencies do it, we should define this at the top level and use `workspace = true` if possible. @IliaZyrin

jgur-psyops: I wanna say we did this to optimize build speed so a workspace build doesn't always trigger this (since even a blank rust crate takes like 5s). There's a similar PR up to do this for the test-utils so they only build if you run the rust test suite.

IliaZyrin: I don't think we gain much from excluding it from the workspace dependencies though, since it's anyway necessary for pretty much every target (marginfi, fuzz, test-utils, kamino-mocks).

IliaZyrin: [Done](https://github.com/mrgnlabs/marginfi-v2/pull/394/commits/a479919950f33e030b267808bf3a0859c1805ee1) in Kamino branch.

PR #369, opened 2025-07-25
Fix transfer_to_new_account double run

jgur-psyops:
## Summary
- add `migrated_to` field on `MarginfiAccount`
- prevent repeated `transfer_to_new_account`
- expose new error `AccountAlreadyMigrated`
- update regression tests and transfer test for new field

## Testing
- `cargo fmt --all`
- ❌ `cargo test --package marginfi marginfi_account_transfer_happy_path -- --nocapture --exact` *(failed: lock file version requires nightly)*

------
https://chatgpt.com/codex/tasks/task_b_6883d39920e48323b6b7e8e6e7dcc211

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
@@ -20,12 +20,21 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
 
     let mut old_account = ctx.accounts.old_marginfi_account.load_mut()?;
 
+    // Prevent multiple migrations from the same account
+    check_eq!(
+        old_account.migrated_to,
+        Pubkey::default(),
+        MarginfiError::AccountAlreadyMigrated
+    );
+
     let mut new_account = ctx.accounts.new_marginfi_account.load_init()?;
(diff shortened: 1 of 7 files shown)


programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -20,13 +20,22 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
 
     let mut old_account = ctx.accounts.old_marginfi_account.load_mut()?;
 
+    // Prevent multiple migrations from the same account
```
Henry-E, line 23: Is an account migration different from an account transfer? 

Just trying to understand if the intention is to only allow an account to transferred once from one owner to another?

Henry-E: Ok, i think i get. Migration transfers all the positions from one account to another. It doesn't transfer ownership of the account itself.

Henry-E (commented):
Prevents repeat migrations, makes sense. Code changes mostly only involve storing extra data and adding a couple of extra checks. Looks safe enough.

IliaZyrin (approved):

PR #370, opened 2025-07-29
Remove pyth migration

IliaZyrin:


programs/marginfi/src/instructions/marginfi_group/add_pool_permissionless.rs
```diff
@@ -99,7 +99,6 @@ pub fn lending_pool_add_bank_permissionless(
     );
     bank.config.oracle_setup = OracleSetup::StakedWithPythPush;
     bank.config.oracle_keys[0] = settings.oracle;
-    bank.config.update_config_flag(true, PYTH_PUSH_MIGRATED);
```
jgur-psyops, line 102: I think we can keep this (staked collateral banks always have the flag set in 0.1.4 or later). Or we set it on StakedSettings and they inherit it through this or propagate.

jgur-psyops:
Reminder to migrate all staging banks: 

- [ ] Done

Reminder to migrate all arena banks:

- [ ] Done

programs/marginfi/src/instructions/marginfi_group/add_pool_permissionless.rs
IliaZyrin: But it is set anyway [here](https://github.com/mrgnlabs/marginfi-v2/pull/370/files/4b90cb6c55fe3a7525cb26d6f18b56295d5a04e9#diff-e88087d7e0383e4d74b4335f08f449bc31de65105e7f9fa7bd5f2993e7abecefL75).

Issue #340, opened 2025-05-14
Timestamp last account change


jgur-psyops:
Currently we timestamp the last update to each balance, backend would like a single field that tracks the most recent update overall so they don’t have to compare for each balance. When they take a state snapshot, they will see if the last update overall is sufficiently recent. This should also include if the emissions account or authority changes.

PR #373, opened 2025-07-31
last_update for user acc

IliaZyrin:
Closes https://github.com/mrgnlabs/marginfi-v2/issues/340

### DIFF
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -1439,8 +1439,7 @@ mod test {
             migrated_from: Pubkey::default(),
             migrated_to: Pubkey::default(),
             last_update: 0,
-            _padding0: [0; 16],
-            _padding0: [0; 13],
+            _padding0: [0; 12],
         };
 
         assert!(acc.get_flag(ACCOUNT_TRANSFER_AUTHORITY_DEPRECATED));
--- a/programs/marginfi/tests/misc/regression.rs
+++ b/programs/marginfi/tests/misc/regression.rs
@@ -46,7 +46,9 @@ async fn account_field_values_reg() -> anyhow::Result<()> {
     assert_eq!(account.account_flags, 0);
     // health cache doesn't exist on these old accounts, but it also doesn't matter since it's read-only
     assert_eq!(account.health_cache, HealthCache::zeroed());
-    assert_eq!(account._padding0, [0; 13]);
+    assert_eq!(account.migrated_from, Pubkey::default());
+    assert_eq!(account.migrated_to, Pubkey::default());
+    assert_eq!(account._padding0, [0; 12]);
 
     let balance_1 = account.lending_account.balances[0];
     assert!(balance_1.is_active());
@@ -124,8 +126,7 @@ async fn account_field_values_reg() -> anyhow::Result<()> {
     );
     assert_eq!(account.account_flags, 0);
     assert_eq!(account.last_update, 0);
-    assert_eq!(account._padding0, [0; 13]);
-
+    assert_eq!(account._padding0, [0; 12]);
 
     let balance_1 = account.lending_account.balances[0];
     assert!(balance_1.is_active());
@@ -203,7 +204,7 @@ async fn account_field_values_reg() -> anyhow::Result<()> {
     );
     assert_eq!(account.account_flags, 0);
     assert_eq!(account.last_update, 0);
-    assert_eq!(account._padding0, [0; 13]);
+    assert_eq!(account._padding0, [0; 12]);
 
     let balance_1 = account.lending_account.balances[0];
     assert!(!balance_1.is_active());
--- a/type-crate/src/types/user_account.rs
+++ b/type-crate/src/types/user_account.rs
@@ -48,7 +48,7 @@ pub struct MarginfiAccount {
     /// If this account has been migrated to another one, store the destination account key
     pub migrated_to: Pubkey, // 32
     pub last_update: u64,
-    pub _padding0: [u64; 13],
+    pub _padding0: [u64; 12],
 }
 
 impl MarginfiAccount {


programs/marginfi/src/instructions/marginfi_account/deposit.rs
```diff
@@ -99,6 +99,7 @@ pub fn lending_account_deposit<'info>(
     )?;
 
     bank_account.deposit(I80F48::from_num(deposit_amount))?;
+    marginfi_account.last_update = bank_account.balance.last_update;
```
jgur-psyops: nit: this syntax using the balance's state is a little harder to understand than just clock.unix_timestamp (as above in accrue interest), since the balance is updated a couple layers deeper.

programs/marginfi/src/instructions/marginfi_account/repay.rs
```diff
@@ -71,6 +71,11 @@ pub fn lending_account_repay<'info>(
 
         amount
     };
+    marginfi_account.last_update = if repay_all {
```
jgur-psyops: here's it's probably easier to use clock time always rather than the balance conditionally

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -24,7 +24,11 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
     let mut old_account = ctx.accounts.old_marginfi_account.load_mut()?;
 
     let mut new_account = ctx.accounts.new_marginfi_account.load_init()?;
-    new_account.initialize(old_account.group, ctx.accounts.new_authority.key());
+    new_account.initialize(
+        old_account.group,
+        ctx.accounts.new_authority.key(),
+        Clock::get()?.unix_timestamp as u64,
+    );
```
jgur-psyops, line 44: I guess we should update the old account here too, though it's not particularly important

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -96,6 +96,12 @@ pub fn lending_account_withdraw<'info>(
             amount_pre_fee
         };
 
+        marginfi_account.last_update = if withdraw_all {
```
jgur-psyops: same as for repay, might as well always use the clock

programs/marginfi/tests/admin_actions/account_transfer.rs
```diff
@@ -35,6 +44,8 @@ async fn marginfi_account_transfer_happy_path() -> anyhow::Result<()> {
     assert_eq!(account_new.authority, new_authority.pubkey());
     // Old account is recorded as the migration source
     assert_eq!(account_new.migrated_from, marginfi_account.key);
+    // last_update only actualized for the new account - old one stays the same
```
jgur-psyops: again should probably update the old account too, but no strong feelings either way tbh

programs/marginfi/tests/user_actions/liquidate.rs
```diff
@@ -257,32 +277,78 @@ async fn marginfi_account_liquidation_success(
...
+            .unwrap()
+    };
+
     let collateral_mint_liquidator_balance = collateral_bank.get_asset_amount(
         liquidator_mfi_ma.lending_account.balances[liquidator_collateral_index]
             .asset_shares
             .into(),
     )?;
-    assert_eq!(
+    assert_eq_noise!(
```
jgur-psyops: is it no longer precise due to the 1 second of interest?

programs/marginfi/tests/user_actions/liquidate.rs
```diff
@@ -257,32 +277,78 @@ async fn marginfi_account_liquidation_success(
...
+        collateral_mint_liquidator_balance,
+        1.
     );
 
     let debt_paid_out = liquidate_amount * 0.975 * collateral_bank_f.get_price().await
         / debt_bank_f.get_price().await;
+    let debt_paid_out_native =
+        I80F48::from(native!(debt_paid_out, debt_bank_f.mint.mint.decimals, f64));
+
+    // We have to account for updated share value here due to 1 second of delay we introduced to check last_update
```
jgur-psyops: it might be better to validate the timing update in a separate test and keep the old one with precise assertions too. better to have both imo.

programs/marginfi/src/instructions/marginfi_account/deposit.rs
IliaZyrin: This is not about the syntax here. It's about consistency between the two. It seems logical to me to have `last_update` exactly match the last update of the balance.

programs/marginfi/src/instructions/marginfi_account/repay.rs
IliaZyrin: Again, the reasoning was to match both `last_update`s if possible (when there is no balance, it's apparently impossible, thus using the clock time).

programs/marginfi/tests/user_actions/liquidate.rs
IliaZyrin: Ok, I don't mind.

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
IliaZyrin: Actually I agree, not sure why I decided that way. But now that we introduced a new field - `migrated_to`, it makes even more sense. So will do!

programs/marginfi/tests/user_actions/liquidate.rs
IliaZyrin: Seems so, but it's a bit weird since the difference is very very small:
`999999.999619483048**72** and 999999.999619483048**71**5`

programs/marginfi/src/instructions/marginfi_account/deposit.rs
IliaZyrin: @Henry-E what do you think about it?

Henry-E: That's a good question. The numbers will still be the same though right? `balance.last_update` ought to be the `clock.unix_timestamp` since that number is fixed for the whole block. Unless there's some situation where those two number don't match because there was no update? 

Either way I disprove of adding more updates into the instruction function body.

programs/marginfi/src/instructions/marginfi_account/repay.rs
Henry-E: Ok, so what Ilia's getting at is that `last_update` might not be the current block's timestamp. Is that correct?

Henry-E:
This functionality is already available via RPC. `getSignaturesForAddress` + `getTransaction` can be used to construct the historical changes to an account.

programs/marginfi/src/instructions/marginfi_account/repay.rs
IliaZyrin: Yeah, but now that I looked into it again, it seems it's impossible for them to be different. I initially thought [this thing](https://github.com/mrgnlabs/marginfi-v2/blob/main/programs/marginfi/src/state/marginfi_account.rs#L1425) can differ, but [it is in fact unconditional](https://github.com/mrgnlabs/marginfi-v2/blob/main/programs/marginfi/src/state/marginfi_account.rs#L1473) down the line.

So Jon is right, it should be easier to just always set it to the current timestamp. I will make the change since we're aiming for simpler logic wherever possible.

IliaZyrin:
> This functionality is already available via RPC. `getSignaturesForAddress` + `getTransaction` can be used to construct the historical changes to an account.

I think it's 1) more complex, 2) potentially incorrect in the cases where there was _a change_ in the account data but we do not consider it _an update_.

Henry-E (approved):
LGTM

PR #377, opened 2025-08-08
Better security policy

abdulrabbani00:


jgur-psyops (approved):
lgtm. I would give a pay band for ui/infra bugs too, minor -$50, medium-$50-$500, high $500-$1500, critical up to $10000

PR #379, opened 2025-08-13
PDA-based mrgn accounts #367

IMEF-FEMI:
### ✅ Completed Features
  - **PDA account creation** with seeds: `[group, authority, account_index, third_party_id]`
  - **Account transfer** from keypair to PDA accounts
  - **Third-party ID restrictions** - ID=42 limited to mocks program CPI
  - **Backward compatibility** - existing keypair methods still work
  - **Comprehensive tests** - both Rust and TypeScript coverage
  - **Mock CPI tests** -- for both rush and TS

  ### 📋 Implementation Details
  - `initialize_account_pda()` instruction
  - `transfer_to_new_account_pda()` instruction
  - PDA derivation in `MarginfiAccount::derive_pda()`

PR #382, opened 2025-08-25
User Accounts with PDA (internal)

jgur-psyops:
Resolving the various merge conflicts since 0.1.4 here for consolidation into the 0.1.5 update.

For overview, see the original: #379 . Thanks for https://github.com/IMEF-FEMI for the contribution

Edits:
- Moved some appropriate portions over the type-crate.
- Added the flashloan check the "new" PDA-enabled transfer.


jgur-psyops:
> Glad i could be of help @jgur-psyops.... and sorry i missed the type-crate and flash loan checks

Not your fault, just something we changed recently! Thanks again, we should get this in production with 1.5 in mid September or so.

programs/marginfi/src/constants.rs
```diff
@@ -56,3 +64,67 @@ pub const MIN_PYTH_PUSH_VERIFICATION_LEVEL: VerificationLevel = VerificationLeve
 /// * Should be ~ $0.50 or around that magnitude
 /// * In lamports
 pub const ACCOUNT_TRANSFER_FEE: u64 = 5_000_000;
+
+/// When creating a mrgn account using a PDA, programs that wish to specify a third_party_id must be
+/// registered here. This confers no other benefits. Creating accounts with third_party_id = 0 or
+/// (the default) or PDA_FREE_THRESHOLD < PDA_FREE_THRESHOLD is freely available to any caller.
```
IliaZyrin:
```suggestion
/// registered here. This confers no other benefits. Creating accounts with third_party_id = 0
/// (the default) or ID < PDA_FREE_THRESHOLD is freely available to any caller.
```

programs/marginfi/src/constants.rs
```diff
@@ -56,3 +64,67 @@ pub const MIN_PYTH_PUSH_VERIFICATION_LEVEL: VerificationLevel = VerificationLeve
...
+    // (10_003, YET_ANOTHER_PROGRAM_ID),
+];
+
+/// * IDs < PDA_FREE_THRESHOLD are "free" (no special CPI restriction), just go ahead and use them
+///
+/// * IDs >= PDA_FREE_THRESHOLD are "restricted": must contact us to register first.
+pub const PDA_FREE_THRESHOLD: u16 = 10_000;
+
+// TODO move to ix_utils after liquidation_remix merged into 0.1.5
+/// Numbers > PDA_FREE_THRESHOLD are restricted, contact us to secure one.
```
IliaZyrin: Duplicating the above comment.

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -106,3 +110,127 @@ impl<'info> TransferToNewAccount<'info> {
...
+        ctx.accounts.global_fee_wallet.key(),
+        group.fee_state_cache.global_fee_wallet,
+        MarginfiError::InvalidFeeAta
+    );
+    anchor_lang::system_program::transfer(ctx.accounts.transfer_fee(), ACCOUNT_TRANSFER_FEE)?;
+
+    let mut old_account = ctx.accounts.old_marginfi_account.load_mut()?;
+
+    check!(
+        !old_account.get_flag(ACCOUNT_IN_FLASHLOAN),
```
IliaZyrin, line 136: Just a placeholder for us to not forget to add the same check for receivership liquidation here.

IliaZyrin (approved):
Looks solid!

type-crate/src/constants.rs
```diff
@@ -15,6 +15,8 @@ pub const STAKED_SETTINGS_SEED: &str = "staked_settings";
 pub const EMISSIONS_AUTH_SEED: &str = "emissions_auth_seed";
 pub const EMISSIONS_TOKEN_ACCOUNT_SEED: &str = "emissions_token_account_seed";
 
+pub const MARGINFI_ACCOUNT_SEED: &str = "marginfi_account";
```
Henry-E, line 18: nit but if you use `MARGINFI_ACCOUNT_SEED: &[u8] = b"marginfi_account"` then there's no conversion needed when using the const in the seeds `[MARGINFI_ACCOUNT_SEED]`

programs/marginfi/src/instructions/marginfi_account/initialize.rs
```diff
@@ -49,3 +52,75 @@ pub struct MarginfiAccountInitialize<'info> {
 
     pub system_program: Program<'info, System>,
 }
+
+pub fn initialize_account_pda(
```
Henry-E, line 61: Could this instruction be in its own file?

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -106,3 +110,127 @@ impl<'info> TransferToNewAccount<'info> {
         )
     }
 }
+
+pub fn transfer_to_new_account_pda(
```
Henry-E, line 119: Same again here, we've mostly been keeping instruction in separate files. Although I know in some files we've stacked them. Don't see any reason to stack there though I guess.

jgur-psyops: If they are very similar we sometimes keep them together e.g. configure_bank has a few basic config ixes and configure_bank_lite also has a few similar ones. Here they are super similar so it seems pretty obvious to keep them in one file.

programs/marginfi/src/instructions/marginfi_account/initialize.rs
jgur-psyops: If they are very similar we sometimes keep them together e.g. configure_bank has a few basic config ixes and configure_bank_lite also has a few similar ones. Here they are super similar so it seems pretty obvious to keep them in one file.

jgur-psyops:
> Again sort of curious where the high level idea for this feature was discussed. If it's in the internal notion, would it be ok to still include a link here, just helpful for context to understand the thought process that went into figuring out what the goals were for the feature and how that made its way into the implementation.
> 
> [#379 (comment)](https://github.com/mrgnlabs/marginfi-v2/pull/379#issue-3320151561) - the comment here is a useful summary but more of a literal summary of the work done.

Yes it's been on the internal roadmap (in Notion) for some time, iirc at least since Q2 but maybe just the start of Q3. Roughly once per quarter I put all the "easier" items up as issues on GH with a more particular description for the community in case anyone wants to tackle them, and in this case they did so big thanks to our awesome community for doing so!

programs/marginfi/src/instructions/marginfi_account/initialize.rs
```diff
@@ -49,3 +52,75 @@ pub struct MarginfiAccountInitialize<'info> {
...
+            authority.key().as_ref(),
+            &account_index.to_le_bytes(),
+            &third_party_id.unwrap_or(0).to_le_bytes(),
+        ],
+        bump
+    )]
+    pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
+
+    /// CHECK: Authority is only used for PDA seed derivation, no signing required
+    pub authority: UncheckedAccount<'info>,
```
Henry-E: nit - In the old instruction this used to be required as a signer and all other platforms require the account whose authority it is to be a signer

programs/marginfi/src/instructions/marginfi_account/initialize.rs
```diff
@@ -49,3 +52,75 @@ pub struct MarginfiAccountInitialize<'info> {
...
+
+    #[account(mut)]
+    pub fee_payer: Signer<'info>,
+
+    /// Instructions sysvar for CPI validation
+    ///
+    /// CHECK: Standard sysvar account
+    #[account(address = anchor_lang::solana_program::sysvar::instructions::id())]
```
Henry-E: nit - personally prefer to import `anchor_lang::solana_program::sysvar` just to keep the constraints clean and readable

Henry-E:
> The transfer_account instruction should use the PDA (with the new authority as the seed) variant moving forward.
Nit, this PR introduces a second transfer instruction instead of just updating the existing instruction.

programs/marginfi/src/instructions/marginfi_account/initialize.rs
jgur-psyops: We have account transfers now, so do we care about that? You can always immediately transfer the account to a new authority (though I guess in that instance you at least pay us $0.50 to do it). I like supporting the use-case where the authority does not yet exist (tho transfer does the same thing).

On one hand you could spam-create accounts for somebody relatively cheaply. On the other hand since they are PDAs now it's really trivial to capture them all and then close them for rent so it's a pretty silly troll.

IDK @IliaZyrin you wanna weigh in?

IliaZyrin: I don't have a strong opinion on this, but it seems a bit more logical to have it as a signer. I guess that would prevent sausage-fingers situations at least.

PR #384, opened 2025-09-03
Panic button internal

jgur-psyops:
See #381 for description

Thanks for FEMI for the assist.

Changes:
- Resolve merge conflicts, mostly with type crate
- Can now extend a pause, during a pause, assuming we have not reached other limits


programs/marginfi/src/state/panic_state.rs
```diff
@@ -0,0 +1,320 @@
...
+        Ok(())
+    }
+
+    fn unpause(&mut self) {
+        self.pause_flags &= !Self::FLAG_PAUSED;
+        self.pause_start_timestamp = 0;
+        self.consecutive_pause_count = 0;
+    }
+
+    fn update_if_expired(&mut self, current_timestamp: i64) {
```
IliaZyrin: Maybe let's rename it to `unpause_if_expired` to make it clearer?

programs/marginfi/src/instructions/marginfi_group/panic_unpause_permissionless.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    require!(
+        fee_state.panic_state.is_expired(current_timestamp),
+        crate::errors::MarginfiError::PauseLimitExceeded
+    );
+
+    fee_state.panic_state.unpause();
+
+    msg!(
+        "Protocol auto-unpaused at timestamp: {} (expired after {} seconds)",
```
IliaZyrin: Misleading message: why is it **auto**-unpaused If we directly call `unpause()`?

test-utils/src/marginfi_group.rs
```diff
@@ -750,4 +750,79 @@ impl MarginfiGroupFixture {
...
+            Some(&self.ctx.borrow().payer.pubkey()),
+            &[&self.ctx.borrow().payer],
+            self.ctx.borrow().last_blockhash,
+        );
+
+        self.ctx
+            .borrow_mut()
+            .banks_client
+            .process_transaction(tx)
+            .await
```
IliaZyrin, line 776: Can we factor this out to a separate (`send_ix()` ?) function to not repeat in each try_*?

jgur-psyops: That's probably a good idea but outside the scope of this update

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -52,6 +52,9 @@ pub fn lending_account_borrow<'info>(
 
     let mut marginfi_account = marginfi_account_loader.load_mut()?;
     let group = &marginfi_group_loader.load()?;
+
+    crate::utils::check_protocol_pause_state_cached(group)?;
```
Henry-E: firstly - nit, import crate::utils::check_protocol_pause_state_cached rather than specify the whole path.

Secondly, yet another check inside of the functions that needs to be called that should be in the constraints validation or a single validation function.

programs/marginfi/src/instructions/marginfi_account/deposit.rs
```diff
@@ -48,9 +48,12 @@ pub fn lending_account_deposit<'info>(
 
     let mut bank = bank_loader.load_mut()?;
     let mut marginfi_account = marginfi_account_loader.load_mut()?;
+    let group = marginfi_group_loader.load()?;
```
Henry-E: nit, `group = marginfi` matches with the other let statements but was previously `group = &marginfi` and we only ever use group as a reference.

programs/marginfi/src/instructions/marginfi_group/panic_pause.rs
```diff
@@ -0,0 +1,46 @@
...
+        fee_state.panic_state.consecutive_pause_count
+    );
+
+    Ok(())
+}
+
+#[derive(Accounts)]
+pub struct PanicPause<'info> {
+    /// Admin of the global FeeState (can trigger panic pause)
+    #[account(mut)]
```
Henry-E: why mut?

programs/marginfi/src/instructions/marginfi_group/panic_unpause.rs
```diff
@@ -0,0 +1,51 @@
+use anchor_lang::prelude::*;
+use marginfi_type_crate::{constants::FEE_STATE_SEED, types::FeeState};
+
+use crate::state::panic_state::PanicStateImpl;
+
+pub fn panic_unpause(ctx: Context<PanicUnpause>) -> Result<()> {
```
Henry-E, line 6: why even bother having an unpause if it only pauses for 15 minutes?

programs/marginfi/src/instructions/marginfi_group/panic_unpause_permissionless.rs
```diff
@@ -0,0 +1,43 @@
+use anchor_lang::prelude::*;
+use marginfi_type_crate::{constants::FEE_STATE_SEED, types::FeeState};
+
+use crate::state::panic_state::PanicStateImpl;
+
+pub fn panic_unpause_permissionless(ctx: Context<PanicUnpausePermissionless>) -> Result<()> {
```
Henry-E, line 6: Oh, so it has to be manually unpaused by someone, even if the 15 minutes run out. Does unpausing really require an instruction? It can't happen automatically?

type-crate/src/types/panic_state_cache.rs
```diff
@@ -0,0 +1,100 @@
...
+
+assert_struct_size!(PanicStateCache, 24);
+assert_struct_align!(PanicStateCache, 8);
+/// Cached panic state information for fast checking during user operations
+#[repr(C)]
+#[cfg_attr(feature = "anchor", derive(AnchorDeserialize, AnchorSerialize))]
+#[derive(Clone, Copy, Default, Zeroable, Pod, Debug, PartialEq, Eq)]
+pub struct PanicStateCache {
+    /// Whether the protocol is currently paused (1 = paused, 0 = not paused)
+    pub pause_flags: u8,
```
Henry-E, line 16: nit - it's not really a flag cause we're not planning to mix states are we?

Henry-E:
As mentioned before, it would be nice to see a validation system that e.g. in this case, by default disallows all instructions from interacting with a group in a paused state. And being explicit about which instructions are allowed to interact when the protocol is paused.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
IliaZyrin: I wanted to comment on the same but was not sure that it's possible to move this (somewhat elegantly) specific check to constraints, because it depends on the time among the other things.

As I understand something like this would work but it would require adding clock to every instruction args. Which I don't think is ideal:

```
#[derive(Accounts)]
pub struct SomeIx<'info> {
    #[account(
        // Fail if protocol is paused and the cache hasn't expired yet.
        constraint = !(
            group.panic_state_cache.is_paused()
            && !group.panic_state_cache.is_expired(clock.unix_timestamp)
        ) @ MarginfiError::ProtocolPaused
    )]
    pub group: Account<'info, MarginfiGroup>,

    /// Clock sysvar for current time
    pub clock: Sysvar<'info, Clock>,
}
```

If you have a better option in mind I'm voting for it though!

jgur-psyops: Yep that was exactly on my TODO list thanks for writing it already hah I will shove it in the constraints now

type-crate/src/types/panic_state_cache.rs
jgur-psyops: We might end up putting more than the pause/unpaused in the bit values here so I wanted to leave some flexability

programs/marginfi/src/instructions/marginfi_group/panic_unpause.rs
jgur-psyops: Up to 30 with extensions. I suppose we don't strictly need it but a 30 minute pause could be very annoying to the consumers so ideally we make it as short as possible whenever needed.

programs/marginfi/src/instructions/marginfi_group/panic_unpause_permissionless.rs
jgur-psyops: We could have propagate handle the check, also mutate the global fee state if needed, and then propagate the already-unpaused state. I think I'll leave that to another PR tho, that's a complicated solution. 

Regardless of if this step is manual the actual functions should resume with no manual intervention (I think) because `panic_state_cache.is_expired` will evaluate true. The unpause function just clears the various flags in the global fee state.

tests/e06_panicMode.spec.ts
```diff
@@ -0,0 +1,488 @@
...
+        ],
+        amount: new BN(0.0000001 * 10 ** ecosystem.wsolDecimals),
+      })
+    );
+    tx.recentBlockhash = await getBankrunBlockhash(bankrunContext);
+    tx.sign(liquidator.wallet);
+    await banksClient.processTransaction(tx);
+  });
+
+  // Note: This is an interesting edge case to consider. While liquidations are allowed to continue
```
jgur-psyops, line 267: @Henry-E @IliaZyrin Thoughts here before we merge? 

Should we allow receivership liquidation to bypass the pause?

I will likely allow the receivership liquidations to bypass the check, which consequently either complicates the check in the constraint that we have here or forces it to move to the handler.

IliaZyrin: I think either both liquidations should be allowed or both - disallowed. We chose the former for the "classic" liquidation, so let's just do the same for receivership.

Henry-E: To be honest I don't quite understand well enough what this pause is trying to achieve and context around it to determine which liquidations should be permitted.

programs/marginfi/src/state/panic_state.rs
```diff
@@ -0,0 +1,364 @@
...
+        }
+
+        require!(
+            self.can_pause(current_timestamp),
+            MarginfiError::PauseLimitExceeded
+        );
+
+        // If already paused and not expired, treats this as an "extend" operation.
+        if self.is_paused_flag() && !self.is_expired(current_timestamp) {
+            self.pause_start_timestamp = self
```
Henry-E, line 31: How does this work exactly? Let's say at `N+x` where `x<pause_duration` if we shift forward the `pause start timestamp` from `N` to `N+pause_duration`. I'm assuming that this doesn't deactivate the pause somehow from between `N+x -> N+pause_duration`?

Henry-E: nit - Perhaps this is one reason using a `pause_start_timestamp`  is less intuitive than using an `end_timestamp`

jgur-psyops: That's the goal, and for you to validate in your review! It does have the awkward side effect of start being in the future, which is indeed somewhat less intuitive.

PR #387, opened 2025-09-09
Liquidation remix cpi restriction

jgur-psyops:
Restricts liquidate_start (and end) to top-level calls only, no CPI liquidations allowed with the new receivership approach.

We had hoped to come up with a more novel approach to ensuring start liquidation is unique, but it is what is. Perhaps another day!

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
@@ -12,7 +12,10 @@ use anchor_lang::{prelude::*, solana_program::sysvar};
 use bytemuck::Zeroable;
 use marginfi_type_crate::{
     constants::ix_discriminators,
-    types::{HealthCache, LiquidationRecord, MarginfiAccount, ACCOUNT_IN_RECEIVERSHIP},
+    types::{
+        HealthCache, LiquidationRecord, MarginfiAccount, ACCOUNT_DISABLED, ACCOUNT_IN_FLASHLOAN,
+        ACCOUNT_IN_RECEIVERSHIP,
+    },
 };
 
 /// (Permissionless) Begins a liquidation: snapshots the account and marks it in receivership. The
@@ -28,11 +31,6 @@ pub fn start_liquidation<'info>(
     {
         let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
         let mut liq_record = ctx.accounts.liquidation_record.load_mut()?;
-
-        check!(
-            !marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
-            MarginfiError::UnexpectedLiquidationState
-        );
 
         // Note: the liquidator can use the health cache state after this ix concludes to plan their
         // liquidation strategy.
@@ -119,7 +117,13 @@ pub struct StartLiquidation<'info> {
     /// Account under liquidation
     #[account(
         mut,
-        has_one = liquidation_record
+        has_one = liquidation_record,
+        constraint = {
+            let acc = marginfi_account.load()?;
+            !acc.get_flag(ACCOUNT_IN_RECEIVERSHIP)
+                && !acc.get_flag(ACCOUNT_IN_FLASHLOAN)
+                && !acc.get_flag(ACCOUNT_DISABLED)
+        } @MarginfiError::UnexpectedLiquidationState
     )]
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
 
--- a/programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
@@ -12,7 +12,10 @@ use fixed::types::I80F48;
 use fixed_macro::types::I80F48;
 use marginfi_type_crate::{
     constants::FEE_STATE_SEED,
-    types::{FeeState, HealthCache, LiquidationRecord, MarginfiAccount, ACCOUNT_IN_RECEIVERSHIP},
+    types::{
+        FeeState, HealthCache, LiquidationRecord, MarginfiAccount, ACCOUNT_DISABLED,
+        ACCOUNT_IN_FLASHLOAN, ACCOUNT_IN_RECEIVERSHIP,
+    },
 };
 
 // TODO more detail
@@ -23,11 +26,6 @@ pub fn end_liquidation<'info>(
     let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
     let mut liq_record = ctx.accounts.liquidation_record.load_mut()?;
     let fee_state = ctx.accounts.fee_state.load()?;
-
-    check!(
-        marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
-        MarginfiError::EndNotLast
-    );
 
     validate_not_cpi_by_stack_height()?;
 
@@ -121,7 +119,13 @@ pub struct EndLiquidation<'info> {
     /// Account under liquidation
     #[account(
         mut,
-        has_one = liquidation_record
+        has_one = liquidation_record,
+        constraint = {
+            let acc = marginfi_account.load()?;
+            acc.get_flag(ACCOUNT_IN_RECEIVERSHIP)
+                && !acc.get_flag(ACCOUNT_IN_FLASHLOAN)
+                && !acc.get_flag(ACCOUNT_DISABLED)
+        } @MarginfiError::UnexpectedLiquidationState
     )]
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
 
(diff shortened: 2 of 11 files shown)


programs/marginfi/src/instructions/marginfi_account/flashloan.rs
```diff
@@ -65,11 +65,7 @@ pub fn check_flashloan_can_start(
...
     let current_ix = instructions::load_instruction_at_checked(current_ix_idx, sysvar_ixs)?;
 
-    check!(
-        get_stack_height() == TRANSACTION_LEVEL_STACK_HEIGHT,
-        MarginfiError::IllegalFlashloan,
-        "Start flashloan ix should not be in CPI"
-    );
+    validate_not_cpi()?;
 
     check!(
```
IliaZyrin: This is also redundant now?

programs/marginfi/src/ix_utils.rs
```diff
@@ -148,6 +151,15 @@ pub fn validate_not_cpi(sysvar: &AccountInfo) {
     }
 }
 
+/// Panics if this the instruction calling this is within a CPI
+pub fn validate_not_cpi() -> MarginfiResult<()> {
+    check!(
+        get_stack_height() == TRANSACTION_LEVEL_STACK_HEIGHT,
+        MarginfiError::NoLiquidationInCpi
```
IliaZyrin: Since it's also used for flash loans ixs, let's change the Error wording, e.g. to:
```
#[msg("This can only be a top-level instruction")] // 6091
ForbiddenInCpi
```

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
```diff
@@ -117,17 +113,18 @@ pub fn check_flashloan_can_start(
         MarginfiError::IllegalFlashloan
     );
 
+    check!(
```
IliaZyrin, line 104: I'm not sure why we need this check here (and in the other places) given that we already have a full-fledged [introspection logic](https://github.com/mrgnlabs/marginfi-v2/blob/liquidation-remix/programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs#L62) in liquidate_start which would prevent this flag to be set anyway.

IliaZyrin: Ok, I see, [we only block top-level ixes](https://github.com/mrgnlabs/marginfi-v2/pull/387/files#diff-6d0975360a9afb7724e5d9125a10325a093b352345993b17a2287e8d73c7dd08R69) there, but still for flashloans it should be enough because these are also non-cpi only.

programs/mocks/src/instructions/start_liquidate.rs
```diff
@@ -0,0 +1,64 @@
+use anchor_lang::prelude::*;
+use anchor_lang::solana_program::instruction::{AccountMeta, Instruction};
+use anchor_lang::InstructionData;
+
+#[derive(Accounts)]
+pub struct StartLiquidationViaCpi<'info> {
```
IliaZyrin, line 6: Nitpick: we usually have struct definition below the ix implementation.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
jgur-psyops: wdym?

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
jgur-psyops: Yeah this is just a sanity check and I also wanted to just not deal with thinking about what happens if these instructions are allowed to comingle. Trying to limit what happens during flashloans generally

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
IliaZyrin: I mean it's another check for CPI, so for the same thing as above.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
IliaZyrin: Then maybe let's stick to this explicit approach everywhere (as it is now) and remove that introspection logic? Since it feels redundant.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
jgur-psyops: I'm not actually that confident we don't need this, I'll the add the check with validate_not_cpi_with_sysvar to receivership I guess.

programs/mocks/src/instructions/start_liquidate.rs
jgur-psyops: Eh it's just the the mocks program it doesn't matter

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
jgur-psyops: I think eventually Henry wants to move these flags or state checks into the constraints in a unified way, so let's keep both for now and then when we do that we can look at how to remove the redundancy because I'm not super confident the state management flags satisfy all the same restrictions are the introspection logic without more tests in place.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
jgur-psyops: Ok now we have both cpi checks (via stack height and sysvar index) in start liquidate and flashloan. I'm not sure why we've had both historically, I think either actually suffices, but I suppose we should keep both just in case.

IliaZyrin: Ok, that's fine.

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
IliaZyrin: Ok.

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -32,6 +33,11 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
         MarginfiError::AccountInFlashloan
     );
 
+    check!(
```
IliaZyrin, line 36: Should we maybe just merge these two checks - for in_flashloan and in_receivership? Looks like they're always together.

jgur-psyops: Maybe, they are a different error but I don't know if that's super helpful. I think we agreed we want to move all checks like this to constraints eventually anyways.

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -62,9 +61,13 @@ pub fn start_liquidation<'info>(
     // Introspection logic
     {
         let sysvar = &ctx.accounts.instruction_sysvar;
+        // TODO set allowed keys to e.g. mrgn, token program, jup, compute, and selection of others
         let ixes = load_and_validate_instructions(sysvar, None)?;
         validate_ix_first(&ixes, ctx.program_id, &ix_discriminators::START_LIQUIDATION)?;
         validate_ix_last(&ixes, ctx.program_id, &ix_discriminators::END_LIQUIDATION)?;
+        // Note: this only validates top-level instructions, all other instructions can still appear
```
Henry-E, line 68: Just verifying what this comment means. 

Is `validate_ixes_exclusive` actually checking that only these 4 instructions can be called in between `start liquidation` and `end liquidation`? `This list essentially bans any ix that's already banned inside CPI` This quote is where the confusion is coming from, what does that comment mean?

Functionally speaking
- `validate ixes exclusive` should include any instructions that are allowed to be called in between `start` and `end` liqudation?
- Is there any other aspects to it that are important to be aware that are hinted at by the comment
- Happy to have the double validation but just checking is the same functionality achieved by checking that an account is not in receivership before  each instruction and only allowing certain instructions to load up account when they are in receivership? Just making sure I'm not missing out on some other aspect of the check happening here

jgur-psyops: It checks that only these appear as top-level instructions. But you could still, for example, call `config_bank` by CPI. That ix doesn't even involve an account (and is admin only), but it could affect the outcome of a liquidation, which is potentially unfair. There are other funny edge cases, for example if the protocol is paused but the pause has not propagated to the group, you can call propagate_fee_state by CPI just before liquidate_end to put the bank into a paused state. 

Ideally none of these edge cases exist and the liquidation sandwich is purely this list of instructions, but that's not currently achievable without some kind of global flag state.

i.e the plan was to restrict calls to any ix outside this list, but there is no clever way to do this unless every ix is restricted to top-level, so as it stands in current implementation this check doesn't really do much except communicate which functions SHOULD be allowed.

One option (the initial idea) is to make `load_and_validate_instructions` only allow Compute, Mrgn, Token Program, and Jup as top level programs. This would essentially complete the requirements above but several restrict what liquidations can do within the tx, for example it would prevent them from using Dflow instead of jup.

PR #393, opened 2025-09-15
Add instruction to clone existing banks

jgur-psyops:
## Background

Setting up banks in staging is tedious and most of the time we just want to test against the existing configuration on mainnet anyways. This non-mainnet only ix allows us to copy a mainnet bank into staging quickly without having to invoke all the various add_bank and config functions to complete the setup.

## Summary
- add a `lending_pool_clone_bank` instruction that copies most configuration from an existing bank owned by the mainnet program id while resetting e.g. share tracking fields
- Also copies emode settings, emissions settings, etc.
- Panics if used on mainnet.
------
https://chatgpt.com/codex/tasks/task_b_68c83d5356388323936f2c7dfacae76c

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/clone_bank.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/clone_bank.rs
@@ -29,7 +29,7 @@ pub fn lending_pool_clone_bank(
     }
 
     // Sanity check
-    if crate::ID == MAINNET_PROGRAM_ID {
+    if crate::ID == MAINNET_PROGRAM_ID || *ctx.program_id == MAINNET_PROGRAM_ID {
         panic!("clone bank cannot run on mainnet deployment");
     }
 
--- a/programs/marginfi/src/lib.rs
+++ b/programs/marginfi/src/lib.rs
@@ -129,6 +129,7 @@ pub mod marginfi {
         marginfi_group::lending_pool_add_bank_with_seed(ctx, bank_config, bank_seed)
     }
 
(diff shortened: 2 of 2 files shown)


programs/marginfi/src/instructions/marginfi_group/clone_bank.rs
```diff
@@ -0,0 +1,251 @@
...
+pub fn lending_pool_clone_bank(
+    ctx: Context<LendingPoolCloneBank>,
+    bank_seed: u64,
+) -> MarginfiResult {
+    if crate::ID != STAGING_ID && crate::ID != LOCALNET_ID {
+        panic!("Staging or localnet only!");
+    }
+
+    // Sanity check
+    if crate::ID == MAINNET_PROGRAM_ID {
```
IliaZyrin: Nitpick but I would remove it, because it looks like some separate check to the one above whereas it's in fact about the same. Confusing.

IliaZyrin (approved):

programs/marginfi/src/instructions/marginfi_group/clone_bank.rs
jgur-psyops: Yeah just me being paranoid wanted to be extra super duper sure it can't be used on mainnet.

IliaZyrin: Ok, I don't have a strong opinion on this. Can be left as is.

Henry-E:
The chances that this has no unintended consequences or side effects is low.

PR #395, opened 2025-09-23
Minor fixes to withdraw in liquidaton and bankrupcy during pause

jgur-psyops:
* handle_bankruptcy can no longer be called during a pause, to mitigate some attacks if it's set to permissionless mode.
* Worthless (zero weight or price) assets are now untouchable during liquidation, since otherwise a liquidator could just take them all.


programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -186,7 +186,16 @@ pub struct LendingAccountWithdraw<'info> {
     #[account(
         mut,
         has_one = group,
-        has_one = liquidity_vault
+        has_one = liquidity_vault,
+        // We want to block withdraw of assets with no weight (e.g. isolated) otherwise the
```
IliaZyrin, line 202: Would be nice to have a test for this - which tries to withdraw the 0-weight asset and fails.

IliaZyrin: If not now, then at least in the future, e.g. create a TODO or GH Issue for it.

IliaZyrin: Thanks!

jgur-psyops:
@Henry-E and @IliaZyrin can you review the new snippet that validates that the price is non-zero during receivership.

programs/marginfi/src/utils.rs
```diff
@@ -316,6 +322,38 @@ pub fn wrapped_i80f48_to_f64(n: WrappedI80F48) -> f64 {
     as_f64
 }
 
+/// Fetch price for a given bank from a properly structured remaining accounts slice as passed to
+/// any risk check. Errors if the bank is not found or the price is zero
+pub fn fetch_asset_price_for_bank<'info>(
```
IliaZyrin, line 327: Can we use it in the other places as well? E.g. in liquidate ix?

jgur-psyops:
@Henry-E @IliaZyrin 

We have an edge case where withdraw_all does not require you to pass the oracle for the account being withdraw from, so we cannot check if it is zero-priced. We can either (1) skip this check (imo preferred since we have no zero-price assets anyways), or (2) force withdraw-all during liquidation to pass the oracle anyways (a weird edge case), or (3) block withdraw-all during liquidation and only allow regular withdraw.

jgur-psyops:
> @Henry-E @IliaZyrin
> 
> We have an edge case where withdraw_all does not require you to pass the oracle for the account ....

I have resolve this by creating a "soft close" state where the position still exists and must be included in oracle checks, but it has a balance of zero. When we implement zero-pricing on the program level that allows skipping oracles, let's make sure to remember that such positions (e.g. those with a zero balance) can also skip oracle checks regardless of the bank settings.

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -83,7 +94,10 @@ pub fn lending_account_withdraw<'info>(
             BankAccountWrapper::find(&bank_loader.key(), &mut bank, lending_account)?;
 
         let amount_pre_fee = if withdraw_all {
-            bank_account.withdraw_all()?
+            // Note: In liquidation we can empty a balance, but it stays on the user account until
```
IliaZyrin: But we've also checked that we calculate the health using the `Maintenance` and `Equity` way, not `Initial` (where it is allowed to omit some oracles). So we're safe in terms of withdrawing all -> it won't let the liquidation succeed when the end health is worse than the start one. Not sure it is good idea to complicate the logic here just for the case which should never give us problems anyway.

jgur-psyops: This fix is similar to the bug with assets that have zero-weight: if an asset has an oracle price of 0 (either because we set it, or oracle error), then the liquidator can just take it all and repay a penny, and health improves. Notably it would be worth 0 in both Equity and Maint in this case so neither would prevent that.

So we want to check that the oracle price isn't 0, but normally withdraw_all lets you completely skip passing the oracle for that asset, so we have this silly hack that keeps the position open to force you to pass the oracle to the function.

But yeah, this may not be worth the complexity, because we never have an oracle price of zero on purpose anyways. It might just be a nice sanity check in case an oracle ever blows up.

IliaZyrin: But you're already checking if the price is zero [above](https://github.com/mrgnlabs/marginfi-v2/pull/395/files#diff-823904705faab762f8a1fe01b204e4161589a8e3bc6bc24a6a85d4e4e33b45deR73). So here we're safe to withdraw all in any case. Or am I missing something?

jgur-psyops: fixed in https://github.com/mrgnlabs/marginfi-v2/pull/401/commits/85e1e920cd10bd14d59647fa49f4cdbb0f73c052 now allows withdraw all as long as the caller still passes all the accounts

PR #399, opened 2025-10-02
feat: separate fee payer and signer in transfer_account

Henry-E:
## Summary
Split authority into two roles in the transfer_account instruction:
- `authority`: signs to authorize transfer
- `fee_payer`: pays fees and rent

## Test plan
- ✅ Separate fee payer test
- ✅ Authority wallet owned by Token Program test (proves signature verification works regardless of wallet ownership)


jgur-psyops (approved):
lgtm

PR #404, opened 2025-10-08
Add 8 Kamino position limit per account

Henry-E:
Prevents accounts from becoming unliquidatable due to CU/heap memory issues during liquidation.

### DIFF
--- a/tests/k17_kaminoLimitsMultiDeposit.spec.ts
+++ b/tests/k17_kaminoLimitsMultiDeposit.spec.ts
@@ -8,14 +8,14 @@
  * |-----------------------------------|-------------------------------------------|----------------------------------------------------------|--------|
  * | Deposit into 8 Kamino banks       | Can user create maximum allowed positions?| ✅ All 8 deposits succeed                                | ✅ k14  |
  * | Try to deposit into 9th Kamino    | Does limit enforcement work?              | ❌ Fails with error 6212                                 | ✅ k14  |
-* | Withdraw & reopen position        | Can user close a position and open new?   | ✅ After withdrawing bank X, can deposit into bank Y     | ✅ k14  |
+ * | Withdraw & reopen position        | Can user close a position and open new?   | ✅ After withdrawing bank X, can deposit into bank Y     | ✅ k14  |
  *
  * ### Complex Multi-Asset Scenarios
  * | Scenario                          | What We're Testing                        | Expected Result                                          | Status |
  * |-----------------------------------|-------------------------------------------|----------------------------------------------------------|--------|
  * | 8 Kamino + 7 regular banks        | Do regular banks count against limit?     | ✅ 15 total positions (only Kamino counted for limit)    | ✅ k17  |
  * | Liquidation with 15 positions     | Can we liquidate complex accounts?        | ✅ Liquidation succeeds despite high account count       | ✅ k17  |
-* | 8 Kamino + 8 regular              | How do the two limits interact?           | ✅ Can fill both limits, then can't add 9th of either    | ❌ TODO |
+ * | 8 Kamino + 8 regular              | How do the two limits interact?           | ✅ Can fill both limits, then can't add 9th of either    | ❌ TODO |
  *
  * ## Liquidation Edge Cases
  *
@@ -52,6 +52,7 @@ import {
   TOKEN_A_RESERVE,
   users,
   verbose,
+  bankrunProgram,
 } from "./rootHooks";
 import { refreshPullOraclesBankrun } from "./utils/bankrun-oracles";
 import {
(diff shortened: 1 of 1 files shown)


programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -937,6 +937,20 @@ impl<'a> BankAccountWrapper<'a> {
                 Ok(Self { balance, bank })
             }
             None => {
+                // Enforce Kamino position limit before creating a new Kamino position
+                if bank.config.asset_tag == ASSET_TAG_KAMINO {
```
jgur-psyops, line 941: if this is a "find" operation (not a create), and you have 8 exactly, doesn't this fail?

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -175,7 +177,8 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
     // Note: Kamino deposits sans-LUT are severely limited.
     const depositsPerTx = 2;
 
-    for (let i = 0; i < kaminoBanks.length; i += depositsPerTx) {
+    // Only deposit into the first MAX_KAMINO_POSITIONS banks
+    for (let i = 0; i < MAX_KAMINO_POSITIONS; i += depositsPerTx) {
```
jgur-psyops, line 182: repeating a deposit to a bank already in the balances list after the last one would validate if it works for a "find" vs "create"

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -225,6 +228,81 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
...
+        amount
+      )
+    );
+
+    depositTx.recentBlockhash = await getBankrunBlockhash(bankrunContext);
+    depositTx.sign(user.wallet);
+    const result = await banksClient.tryProcessTransaction(depositTx);
+
+    // Should fail with error 6212 (0x1844 in hex) - KaminoPositionLimitExceeded
+    if (!result.result) {
```
jgur-psyops: use the assert Bankrun error helper function it has support for more nuanced failure eg during simulation

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -225,6 +228,81 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
...
+      throw new Error(
+        "Expected KaminoPositionLimitExceeded error (0x1844 / 6212), got different error:\n" +
+          logs.join("\n")
+      );
+    }
+  });
+
   // Note: the memory performance of 15 deposits and one borrow should be similar; packing that into
   // one TX without a LUT is probably not posssible.
   it("(user 0) Borrows 15 positions against 1 - validates max borrows possible", async () => {
```
jgur-psyops, line 424: so this test is no longer against 15 positions?

programs/marginfi/src/state/marginfi_account.rs
Henry-E: This is the match `None` branch. Matching some skips this check

tests/k14_limitsWithKamino.spec.ts
Henry-E: Already adding a test that tests withdrawing from the 8th bank and depositing to a 9th works. That'll cover it.

tests/k17_kaminoLimitsMultiDeposit.spec.ts
```diff
@@ -0,0 +1,991 @@
...
+      accountKeypair,
+    ]);
+
+    users[0].accounts.set(USER_ACCOUNT, accountKeypair.publicKey);
+  });
+
+  it("(user 0) Deposits into 8 Kamino + 7 regular TOKEN_A banks", async () => {
+    const user = users[0];
+    const userAccount = user.accounts.get(USER_ACCOUNT);
+    // Reduced from 20 to 10 to maintain similar total collateral with 15 positions (15*10=150 vs original 8*20=160)
```
jgur-psyops: reduced from what? where is 20?

tests/k17_kaminoLimitsMultiDeposit.spec.ts
```diff
@@ -0,0 +1,991 @@
...
+    const depositAmount = new BN(10 * 10 ** ecosystem.tokenADecimals);
+
+    // Deposit into each bank sequentially
+    for (let i = 0; i < MAX_KAMINO_DEPOSITS; i++) {
+      const bank = kaminoBanks[i];
+      const market = kaminoMarkets[i];
+      const reserve = kaminoReserves[i];
+
+      // Derive the obligation for this bank
```
jgur-psyops: ?

tests/k17_kaminoLimitsMultiDeposit.spec.ts
Henry-E: Just maths around making liquidations work when i kept changing the number of banks involved, can delete it

tests/k17_kaminoLimitsMultiDeposit.spec.ts
Henry-E: sure, can delete

tests/k14_limitsWithKamino.spec.ts
Henry-E: No changes were made to this test

PR #411, opened 2025-10-16
Multi-point curves

jgur-psyops:
Implementation of a seven-point curve structure that enables deeper control of interest rates at various utilization levels. Deprecates the simple three-point plateau rate/max rate model.

The new curve approach allows the admin to set up to 5 points (plus the points at util = 0 and util = 100). Each point corresponding to a utilization and base interest rate. Points must be in ascending order (util increasing, rates increasing or equal to previous). To determine what the rate is at a given utilization, we linearly interpolate between the nearest two points. 

## New Fields
* Banks now have:
   * zero_util_rate - The base rate at utilization = 0. A %, as u32, out of 1000%, e.g. 100% = 0.1 * u32::MAX
   * hundred_util_rate - The base rate at full utilization (100%). A %, as u32, out of 1000%, e.g. 100% = 0.1 * u32::MAX
   * points - The base rate at various points between 0 and 100% utilization, exclusive. Exactly five points large. Unused points have util = 0. Util is a %, as u32, out of 100%, e.g. 50% = .5 * u32::MAX. Rate is a %, as u32, out of 1000%, e.g. 100% = 0.1 * u32::MAX.
   * curve_type - For all newly created banks, INTEREST_CURVE_SEVEN_POINT (1). For all banks created prior to this update, INTEREST_CURVE_LEGACY (0). After migration is completed, all banks will have INTEREST_CURVE_SEVEN_POINT.

## Breaking Changes (admin instructions)

* All config instructions affecting interest (add_pool, add_pool_with_seed, configure_bank, configure_bank_interest_only) now take fundamentally different arguments.

## New instructions
* migrate_bank (permissionless) - convert a bank from the legacy curve setup to the new seven-point approach. Does not affect bank operations in any way, the bank is immediately able to function as normal.

## Other Notes

* add_bank_permissionless, and add_pool (kamino) now create different placeholders (using the new curve system). Staked and Kamino banks created prior to this update should still be migrated so they don't die when "accruing interest" (even though it doesn't do anything for them, since they never earn interest).
* Banks on the legacy approach will continue to function as normal before/after migration. The legacy approach will cease to function in 1.7.

## Samples

See the referenced functions for more details on how to convert to/from a float to a u32 representation or to a 5-length points slice.

Sample points creation (Rust):
```
points: make_points(&vec![RatePoint::new(
    p100_to_u32(I80F48!(0.9)),
    p1000_to_u32(I80F48!(1)),
)]),
```

Sample points creation (TS):
```
points: makeRatePoints([0.5], [0.6])
```

## Further Work

TODO tests:
* [x] Unit tests of lerp, base rate calc, and validate seven point
* [x] Demonstrate valid curves
* [ ] Integration tests with multi-point curves

TODO in another PR because this one is already long enough:
* [x] migration ix for legacy banks
* [x] Integration tests for bank migration
* [ ] Maybe allow Staked/Kamino banks to skip accrue interest entirely under all circumstances?

### DIFF
--- a/programs/marginfi/src/lib.rs
+++ b/programs/marginfi/src/lib.rs
@@ -544,8 +544,8 @@ pub mod marginfi {
     pub fn migrate_curve(ctx: Context<MigrateCurve>) -> MarginfiResult {
         marginfi_group::migrate_curve(ctx)
     }
-  
-    // Kamino integration instructions
+
+    /****** Kamino integration instructions *****/
 
     /// (permissionless) Initialize a Kamino obligation for a marginfi bank
     /// * amount - In token, in native decimals. Must be >10 (i.e. 10 lamports, not 10 tokens). Lost


clients/rust/marginfi-cli/src/entrypoint.rs
```diff
@@ -200,6 +203,45 @@ pub enum RiskTierArg {
     Isolated,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct RatePointArg {
+    pub util: u32,
+    pub rate: u32,
+}
+
+impl FromStr for RatePointArg {
```
IliaZyrin, line 212: Maybe let's just implement this directly for `RatePoint` to not introduce redundant entities.

programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
```diff
@@ -47,9 +47,9 @@ pub fn log_pool_info(bank: &Bank) {
         wrapped_i80f48_to_f64(interest.protocol_origination_fee)
     );
     msg!(
-        "Plateau: {:?} Optimal: {:?} Max: {:?}",
-        wrapped_i80f48_to_f64(interest.plateau_interest_rate),
-        wrapped_i80f48_to_f64(interest.optimal_utilization_rate),
-        wrapped_i80f48_to_f64(interest.max_interest_rate)
+        "Rate at 0: {:?} points: {:?}: rate at 100: {:?}",
```
IliaZyrin:
```suggestion
        "Rate at 0: {:?} points: {:?} rate at 100: {:?}",
```

programs/marginfi/src/instructions/marginfi_group/add_pool_permissionless.rs
```diff
@@ -54,11 +54,19 @@ pub fn lending_pool_add_bank_permissionless(
 
     // Note: Some placeholder values are non-zero to handle downstream validation checks.
     let default_ir_config = InterestRateConfig {
-        optimal_utilization_rate: I80F48!(0.4).into(),
-        plateau_interest_rate: I80F48!(0.4).into(),
+        // TODO deprecate in 1.7
```
IliaZyrin: Why not remove right away?

programs/marginfi/src/state/interest_rate.rs
```diff
@@ -172,6 +242,90 @@ impl InterestRateCalc {
...
+        start_x: I80F48,
+        start_y: I80F48,
+        end_x: I80F48,
+        end_y: I80F48,
+        target_x: I80F48,
+    ) -> Option<I80F48> {
+        if end_x <= start_x {
+            return Some(start_y);
+        }
+        if target_x < start_x {
```
IliaZyrin, line 314: I suggest to return `None` in this case as well (and if `target_x > end_x` too), since otherwise we can miss an error in the calculations on the caller level.

type-crate/src/types/interest_rate.rs
```diff
@@ -28,69 +35,187 @@ pub struct InterestRateConfig {
...
+    // pub plateau_interest_rate: Option<WrappedI80F48>,
+    // // TODO deprecate
+    // pub max_interest_rate: Option<WrappedI80F48>,
     pub insurance_fee_fixed_apr: Option<WrappedI80F48>,
     pub insurance_ir_fee: Option<WrappedI80F48>,
     pub protocol_fixed_fee_apr: Option<WrappedI80F48>,
     pub protocol_ir_fee: Option<WrappedI80F48>,
     pub protocol_origination_fee: Option<WrappedI80F48>,
+
+    /// The base rate at utilizatation = 0
```
IliaZyrin: A few typos in this file: `utilizatation` -> `utilization`

IliaZyrin (approved):
Great!

clients/rust/marginfi-cli/src/entrypoint.rs
jgur-psyops: I thought about it, but wanted to avoid doing string stuff in the program. Not sure if this increases the size at compile time, probably not, but you never know. 

And I don't love supporting the CLI in general hah.

programs/marginfi/src/instructions/marginfi_group/add_pool_permissionless.rs
jgur-psyops: That's fair, I guess removing it is a equivalent to setting it to zero. Pretty sure zero = padding.

programs/marginfi/src/state/interest_rate.rs
jgur-psyops: This handles the edge case where somehow the points got out of order, we just return the starting (lower) rate. I suppose we could error instead, gotta think about if that's better or not. Hopefully it doesn't matter either way since point order is enforced at the time of setup!

jgur-psyops: Ok after thinking about it I think I agree it's better to error here, in the unlikely event the points did end up out order at this stage.

programs/marginfi/src/state/interest_rate.rs
```diff
@@ -55,13 +75,49 @@ impl InterestRateConfigImpl for InterestRateConfig {
...
+            }
+        }
+
+        // Points must be strictly increasing in util, and non-decreasing in rate
+        for i in 1..used.len() {
+            let prev = &used[i - 1];
+            let curr = &used[i];
+
+            check!(prev.util < curr.util, MarginfiError::InvalidConfig);
+            check!(prev.rate <= curr.rate, MarginfiError::InvalidConfig);
```
Henry-E: There's a lot of invalid config errors, which might make debugging harder. The main saving grace is perhaps that anchor error gives the line number where the error occurred.

type-crate/src/types/interest_rate.rs
```diff
@@ -28,69 +35,174 @@ pub struct InterestRateConfig {
...
+        out[i] = *p;
+    }
+    out
+}
+
+/// Useful when converting an I80F48 (e.g. apr) into a percentage from 0-1000. Clamps to 1000% if
+/// exceeding that amount. Clamps to zero for negative inputs.
+pub fn p1000_to_u32(value: I80F48) -> u32 {
+    let max_percent: I80F48 = I80F48::from_num(10.0); // 1000%
+    let clamped: I80F48 = value.min(max_percent).max(I80F48::ZERO);
```
Henry-E, line 104: Are we sure we want to clamp silently vs. explicitly fail with invalid inputs?

type-crate/src/types/interest_rate.rs
```diff
@@ -28,69 +35,174 @@ pub struct InterestRateConfig {
...
+    let ratio: I80F48 = clamped / max_percent;
```
Henry-E, line 105: It's an edge case, maybe we don't care, where value > 0 but rounds to 0 because it's so small.

Any value below ~1.16e-10 (0.000000000116) encodes to 0

type-crate/src/types/interest_rate.rs
```diff
@@ -28,69 +35,174 @@ pub struct InterestRateConfig {
...
+    let mut out = [RatePoint::default(); CURVE_POINTS];
+    for (i, p) in points.iter().take(CURVE_POINTS).enumerate() {
+        out[i] = *p;
+    }
+    out
+}
+
+/// Useful when converting an I80F48 (e.g. apr) into a percentage from 0-1000. Clamps to 1000% if
+/// exceeding that amount. Clamps to zero for negative inputs.
+pub fn p1000_to_u32(value: I80F48) -> u32 {
```
Henry-E: i don't personally find these function names that intuitive and they're both so similar to each other could definitely see someone getting mixed up between p100 and p1000 and putting in the wrong one.

One alternative
`convert_util_0_to_1_u32`
`convert_apr_0_to_10_u32` or `encode`
just more directly what's happening.

type-crate/src/types/interest_rate.rs
```diff
@@ -28,69 +35,174 @@ pub struct InterestRateConfig {
...
+    /// The base rate at various points between 0 and 100%, exclusive. Essentially a piece-wise
+    /// linear curve.
+    /// * always in ascending order, e.g. points[0] = first kink point, points[1] = second kink
+    ///   point, and so forth.
+    /// * points where util = 0 are unused
+    pub points: [RatePoint; 5],
+
+    /// Determines which interest rate curve implementation is active.
+    /// 0 = legacy three point curve, 1 = multi-point curve.
+    pub curve_type: u8,
```
Henry-E: This is going to default to 1 for the foreseeable future so maybe better to leave it out config compact?

programs/marginfi/src/state/interest_rate.rs
```diff
@@ -115,32 +181,37 @@ impl InterestRateCalc {
...
 
-        let fee_ir = insurance_fee_rate + group_fee_rate + protocol_fee_rate;
-        let fee_fixed = insurance_fee_fixed + group_fee_fixed + protocol_fee_fixed;
+        let fee_ir: I80F48 = insurance_fee_rate + group_fee_rate + protocol_fee_rate;
+        let fee_fixed: I80F48 = insurance_fee_fixed + group_fee_fixed + protocol_fee_fixed;
 
-        let base_rate_apr = self.interest_rate_curve(utilization_ratio)?;
+        let base_rate_apr: I80F48 = if self.curve_type == INTEREST_CURVE_SEVEN_POINT {
+            self.interest_rate_multipoint_curve(utilization_ratio)?
+        } else {
```
Henry-E: Instead of if else, could match LEGACY, SEVEN and error if neither of those. With this case if somehow someone accidentally put in >1 curve type it would default to 0 silently.

programs/marginfi/src/state/interest_rate.rs
```diff
@@ -55,13 +75,49 @@ impl InterestRateConfigImpl for InterestRateConfig {
...
+            } else {
+                // No "holes": non-zero util after padding is not allowed.
+                check!(!seen_padding, MarginfiError::InvalidConfig);
+
+                used.push(*p);
+            }
+        }
+
+        // Points must be strictly increasing in util, and non-decreasing in rate
+        for i in 1..used.len() {
```
Henry-E, line 111: Maybe I'm wrong here but do we check zero_util_rate < point_1 and point_n < hundred util rate. Is that even necessary / consistent with what's happening here?

jgur-psyops: I think so. We validate points are increasing in the snippet you linked, then we have `zero <= hundred && used.iter().all(|p| zero <= p.rate && p.rate <= hundred)`, so the zero rate < 100 rate, and each point > zero and <= hundred.

type-crate/src/types/interest_rate.rs
jgur-psyops: Fine either way imo. I think we should clamp due to rounding issues in the math conversions e.g. if you want to set exactly 1000% we wouldn't want it to error because you passed 1000.01%.

type-crate/src/types/interest_rate.rs
jgur-psyops: I think that's fine. If we ever had a dynamic system where the rates were being assigned automatically and it decided to try a rate that tiny then rounding to zero would be acceptable.

type-crate/src/types/interest_rate.rs
jgur-psyops: Let's use the power of latin I guess. Centi milli?

PR #413, opened 2025-10-22
Migrate curve functionality

jgur-psyops:
* Migrates legacy curves to the new curve type

* Integration test showing interest accrues before/after in the same way.

### DIFF
--- a/programs/marginfi/fuzz/src/account_state.rs
+++ b/programs/marginfi/fuzz/src/account_state.rs
@@ -13,7 +13,8 @@ use anchor_spl::token_2022::spl_token_2022::{
     state::Mint,
 };
 use bumpalo::Bump;
-use marginfi::{state::bank::BankVaultType};
+use marginfi::state::bank::BankVaultType;
+use marginfi_type_crate::constants::FEE_STATE_SEED;
 use pyth_solana_receiver_sdk::price_update::{PriceUpdateV2, VerificationLevel};
 use safe_transmute::transmute_to_bytes_mut;
 use solana_program::{
@@ -21,7 +22,6 @@ use solana_program::{
 };
 use solana_sdk::{signature::Keypair, signer::Signer};
 use std::mem::size_of;
-use marginfi_type_crate::constants::FEE_STATE_SEED;
 
 pub struct AccountsState {
     pub bump: Bump,
--- a/programs/marginfi/fuzz/src/lib.rs
+++ b/programs/marginfi/fuzz/src/lib.rs
@@ -15,7 +15,7 @@ use fixed_macro::types::I80F48;
 use marginfi::{errors::MarginfiError, instructions::LendingPoolAddBankBumps};
 use marginfi::{instructions::LendingPoolConfigureBankOracleBumps, state::bank::BankVaultType};
 use marginfi_type_crate::types::{
-    make_points, p1000_to_u32, p100_to_u32, RatePoint, INTEREST_CURVE_SEVEN_POINT,
+    centi_to_u32, make_points, milli_to_u32, RatePoint, INTEREST_CURVE_SEVEN_POINT,
 };
 use marginfi_type_crate::{
     constants::FEE_STATE_SEED,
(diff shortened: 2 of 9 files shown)


programs/marginfi/src/instructions/marginfi_group/accrue_bank_interest.rs
```diff
@@ -16,6 +16,7 @@ pub fn lending_pool_accrue_bank_interest(
         ctx.accounts.bank.key(),
     )?;
 
+    // TODO see if we can move this within accrue_interest to save some CU
```
IliaZyrin, line 19: That's what I did initially but then it was not playing well with some situations. Don't remember the details at this point but my solution was to use previous cached values as a basis for accruing calculations and it was not always the proper logic.

programs/marginfi/src/instructions/marginfi_group/migrate_curve.rs
```diff
@@ -0,0 +1,51 @@
...
+    irc.points = make_points(&[point]);
+    irc.curve_type = INTEREST_CURVE_SEVEN_POINT;
+
+    // Zero out the old fields so they can be recycled
+    irc.plateau_interest_rate = I80F48::ZERO.into();
+    irc.max_interest_rate = I80F48::ZERO.into();
+    irc.optimal_utilization_rate = I80F48::ZERO.into();
+
+    let interest = bank.config.interest_rate_config;
+    msg!(
```
Henry-E: Probably not necessary here but wondering for logging in general where we want to easily validate a simulated change about adding
- bank pukey in the log
- the previous interest rate points

Just as a convention, i.e. having bank pubkey + previous values in the msg log as well as the newly set values.

programs/marginfi/src/instructions/marginfi_group/migrate_curve.rs
```diff
@@ -0,0 +1,51 @@
...
+pub fn migrate_curve(ctx: Context<MigrateCurve>) -> Result<()> {
+    let mut bank = ctx.accounts.bank.load_mut()?;
+    let irc = &mut bank.config.interest_rate_config;
+
+    if irc.curve_type == INTEREST_CURVE_SEVEN_POINT {
+        msg!("Already migrated, doing nothing");
+        return Ok(());
+    }
+
+    irc.zero_util_rate = 0;
```
Henry-E: It might be overkill but before setting the rates, may as well just do a quick validation that they're valid. 
- optimal utilization rate >= 0 and <= 1 in I80F48
- plateau interest rate and max interest rate >= 0 and <= 10
- and maybe plateau <= max interest 
Although I get you have clamped max min in the p_to_u32 and make_points also validates this probably, it might be good to know if any banks accidentally triggered this.

programs/marginfi/src/instructions/marginfi_group/migrate_curve.rs
```diff
@@ -0,0 +1,51 @@
...
+use anchor_lang::prelude::*;
+use fixed::types::I80F48;
+use marginfi_type_crate::types::{
+    make_points, p1000_to_u32, p100_to_u32, Bank, RatePoint, INTEREST_CURVE_SEVEN_POINT,
+};
+
+#[derive(Accounts)]
+pub struct MigrateCurve<'info> {
+    #[account(mut)]
+    pub bank: AccountLoader<'info, Bank>,
```
Henry-E, line 13: For pig iron maybe include group and have has_one=group

programs/marginfi/src/instructions/marginfi_group/migrate_curve.rs
jgur-psyops: I'll just call validate again before and after.

programs/marginfi/src/instructions/marginfi_group/migrate_curve.rs
jgur-psyops: I don't see any benefit and this would require loading the group for the fixture banks (pre_migration_bank1's group is not loaded into bankrun, just the bank).

PR #424, opened 2025-10-29
Fixed oracle setup

jgur-psyops:
Adds a fixed-price oracle type. Banks using this type have a declared price that can be read at `bank.config.fixed_price`. They use no external oracle, the price is completely set by the group administrator and never goes stale. This is useful for e.g. stablecoins, assets that are worthless and/or being sunset, or assets that are extremely stable in price for some other reason.

## New fields
* `OracleSetup` has a new enum value: Fixed
* `bank.config.fixed_price`: A WrappedI80F48, storing the declared price for Fixed banks, does nothing for banks for other OracleSetups. Note: a bank can swap from Fixed to another OracleSetup type. If it does so, this field might be populated with a junk value.

## Breaking Changes (Everyone)
 * All instructions that require risk checks (borrow, liquidate, withdraw, etc) must now accommodate fixed-price banks. For these banks, pass just the bank in remaining_accounts, no oracle is required. See `load_observation_account_metas` for a Rust example. For Typescript, assuming you have a list of banks and loaded bank accounts:
 ```
    for (let i = 0; i < bankAccs.length; i++) {
      const setup = bankAccs[i].account.config.oracleSetup;
      const keys = bankAccs[i].account.config.oracleKeys;
      if ("fixed" in setup) {
        remainingAccounts.push([banks[i]]);
      } else if ("stakedWithPythPush" in setup) {
        remainingAccounts.push([banks[i], keys[0], keys[1], keys[2]]);
      } else if ("kaminoPythPush" in setup) {
        remainingAccounts.push([banks[i], keys[0], keys[1]]);
      } else if ("kaminoSwitchboardPull" in setup) {
        remainingAccounts.push([banks[i], keys[0], keys[1]]);
      } else {
        remainingAccounts.push([banks[i], keys[0]]);
      }
    }
```

## Breaking Changes (Liquidators)
* The classic liquidate instruction has two new arguments, now requiring the caller to declare the length of the liquidator's and liquidatee's remaining accounts slice. In normal usage, this simply looks like: 
```
        await liquidateIx(liquidator.mrgnProgram, {
          //.... base accounts
          remaining: [
            assetOracle,
            liabOracle,
            ...liquidatorAccounts,
            ...liquidateeAccounts,
          ],
          amount: amt,
          liquidateeAccounts: liquidateeAccounts.length,
          liquidatorAccounts: liquidatorAccounts.length,
        })
```
* If any asset or liability is fixed, note that you do not have to pass that oracle (see fixed oracle tests in m01/m02 for examples), e.g.:
```
        await liquidateIx(liquidator.mrgnProgram, {
          //.... base accounts
          remaining: [
            // no asset oracle if fixed
            // no liab oracle if fixed
            // liquidator/ee accounts also respect the fixed-oracle rules above.
            ...liquidatorAccounts,
            ...liquidateeAccounts,
          ],
          amount: amt,
          liquidateeAccounts: liquidateeAccounts.length,
          liquidatorAccounts: liquidatorAccounts.length,
        })
```

## Additional Notes on Fixed Oracle Banks
* The oracle_keys[0] slot of Fixed oracle banks is always pubkey default. Other oracle_keys[] fields might still be populated, e.g. the lst mint and sol pool for staked banks, or the reserve for the Kamino banks, this is to simplify switching these banks back into a normal oracle operational mode if needed.
* Staked collateral banks can't take a fixed price. Kamino banks CAN (and so can future integrator banks).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -69,7 +69,7 @@ jobs:
         cargo install cargo-nextest --version "0.9.81" --locked
       shell: bash
 
-    - run: ./scripts/test-program.sh all --sane
+    - run: ./scripts/test-program.sh all
 
   fuzz:
     name: Fuzz The marginfi Program
--- a/scripts/test-program.sh
+++ b/scripts/test-program.sh
@@ -1,33 +1,47 @@
 #!/usr/bin/env bash
-set -e
+set -euo pipefail
 
 ROOT=$(git rev-parse --show-toplevel)
-cd $ROOT
+cd "$ROOT"
 
-program_lib_name=$1
-loglevel=$2
+program_lib_name=${1-}
 
-if [ -z "$program_lib_name" ]; then
-    echo "Usage: $0 <program_lib_name>"
-    exit 1
+if [[ -z "${program_lib_name}" ]]; then
+  echo "Usage: $0 <program_lib_name|all> [-- <extra test filters/flags for the test binary>]"
+  exit 1
 fi
 
-if [ "$loglevel" == "--sane" ]; then
-    loglevel=warn
-    nocapture="--test-threads=1"
+# Defaults tuned for a weak 2-core CI
+JOBS="${NEXTEST_JOBS:-1}"
+RETRIES="${NEXTEST_RETRIES:-2}" 
+export CARGO_BUILD_JOBS="${CARGO_BUILD_JOBS:-2}"
+
+if [[ "${program_lib_name}" == "all" ]]; then
+  package_filter=()
 else
-    loglevel=debug
-    nocapture="--nocapture"
+  package_filter=(--package "${program_lib_name}")
 fi
 
-if [ "$program_lib_name" == "all" ]; then
-    package_filter=""
-else 
-    package_filter="--package $program_lib_name"
+shift 1 || true
+extra_params=( "$@" )
+
+# Minimal logging
+export SBF_OUT_DIR="$ROOT/target/deploy"
+export RUST_LOG="solana_runtime::message_processor::stable_log=warn"
+
+cmd=(cargo nextest run
+     --no-fail-fast
+     "${package_filter[@]}"
+     --features=test,test-bpf
+     -j "${JOBS}"
+     --retries "${RETRIES}"
+     --failure-output=immediate
+     --success-output=never
+)
+
+if [[ ${#extra_params[@]} -gt 0 ]]; then
+  cmd+=(-- "${extra_params[@]}")
 fi
 
-extra_params="${@:3}"
-
-cmd="SBF_OUT_DIR=$ROOT/target/deploy RUST_LOG=solana_runtime::message_processor::stable_log=$loglevel cargo nextest run --no-fail-fast $package_filter --features=test,test-bpf $nocapture -- $extra_params"
-echo "Running: $cmd"
-eval "$cmd"
+echo "Running: ${cmd[*]}"
+exec "${cmd[@]}"

### CODE REVIEW

File: clients/rust/marginfi-cli/src/processor/mod.rs [hunk 1]
```diff
@@ -1313,11 +1324,7 @@ pub fn show_oracle_ages(config: Config, only_stale: bool) -> Result<()> {
                 *b.config.oracle_keys.clone().first().unwrap(),
             )
         })
-        .partition(|(setup, _, _, _)| match setup {
-            OracleSetup::PythPushOracle => true,
-            OracleSetup::SwitchboardPull => false,
-            _ => panic!("Unknown oracle setup"),
-        });
+        .partition(|(setup, _, _, _)| matches!(setup, OracleSetup::PythPushOracle));
```
IliaZyrin: I think we should also include KaminoPythPush here.

IliaZyrin: And StakedPythPush.

programs/marginfi/fuzz/src/account_state.rs
```diff
@@ -261,7 +261,7 @@ impl AccountsState {
         rent: Rent,
     ) -> AccountInfo {
         let data_len = unpadded_len;
-        println!("len: {:?}", data_len);
+        // println!("len: {:?}", data_len);
```
IliaZyrin: Let's remove if not needed.

programs/marginfi/fuzz/src/lib.rs
```diff
@@ -797,6 +797,9 @@ impl<'state> MarginfiFuzzContext<'state> {
         let mut liquidatee_remaining_accounts =
             liquidatee_account.get_remaining_accounts(&self.get_bank_map(), vec![], vec![]);
 
+        // Note: this must happen before append because it mutably drains the source vec
+        let liquidator_accounts = liquidator_remaining_accounts.len() as u8;
```
IliaZyrin: Nitpick but I would rename these to `*_accounts_num` to make it obvious it's not the accounts themselves.

programs/marginfi/fuzz/src/lib.rs
```diff
@@ -1207,14 +1212,14 @@ mod tests {
...
-            let (assets, liabs) = re
+            let (_assets, _liabs) = re
                 .get_account_health_components(
                     marginfi::state::marginfi_account::RiskRequirementType::Maintenance,
                     &mut None,
                 )
                 .unwrap();
 
-            println!("assets {assets} liabs: {liabs}");
+            // println!("assets {assets} liabs: {liabs}");
```
IliaZyrin: Again, let's just remove these logs right away if they are no longer useful.

programs/marginfi/fuzz/src/user_accounts.rs
```diff
@@ -100,7 +100,20 @@ impl<'info> UserAccount<'info> {
...
-                [bank_accounts.bank.clone(), bank_accounts.oracle.clone()]
+
+                let bank: Bank = {
+                    *AccountLoader::<Bank>::try_from(&bank_accounts.bank)
+                        .unwrap()
+                        .load()
+                        .unwrap()
+                };
+
+                // For Fixed-price banks: only bank account otherwise: include bank and oracle
```
IliaZyrin: Nitpick for clarity:
```suggestion
                // For Fixed-price banks: only bank account, otherwise: include bank and oracle
```

programs/marginfi/src/instructions/marginfi_group/set_fixed_oracle_price.rs
```diff
@@ -0,0 +1,73 @@
...
+    if bank.config.asset_tag == ASSET_TAG_STAKED {
+        msg!("Staked banks cannot set a fixed price");
+        return err!(MarginfiError::StakedPythPushWrongAccountOwner);
+    }
+
+    bank.config.oracle_setup = OracleSetup::Fixed;
+    // Note: We leave the other keys in place to make it easier to restore Kamino/Staked/etc banks
+    // to their original state. This can leave fixed banks in a somewhat awkward-looking state where
+    // oracles[0] is empty and other slots are not.
+    bank.config.oracle_keys[0] = Pubkey::default();
```
IliaZyrin, line 35: Why do we set the 0th one to default? Isn't setting the `oracle_setup` enough by itself?

programs/marginfi/src/state/price.rs
```diff
@@ -61,28 +61,25 @@ pub trait PriceAdapter {
...
-            clock,
-            bank_config.get_oracle_max_age(),
-        )
+        Self::try_from_bank_with_max_age(bank, ais, clock, bank.config.get_oracle_max_age())
     }
 
-    pub fn try_from_bank_config_with_max_age<'info>(
-        bank_config: &BankConfig,
+    pub fn try_from_bank_with_max_age<'info>(
+        bank: &Bank,
```
IliaZyrin, line 77: Why is this change if we anyway use only the config fields internally?

programs/marginfi/tests/user_actions/liquidate.rs
```diff
@@ -1196,11 +1200,14 @@ async fn marginfi_account_liquidation_failure_bank_not_liquidatable() -> anyhow:
     Ok(())
 }
 
+// ??? this test is occasionally too slow to complete, shave off some tests cases?
```
IliaZyrin, line 1203: Yes, let's do it. Just leave a couple ones.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -91,6 +91,8 @@ use marginfi_type_crate::types::{Bank, MarginfiAccount, MarginfiGroup, ACCOUNT_I
 pub fn lending_account_liquidate<'info>(
     mut ctx: Context<'_, '_, 'info, 'info, LendingAccountLiquidate<'info>>,
     asset_amount: u64,
+    liquidatee_accounts: u8,
```
IliaZyrin, line 94: I don't really like the fact that we now need to pass these counts. Thought about possible workarounds for this but didn't come up with anything nicer. I guess it's not that big of a problem though, given that it's only for the old approach.

programs/marginfi/src/instructions/marginfi_group/set_fixed_oracle_price.rs
jgur-psyops: I thought some developers might have a flow where they don't even bother to read the type and just pass the keys as they appear, in this case their flow can ignore passing accounts if index 0 is pubkey default.

programs/marginfi/tests/user_actions/liquidate.rs
jgur-psyops: I bumped the retries on the test-remix and added a threads arg, it no longer fails (on my machine) so I guess let's revisit this if it starts to, I don't remember which test cases are important any longer.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
jgur-psyops: Yeah it sucks, the alternative is we change the pack order so all the banks come first, then the oracles. We could then read bank info before starting on the oracle slice. But changing pack order is more annoying than this, I think.

Actually making this change for tests wasn't too painful, you already have the liquidator and liquidatee slices most of the time, just calling len() on them isn't too bad.

programs/marginfi/src/state/price.rs
jgur-psyops: Initially, I was going to put the FixedPrice on the Bank, and then I was too lazy to revert this change. Why not!

PR #426, opened 2025-10-31
Allow kamino refreshes before start_liquidate ix

IliaZyrin:
This change is necessary to enable new (receivership) liquidation approach on accounts holding Kamino positions.


programs/marginfi/src/ix_utils.rs
```diff
@@ -53,10 +54,11 @@ pub fn validate_ix_first(
         let discrim = &instruction.data[0..8];
 
         // If this is the first non-compute ix, it is either the setup ix, or we fail
-        if !first_non_compute_ix_encountered {
+        if !start_ix_encountered {
             if instruction.program_id == *program_id && discrim == expected_hash {
-                first_non_compute_ix_encountered = true;
-            } else {
+                start_ix_encountered = true;
+            } else if !(instruction.program_id == *program_id && allowed_hashes.contains(&discrim))
```
jgur-psyops: I think we also want to check that the program id is one of the two allowed (compute or kamino) otherwise there can be a hash collision with an ix from a different program

programs/marginfi/src/ix_utils.rs
```diff
@@ -28,17 +28,18 @@ pub fn get_discrim_hash(namespace: &str, name: &str) -> [u8; 8] {
...
     program_id: &Pubkey,
     expected_hash: &[u8],
+    allowed_ixs: &[(Pubkey, &[u8])],
 ) -> MarginfiResult<()> {
     let compute_budget_key = COMPUTE_PROGRAM_KEY;
-    let mut first_non_compute_ix_encountered = false;
+    let mut expected_ix_encountered = false;
 
     for instruction in ixes.iter() {
         if instruction.program_id == compute_budget_key {
```
jgur-psyops, line 45: If I had to nit, I would probably not take `allowed_ixs` as an arg, and just have added something like
```
        if instruction.program_id == kamino && (hash == RefreshReserve::DISCRIMINATOR || hash == RefreshObligation::DISCRIMINATOR) {
            continue;
        }
```

but since we are probably going to have to do this again for Drift and solend and etc I can see the advantage of the current approach, tho it's going to look a little goofy too once that list gets long.

jgur-psyops (approved):
LGTM, and really cool we have Kamino mocks in the rust test suite now!

programs/marginfi/src/ix_utils.rs
IliaZyrin: We will think how to improve it when there is a new integration!

PR #427, opened 2025-11-04
Forced deleverage

IliaZyrin:
This introduces the concept of a **forced** (by risk admin) **deleverage** - similar to liquidation but applicable to healthy accounts as well.

For this two new instructions introduced:
- `start_deleverage` (similar to `start_liquidation`)
- `end_deleverage` (similar to `end_liquidation`)

Both can be run only by the new kind of admin - `risk_admin` - set for a given group. The same kind of admin is now also responsible for bankruptcy handling (see `handle_bankruptcy`).

Deleverage doesn't pay any fees but enforces the same restrictions on the allowed ixes as the liquidation. It also (just like liquidation) requires the target user account's health to not become worse as a result of deleverage.

There is a newly introduced **withdrawal limit** set **per group** which serves as the protection mechanism against huge unwanted (e.g. if risk admin is compromised) withdrawals. It can be set using the respectively named instruction.

The last addition is "tokenless repayments" - if corresponding flag is set for a given bank, risk admin can (as part of deleveraging) **repay** the user debts in it **without** actually transferring **tokens**. Even though it still does withdraw the collateral.
This is to be used for sunsetting the very illiquid tokens from the platform.

### DIFF
--- a/programs/marginfi/tests/admin_actions/deleverage.rs
+++ b/programs/marginfi/tests/admin_actions/deleverage.rs
@@ -261,21 +261,6 @@ async fn deleverage_tokenless_up_to_limit() -> anyhow::Result<()> {
     assert!(res.is_err());
     assert_custom_error!(res.unwrap_err(), MarginfiError::ZeroWithdrawalLimit);
 
-    test_f
-        .marginfi_group
-        .try_update_deleverage_withdrawal_limit(10)
-        .await?; // cap the deleverages by $10
-    test_f
-        .marginfi_group
-        .try_lending_pool_configure_bank(
-            usdc_bank,
-            BankConfigOpt {
-                tokenless_repayments_allowed: Some(true), // enable repayments with no USDC on the account
-                ..BankConfigOpt::default()
-            },
-        )
-        .await?;
-
     // LP provides initial liquidity
     let lp_usdc_acc = test_f.usdc_mint.create_token_account_and_mint_to(200).await;
     lp.try_bank_deposit(lp_usdc_acc.key, usdc_bank, 100, None)
@@ -294,13 +279,28 @@ async fn deleverage_tokenless_up_to_limit() -> anyhow::Result<()> {
         .try_bank_borrow(user_token_usdc.key, usdc_bank, 20.0)
         .await?;
 
+    test_f
+        .marginfi_group
+        .try_update_deleverage_withdrawal_limit(10)
+        .await?; // cap the deleverages by $10
+    test_f
+        .marginfi_group
+        .try_lending_pool_configure_bank(
+            usdc_bank,
+            BankConfigOpt {
+                tokenless_repayments_allowed: Some(true), // enable repayments with no USDC on the account
+                ..BankConfigOpt::default()
+            },
+        )
+        .await?;
+
     let (record_pk, _bump) = Pubkey::find_program_address(
         &[LIQUIDATION_RECORD_SEED.as_bytes(), deleveragee.key.as_ref()],
         &marginfi::ID,
     );
 
     // Risk admin will withdraw some sol and will "repay" some usdc (in fact there is no usdc on the ATA, so it will be tokenless)
-    let risk_admin_usdc_acc = test_f.usdc_mint.create_empty_token_account().await;
+    let risk_admin_usdc_acc = test_f.usdc_mint.create_token_account_and_mint_to(10).await;
     let risk_admin_sol_acc = test_f.sol_mint.create_empty_token_account().await;
 
     // Note: deleveraging also (like liquidation) uses liquidation record - to ensure we do not worsen the account health.
@@ -317,7 +317,7 @@ async fn deleverage_tokenless_up_to_limit() -> anyhow::Result<()> {
         .make_bank_withdraw_ix(risk_admin_sol_acc.key, sol_bank, 1.0, None, true)
         .await;
 
-    // "Repay" $10 with nothing
+    // Repay $10 (even though tokenless repayments are allowed, if withdraw_all is false, we will repay WITH tokens)
     let repay_ix = deleveragee
         .make_bank_repay_ix(risk_admin_usdc_acc.key, usdc_bank, 10.0, None)
         .await;
@@ -345,7 +345,7 @@ async fn deleverage_tokenless_up_to_limit() -> anyhow::Result<()> {
     let risk_admin_sol_tokens = risk_admin_sol_acc.balance().await;
     assert_eq!(risk_admin_sol_tokens, native!(1.0, "SOL", f64));
     let risk_admin_usdc_tokens = risk_admin_usdc_acc.balance().await;
-    assert_eq!(risk_admin_usdc_tokens, native!(0, "USDC")); // still nothing, well, sad but true!
+    assert_eq!(risk_admin_usdc_tokens, native!(0, "USDC"));
 
     // Now let's try to seize more (and thus "repay" more) and see it failing due to daily withdrawal limit reach
     let start_ix = deleveragee


programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -86,6 +68,78 @@ pub fn end_liquidation<'info>(
...
+    let mut liq_record = ctx.accounts.liquidation_record.load_mut()?;
+
+    end_receivership(
+        &mut marginfi_account,
+        &mut liq_record,
+        ctx.remaining_accounts,
+        true,
+    )?;
+
+    // TODO: Do we want to emit an event in this case?
```
IliaZyrin: @jgur-psyops what do you think?

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -86,6 +68,78 @@ pub fn end_liquidation<'info>(
...
+        .check_pre_liquidation_condition_and_get_account_health(
+            None,
+            &mut Some(&mut post_hc),
+            ignore_healthy,
+        )?;
+    let (post_assets_equity, post_liabilities_equity) = risk_engine
+        .get_account_health_components(RiskRequirementType::Equity, &mut Some(&mut post_hc))?;
+    marginfi_account.health_cache = post_hc;
+
+    // health must improve - TODO: should we relax it to 'not worsen' rather? especially now that we have deleveraging
```
IliaZyrin: @jgur-psyops what do you think?

This would also actually match the Error name if we do it.

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
jgur-psyops: doesn't hurt! let's do it

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
jgur-psyops: yeah not worsen is reasonable. not sure if we must cap profits given that it's admin only I think we can be generous to give us flex in case we have to deleverage something with weird pricing

programs/marginfi/src/instructions/marginfi_group/handle_bankruptcy.rs
```diff
@@ -55,7 +55,7 @@ pub fn lending_pool_handle_bankruptcy<'info>(
 
     if !bank.get_flag(PERMISSIONLESS_BAD_DEBT_SETTLEMENT_FLAG) {
         check!(
-            ctx.accounts.signer.key() == marginfi_group_loader.load()?.admin,
+            ctx.accounts.signer.key() == marginfi_group_loader.load()?.risk_admin,
```
jgur-psyops: nit: either risk_admin OR regular admin probably fine here, but I don't think we're going to use the latter so nbd.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -26,10 +27,16 @@ pub fn configure(
...
     marginfi_group.update_limit_admin(new_limit_admin);
     marginfi_group.update_emissions_admin(new_emissions_admin);
+    marginfi_group.update_risk_admin(new_risk_admin);
     marginfi_group.set_arena_group(is_arena_group)?;
 
-    let clock = Clock::get()?;
-    marginfi_group.fee_state_cache.last_update = clock.unix_timestamp;
+    // The fuzzer should ignore this because the "Clock" mock sysvar doesn't load until after the
+    // group is init. Eventually we might fix the fuzzer to load the clock first...
+    #[cfg(not(feature = "client"))]
```
jgur-psyops, line 35: As yes it's been a while since I hit this particular footgun...

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -128,6 +113,56 @@ pub fn start_liquidation<'info>(
...
+) -> MarginfiResult {
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    let mut liq_record = ctx.accounts.liquidation_record.load_mut()?;
+    liq_record.liquidation_receiver = ctx.accounts.risk_admin.key();
+    start_receivership(
+        &mut marginfi_account,
+        &mut liq_record,
+        ctx.remaining_accounts,
+        true,
+    )?;
```
jgur-psyops, line 69: I think we still need to do some tx validation here too. A start without an end is an easy footgun and that would kinda brick the account. We can probably keep the cpi restriction too, since it's safe to assume the admin will not be a MS: these txes would be pretty much impossible to run through a MS.

That just leaves `validate_ixes_exclusive`, which is more like a list of suggestions, so I guess we should think about what else an admin might want to also run within the tx here, and I can't think of anything beyond those 5 already listed.

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -86,6 +68,83 @@ pub fn end_liquidation<'info>(
...
+
+/// (Permissioned) Ends a deleverage. Records the liquidation event in the user's record.
+/// * Fails if account is less healthy than it was at start
+/// Note: no fees taken.
+pub fn end_deleverage<'info>(
+    ctx: Context<'_, '_, 'info, 'info, EndDeleverage<'info>>,
+) -> MarginfiResult {
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    let mut liq_record = ctx.accounts.liquidation_record.load_mut()?;
+
```
jgur-psyops, line 90: Here as with start I think we can keep the no-cpi restriction.

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
IliaZyrin: Same here regarding the allowed ixes. It seems to be a complete list.

So I guess I will just copy all the checks in the common part as well.

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -86,6 +68,84 @@ pub fn end_liquidation<'info>(
...
+    marginfi_account.health_cache = post_hc;
+
+    // health must not get worse
+    if pre_health > post_health {
+        msg!(
+            "pre_health > post_health: {} >= {}",
+            pre_health,
+            post_health
+        );
+        return Err(MarginfiError::WorseHealthPostLiquidation.into());
```
jgur-psyops: nit: err! macro

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -165,3 +213,49 @@ impl Hashable for StartLiquidation<'_> {
...
+    pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
+
+    /// The associated liquidation record PDA for the given `marginfi_account`
+    #[account(mut)]
+    pub liquidation_record: AccountLoader<'info, LiquidationRecord>,
+
+    #[account(
+        has_one = risk_admin,
+        constraint = (
+            !group.load()?.is_protocol_paused()
```
jgur-psyops: I suppose we can allow this when paused, since it's admin only. if the admin is pwned they'd probably do the maximum damage nearly instantly anyways

programs/marginfi/src/instructions/marginfi_account/repay.rs
```diff
@@ -77,27 +78,32 @@ pub fn lending_account_repay<'info>(
...
-        repay_amount_pre_fee,
-        signer_token_account.to_account_info(),
-        bank_liquidity_vault.to_account_info(),
-        authority.to_account_info(),
-        maybe_bank_mint.as_ref(),
-        token_program.to_account_info(),
-        ctx.remaining_accounts,
-    )?;
+    if authority.key() == group.risk_admin && bank.get_flag(TOKENLESS_REPAYMENTS_ALLOWED) {
+        // In some rare cases (e.g. super illiquid token sunset) we allow risk admin
```
jgur-psyops, line 87: I dunno if it's as simple as doing nothing. once we do this, there is not enough asset in the liquidity vault to repay all lenders (because we didn't put any back here). if there's a run on the bank it'll end being first come first served, where what we really wanna do is have all lenders socialize the loss that was incurred by missing this repay.

so I think we wanna mark down the asset share value here similar to what a bankrupcy does.

otoh, if this is hard, we can totally punt on this, let it be first-come first-served, and let everyone else go through a claims portal to get made whole.

that actually might be easier than having everyone take a discount since then everyone must go to the claims portal.

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -55,37 +55,41 @@ pub fn lending_account_withdraw<'info>(
...
 
-    {
-        let group = &marginfi_group_loader.load()?;
+            // Validate price is non-zero during liquidation/deleverage to prevent exploits with stale oracles
+            check!(price > I80F48::ZERO, MarginfiError::ZeroAssetPrice);
 
-        let mut bank = bank_loader.load_mut()?;
+            price
+        } else {
+            // TODO: force users to always pass the oracle, even in case of withdraw_all, to correctly update withdrawn equity.
```
jgur-psyops: I suppose if we implement this on a per bank basis we can actually let this window be denominated in tokens instead so we dont need price, that's one alternative anyways.

programs/marginfi/src/instructions/marginfi_account/repay.rs
jgur-psyops: Upon discussion we agreed doing nothing is fine! Lenders will eventually run out of money and be unable to withdraw, at that point we refund who's left OTC.

type-crate/src/types/group.rs
```diff
@@ -44,12 +44,18 @@ pub struct MarginfiGroup {
     /// Can modify the emissions `flags`, `emissions_rate` and `emissions_mint`, but nothing else,
     /// for every bank under this group
     pub delegate_emissions_admin: Pubkey,
+    /// Can run bankruptcy and forced deleverage ixes to e.g. sunset risky/illiquid assets
+    pub risk_admin: Pubkey,
```
jgur-psyops, line 48: Just noticed this comes before the panic state cache so it would corrupt that data in prod, but we don't currently use panic state cache in prod, so it's actually fine. Nonetheless, gna move it after the cache so it doesn't step on the prod data by chance.

PR #430, opened 2025-11-05
Bank metadata feature

jgur-psyops:
Enables the storing of bank ticker and plain-English description on chain. No more forgetting which bank key is which...

### DIFF
--- a/programs/marginfi/fuzz/src/lib.rs
+++ b/programs/marginfi/fuzz/src/lib.rs
@@ -1077,6 +1077,7 @@ fn initialize_marginfi_group<'a>(
         admin.key(), // curve_admin
         admin.key(), // limit_admin
         admin.key(), // emissions_admin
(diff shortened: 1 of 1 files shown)


type-crate/src/types/bank_metadata.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// * utf-8
+    pub ticker: [u8; 64],
+    /// The token's plain english descripion, e.g US Dollar Coin
+    /// * utf-8
+    pub description: [u8; 128],
+    /// Reserved for future use. Room for a very small icon or something else cool
+    pub data_blob: [u8; 256],
+
+    /// The last data byte in description (padding follows)
+    pub end_description_byte: u16,
```
IliaZyrin, line 31: Isn't `u8` enough for this?

PR #437, opened 2025-11-14
Reduce only bank worth for borrows

IMEF-FEMI:
# Summary

  Implement ReduceOnly bank collateral handling to prevent new loans while maintaining existing position health.

  Changes

  Core Risk Engine Logic: (programs/marginfi/src/state/marginfi_account.rs:284-295)

 ## Test Coverage

 Rust integration tests (programs/marginfi/tests/misc/operational_state.rs):

  TypeScript integration tests (tests/14_reduceOnlyBanks.spec.ts):
  - User deposit/borrow flows with ReduceOnly constraints
  - Mixed collateral scenarios with both ReduceOnly and operational banks


programs/marginfi/tests/misc/operational_state.rs
```diff
@@ -1,10 +1,25 @@
...
 use marginfi::prelude::MarginfiError;
-use marginfi_type_crate::types::{BankConfig, BankConfigOpt, BankOperationalState};
+use marginfi_type_crate::{
+    constants::EXP_10_I80F48,
+    types::{BankConfig, BankConfigOpt, BankOperationalState},
+};
 use pretty_assertions::assert_eq;
 use solana_program_test::*;
 
+use fixed::types::I80F48 as RuntimeI80F48;
```
IliaZyrin: You don't need this alias, it will work as is:
```suggestion
use fixed::types::I80F48;
```

tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+              },
+            },
+          })
+        )
+      );
+
+
+      bank = await program.account.bank.fetch(bankKey);
+      assert.deepEqual(bank.config.operationalState, { operational: {} });
+    } finally {
```
jgur-psyops, line 86: you don't have to, we just throw if tests fail, they all have to pass in CI before we merge anyways so doesn't matter if the others pass and one fails.

tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+import { Marginfi } from "../target/types/marginfi";
+import { bankKeypairA, bankKeypairSol, bankKeypairUsdc, ecosystem, groupAdmin, marginfiGroup, oracles, users } from "./rootHooks";
+import { expectFailedTxWithError } from "./utils/genericTests";
+import { assert } from "chai";
+import { defaultBankConfigOptRaw } from "./utils/types";
+import { accountInit, borrowIx, composeRemainingAccounts, depositIx, withdrawIx } from "./utils/user-instructions";
+import { USER_ACCOUNT } from "./utils/mocks";
+
+describe("Reduce-Only Bank Tests", () => {
+  const program = workspace.Marginfi as Program<Marginfi>;
```
jgur-psyops, line 14: This is fine, but FYI if you want to do a "quick" throwaway test that doesn't have to clean up after itself, use the `genericMultiBankTestSetup` with bankrun like e.g. zb01. The bankrun tests run faster too so those are usually the preference.

The non-bankrun tests that run in sequence are designed to run into weird edge cases because other tests didn't clean up. The usually only clean up settings, so like here you'd probably take the banks out of reduce at the end and that's it.

tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+  it("(user 0) ReduceOnly collateral is worthless for new loans - should fail with RiskEngineInitRejected", async () => {
+    const user = users[0];
+    const userAccount = user.accounts.get(USER_ACCOUNT);
+
+
+    const depositAmountTokenA = 0.5;
+    const depositAmountTokenA_native = new BN(
+      depositAmountTokenA * 10 ** ecosystem.tokenADecimals
+    );
+
```
jgur-psyops, line 113: send a health pulse (see 11_)  before/after the configBank and read that the: 
* `assetValue` was expected before (non-zero suffices here, or you can do the math: 0.5 * 10 * weight * (1-oracle_conf)), where weight is probably whatever is the default for the bank and oracle confidence is `ORACLE_CONF_INTERVAL * CONF_INTERVAL_MULTIPLE`) 
* declines to 0 to after
* the `assetValueMaint` is unchanged before/after

programs/marginfi/tests/misc/operational_state.rs
```diff
@@ -244,3 +258,371 @@ async fn marginfi_group_bank_reduce_only_deposit_failure() -> anyhow::Result<()>
...
+    let borrower_token_account_sol_2 = test_f.sol_mint.create_empty_token_account().await;
+    let res = borrower_mfi_account
+        .try_bank_borrow(borrower_token_account_sol_2.key, sol_bank_f, 1)
+        .await;
+
+    // Should fail because USDC collateral is worth 0 for new borrows (Initial check)
+    assert!(res.is_err());
+    assert_custom_error!(res.unwrap_err(), MarginfiError::RiskEngineInitRejected);
+
+    // The position is still healthy though - verify by checking the account still exists and has the loan
```
jgur-psyops, line 438: Here you could show a health_pulse reports healthy (health_cache.flags has `HEALTH_CACHE_HEALTHY` set, and/or assetValueMaint > liabilityValueMaint), or that a liquidate still fails with a `HealthyAccount` error, but that one is more annoying to set up since you need a second account to be the liquidator.

tests/14_reduceOnlyBanks.spec.ts
IMEF-FEMI: Thank you SM for this... this took so much of my time trying to figure out a bypass without the clean up

I'll be using this from henceforth

tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -0,0 +1,488 @@
...
+
+      const confidence = ORACLE_CONF_INTERVAL * CONF_INTERVAL_MULTIPLE;
+      const adjustedAssetPrice = oracles.tokenAPrice * (1.0 - confidence);
+      const bankA = await program.account.bank.fetch(bankKeypairA.publicKey);
+      const assetWeightInit = wrappedI80F48toBigNumber(bankA.config.assetWeightInit).toNumber();
+      const expectedAssetValue = depositAmountTokenA * adjustedAssetPrice * assetWeightInit;
+
+      // Verify assetValue matches expected calculation before ReduceOnly
+      assert.approximately(
```
jgur-psyops, line 153: Nit: There's an `assertI80F48Equal ` and `assertI80F48Approx` shorthand for these.

tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -0,0 +1,488 @@
...
+        )
+      );
+
+      const accAfter = await program.account.marginfiAccount.fetch(userAccount);
+      const cacheAfter = accAfter.healthCache;
+      const assetValueAfter = wrappedI80F48toBigNumber(cacheAfter.assetValue);
+      const assetValueMaintAfter = wrappedI80F48toBigNumber(cacheAfter.assetValueMaint);
+
+      // Verify assetValue drops to 0 after ReduceOnly
+      assert.equal(assetValueAfter.toNumber(), 0, "Asset value should be 0 after ReduceOnly");
```
jgur-psyops, line 192: Nit: There's an `assertI80F48Equal ` and `assertI80F48Approx ` shorthand for these.

Issue #417, opened 2025-10-23
Price in bank cache


jgur-psyops:
Whenever an instruction modifies the bank in any way (e.g. borrow, withdraw, anything that affects shares) and that instruction uses a price, the bank cache should record the oracle price that was used. Note that this doesn't apply to e.g. deposit, since no price is consumed there.

The purpose of this is to enable ballpark pricing for a given bank's assets without separately fetching the oracle or reading price from e.g. birdeye, assuming it has enough activity to update fairly regularly, which most of our major banks already do.

jgur-psyops:
Closed in #446

PR #440, opened 2025-11-19
feat: price in bank cache implementation (#417)

sqhell:
This PR is to satisfy issue #417


sqhell:
@IMEF-FEMI @jgur-psyops I'm very happy this got merged! I would like to know more about how to do PRs better in the future. I am available right now also if you're hiring :)

PR #441, opened 2025-11-19
Emode additional rate guards

IMEF-FEMI:
## Summary

  Implements additional safety guards for emode (efficiency mode) configurations to prevent under-collateralized lending
  scenarios. This addresses the risk of emode admins inadvertently or maliciously setting collateral/liability weight
  configurations that would allow under-collateralized positions.

  **Key safeguards:**
  - Enforces `CW < LW` (collateral weight must be strictly less than liability weight) to prevent infinite or negative leverage
  - Validates theoretical maximum leverage using the formula: `L = 1 / (1 - CW/LW)`
  - Enforces configurable group-level maximum leverage limit (default: 20x, range: 1x-100x)
  - Stores `max_emode_leverage` in bank cache for efficient validation during emode configuration
  - Validates both initial and maintenance weight configurations against leverage limits

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
@@ -20,7 +20,6 @@ pub fn lending_pool_configure_bank(
     bank_config: BankConfigOpt,
 ) -> MarginfiResult {
     let mut bank = ctx.accounts.bank.load_mut()?;
-    let group = ctx.accounts.group.load()?;
 
     // If settings are frozen, you can only update the deposit and borrow limits, everything else is ignored.
     if bank.get_flag(FREEZE_SETTINGS) {


type-crate/src/types/bank_cache.rs
```diff
@@ -34,8 +34,13 @@ pub struct BankCache {
     ///   earned on that amount.
     /// * in token, in native decimals, as I80F48
     pub accumulated_since_last_update: WrappedI80F48,
+    /// Maximum allowed theoretical leverage for emode configurations.
+    /// L = 1 / (1 - CW/LW) where CW is collateral weight and LW is liability weight.
+    /// A value of 20 means positions can theoretically leverage up to 20x through recursive borrowing.
+    pub max_emode_leverage: WrappedI80F48,
```
jgur-psyops: I don't think we need to cache this on each bank, it looks like we can just read it from the `group` whenever we need it.

type-crate/src/types/group.rs
```diff
@@ -57,7 +57,9 @@ pub struct MarginfiGroup {
     /// Can modify a Bank's metadata, and nothing else.
     pub metadata_admin: Pubkey,
 
-    pub _padding_0: [[u64; 2]; 12],
+    pub emode_max_leverage_cache: WrappedI80F48,
```
jgur-psyops: nit: If it's a positive number in a pretty limited range (here's it's probably going to be like 0-100 or something at most), we sometimes use a u32 to save space (`zero_util_rate` is one example). Here I think it's fine to keep the 2-byte I80F48 since we have plenty of space in the group anyways, but in Bank and Accounts the space is more limited so we're trying to conserve it.

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
+
+    // Ensure collateral weight < liability weight (strictly less than)
+    check!(
+        collateral_weight < liability_weight,
+        MarginfiError::BadEmodeConfig
+    );
+
+    //  ratio =  CW/LW
+    let ratio = collateral_weight
+        .checked_div(liability_weight)
```
jgur-psyops: nit: you can probably use regular div to save that 2 CU since you already checked not zero above.

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
+        .ok_or(MarginfiError::MathError)?;
+
+    // denominator = 1 - CW/LW
+    let denominator = I80F48::ONE
+        .checked_sub(ratio)
```
jgur-psyops: likewise here since you validated CW < LW then it must follow that CW/LW < 1 so this doesn't have to be checked math.

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
-use crate::{check, errors::MarginfiError, prelude::MarginfiResult};
+use crate::{
+    check, errors::MarginfiError, prelude::MarginfiResult, state::bank_config::BankConfigImpl,
+    state::marginfi_account::RequirementType,
+};
+
+/// Default Maximum allowed theoretical leverage for emode configurations.
+/// L = 1 / (1 - CW/LW) where CW is collateral weight and LW is liability weight.
+/// A value of 20 means positions can theoretically leverage up to 20x through recursive borrowing.
+pub const DEFAULT_MAX_EMODE_LEVERAGE: I80F48 = I80F48::lit("20");
```
jgur-psyops: There's an I80F8! macro for constants that lets you use a number instead of a string.

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
+        let liab_init_w: I80F48 = bank_config.get_weight(
+            RequirementType::Initial,
+            marginfi_type_crate::types::BalanceSide::Liabilities,
+        );
+        let liab_maint_w: I80F48 = bank_config.get_weight(
+            RequirementType::Maintenance,
+            marginfi_type_crate::types::BalanceSide::Liabilities,
+        );
+
+        let max_allowed_leverage: I80F48 = bank_cache.max_emode_leverage.into();
```
jgur-psyops: So here I think instead of reading it from the bank cache it can be passed directly to the function since that appears to be the only use of cache here.

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
-                asset_init_w >= I80F48::ZERO && asset_init_w <= I80F48::ONE,
+                max_leverage_init <= max_allowed_leverage,
                 MarginfiError::BadEmodeConfig
             );
-            let two_hundred_percent: I80F48 = I80F48::ONE + I80F48::ONE;
+
+            let max_leverage_maint = calculate_max_leverage(asset_maint_w, liab_maint_w)?;
             check!(
-                asset_init_w <= two_hundred_percent && asset_maint_w <= two_hundred_percent,
+                max_leverage_maint <= max_allowed_leverage,
```
jgur-psyops: Now that I think about it, I think max maintenance weight should be separately configurable. This could be quite a bit higher that the initial weight.

programs/marginfi/src/state/emode.rs
IliaZyrin: Agreed!

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,117 @@
...
+}
+
 impl EmodeSettingsImpl for EmodeSettings {
-    fn validate_entries(&self) -> MarginfiResult {
+    fn validate_entries_with_liability_weights(
+        &self,
+        bank_config: &BankConfig,
+        bank_cache: &marginfi_type_crate::types::BankCache,
+    ) -> MarginfiResult {
+        let liab_init_w: I80F48 = bank_config.get_weight(
```
IliaZyrin, line 82: Nit: I would just use the weights directly, i.e. `bank_config.liability_weight_init.into()` and `bank_config.liability_weight_maint.into()` respectively.

tests/e07_emodeMaxLeverage.spec.ts
```diff
@@ -0,0 +1,351 @@
...
+      );
+
+      tx.recentBlockhash = await getBankrunBlockhash(bankrunContext);
+      tx.sign(groupAdmin.wallet);
+      const result = await banksClient.tryProcessTransaction(tx);
+      // 6075 (BadEmodeConfig)
+      assertBankrunTxFailed(result, "0x17bb");
+    });
+
+    it("(admin) Configure group with max leverage of 0 - should fail", async () => {
```
IliaZyrin: Nitpick but I think this is redundant, we already checked "< 1" case above.

IliaZyrin (approved):
Looks good!

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -57,6 +58,9 @@ pub fn lending_pool_configure_bank(
         });
     }
 
+    // Update bank cache to sync max_emode_leverage from group
+    bank.update_bank_cache(&group)?;
```
jgur-psyops: remove? (doesn't this do nothing now?)

programs/marginfi/src/state/emode.rs
```diff
@@ -1,35 +1,123 @@
...
+    let ratio = collateral_weight / liability_weight;
+
+    // denominator = 1 - CW/LW
+    let denominator = I80F48::ONE - ratio;
+
+    check!(denominator > I80F48::ZERO, MarginfiError::BadEmodeConfig);
+
+    //  leverage: 1 / (1 - CW/LW)
+    let leverage = I80F48::ONE
+        .checked_div(denominator)
```
jgur-psyops: nit: I think since you validated non-zero you don't have to check div here.

programs/marginfi/src/state/emode.rs
```diff
@@ -97,45 +185,86 @@ mod tests {
...
+        let emode_max_maint_leverage = basis_to_u32(DEFAULT_MAINT_MAX_EMODE_LEVERAGE);
+
+        // Using asset weight greater than liability weight is invalid (CW >= LW).
         let entry = EmodeEntry {
             collateral_bank_emode_tag: 1,
             flags: 0,
             pad0: [0u8; 5],
-            asset_weight_init: I80F48!(1.2).into(),
-            asset_weight_maint: I80F48!(1.3).into(),
+            asset_weight_init: I80F48!(1.2).into(), // Equals liab_init_w (invalid!)
```
jgur-psyops, line 240: I tried to split this into two invokes, one where init weight exceeds, another where maint weight exceeds, and realized you can't!? 

Is it actually true that if asset_weight_maint is good then asset_weight_init must be good? Some of the checks could simplify in that case.

I think it might be: since `liab_init < liab_maint`, and `asset_init > asset_maint`, if `liab_maint < asset_maint`, then `liab_init < asset_init` follows.

jgur-psyops: To be clear we can keep the redundant checks, but we should document it in `validate_entries_with_liability_weights` probably right before `check!(asset_init_w < liab_init_w, MarginfiError::BadEmodeConfig);`

IMEF-FEMI: Thanks for pointing this out. i think Its better i remove the redundant checks to make it a bit cleaner 😁

PR #443, opened 2025-11-21
CU optimization

IliaZyrin:


Henry-E:
There's going to be more of these functions coming in with drift and solend as well. We can deal with them later I guess.

tests/k17_kaminoLimitsMultiDeposit.spec.ts
```diff
@@ -982,7 +982,7 @@ describe("k17: Limits test - 8 Kamino + 7 regular TOKEN_A deposits, liquidation
 
     // Add compute budget
     const computeBudgetIx = ComputeBudgetProgram.setComputeUnitLimit({
-      units: 2_000_000,
+      units: 850_000,
```
Henry-E, line 985: Out of interest what's the CU impact of this change? How bad was the original?

IliaZyrin: It was around 1.2mln!

Henry-E: And for future understanding, where does this compute usage all come from? What's actually the biggest offender. 1.2m seems crazy when we're still doing the same maths computation here.

Henry-E: It implies that simply calling scaled supplies takes up 400k compute units?

programs/kamino-mocks/src/state.rs
```diff
@@ -166,24 +152,30 @@ impl MinimalReserve {
 
     /// Wrapper for i128 values (used by Switchboard)
     #[inline]
-    pub fn adjust_i128(&self, raw: i128) -> Result<i128> {
+    pub fn adjust_i128(&self, raw: i128, liq_to_col_ratio: I80F48) -> Result<i128> {
```
Henry-E: We don't need self anymore

programs/marginfi/src/instructions/kamino/local_tests.rs
```diff
@@ -178,56 +142,62 @@ mod tests {
...
-        let res = bank.adjust_u64(32768);
+        let res = bank.adjust_u64(32768, liq_to_col_ratio);
         assert!(res.is_ok());
         assert_eq!(res.unwrap(), 32768);
 
         // Note: More typical decimal values go longer before overflow
         let bank = generic_reserve(u64::MAX, 8, u64::MAX);
 
-        let res = bank.adjust_u64(32_76_800_000_000);
+        let res = bank.adjust_u64(32_76_800_000_000, liq_to_col_ratio);
```
Henry-E: this is using `liq_to_col_ratio` from the previous bank, not the current bank on line 159

Henry-E: Same pattern on the rest of the tests I think,

programs/marginfi/src/state/price.rs
```diff
@@ -261,11 +261,20 @@ impl OraclePriceFeedAdapter {
                 let mut price_feed =
                     PythPushOraclePriceFeed::load_checked(account_info, clock, max_age)?;
 
-                // Adjust Pyth prices & confidence in place
-                price_feed.price.price = reserve.adjust_i64(price_feed.price.price)?;
-                price_feed.ema_price.price = reserve.adjust_i64(price_feed.ema_price.price)?;
-                price_feed.price.conf = reserve.adjust_u64(price_feed.price.conf)?;
-                price_feed.ema_price.conf = reserve.adjust_u64(price_feed.ema_price.conf)?;
+                let (total_liq, total_col) = reserve.scaled_supplies()?;
+                if total_col > I80F48::ZERO {
```
Henry-E, line 265: I don't super like moving the zero check outside of the inside of the function since if for whatever reason someone ever decides to reuse these functions they might forget to include that. Maybe add a comment to ensure this check and/or maybe a test that makes sure that if there's ever a 0 mint on kamino it errors as we would expect it to and it's not a risk. If you don't think a test is needed, that's fine too.

programs/marginfi/src/state/price.rs
```diff
@@ -261,11 +261,20 @@ impl OraclePriceFeedAdapter {
...
+                    let liq_to_col_ratio = total_liq / total_col;
```
Henry-E, line 266: Also just double check, we used to do this calculation as checked maths. Do you reckon it's ok to do raw `/` here?

tests/k17_kaminoLimitsMultiDeposit.spec.ts
IliaZyrin: Yep, so reusing `scaled_supplies` call results decreased the amount of CUs for from 1.2m to 950k. And then reusing `liq_to_col_ratio` gave additional 100k of improvement.

Basically just implemented the suggestions from [this Jon's PR](https://github.com/mrgnlabs/marginfi-v2/pull/403).

programs/marginfi/src/state/price.rs
IliaZyrin: Yes, that's why we only do this in case of `total_col` being non-zero.

programs/marginfi/src/state/price.rs
IliaZyrin: Actually the thing is that there is now no internal function which would potentially divide by zero. So nothing to test for. All `adjust_*` functions simply multiply by ratio.

tests/k17_kaminoLimitsMultiDeposit.spec.ts
Henry-E: That's great news. Actually if we make these changes to the drift and solend PR, then we might be able to remove the restrictions on account numbers. Ok, I'll leave that work limiting the deposit amounts for now and let's revisit once the drift and solend PR is up.

programs/marginfi/src/state/price.rs
Henry-E: Ok, perfect

programs/marginfi/src/state/price.rs
Henry-E: Good stuff

PR #444, opened 2025-11-26
Add Drift and Solend integration

Henry-E:
## Summary

Adds support for Drift and Solend protocol integrations, allowing users to use positions from these protocols as collateral in p0.

## TODOs before merge

- [x] Fix failing non-drift/solend tests
- [x] Merge 1.16 once squashed onto main and incorporate related changes
- [x] ~~Review Ilya's memory and CU efficiency changes from 1.16 PR for inclusion.~~ Will include this in a follow up PR potentially but not worth holding off for.
- [x] Need to test 8+ banks / add a bank limit integration wide of 8 - 10 banks.
- [x] Add support for new liquidation approach
- [x] Verify if Abdul's type crate import issues are resolved
- [x] Monitor validate_config memory issues (currently not appearing but may resurface)

### DIFF
--- a/programs/marginfi/src/instructions/solend/withdraw.rs
+++ b/programs/marginfi/src/instructions/solend/withdraw.rs
@@ -23,9 +23,11 @@ use anchor_spl::token_interface::{
 };
 use bytemuck::Zeroable;
 use fixed::types::I80F48;
-use marginfi_type_crate::constants::LIQUIDITY_VAULT_AUTHORITY_SEED;
 use marginfi_type_crate::types::{
     Bank, HealthCache, MarginfiAccount, MarginfiGroup, ACCOUNT_DISABLED, ACCOUNT_IN_RECEIVERSHIP,
+};
+use marginfi_type_crate::{
+    constants::LIQUIDITY_VAULT_AUTHORITY_SEED, types::ACCOUNT_IN_DELEVERAGE,
 };
 use solend_mocks::cpi::accounts::WithdrawObligationCollateralAndRedeemReserveCollateral;
 use solend_mocks::cpi::withdraw_obligation_collateral_and_redeem_reserve_collateral;
@@ -119,7 +121,7 @@ pub fn solend_withdraw<'info>(
         };
 
         // Track withdrawal limit for risk admin during deleverage
-        if ctx.accounts.authority.key() == group.risk_admin {
+        if marginfi_account.get_flag(ACCOUNT_IN_DELEVERAGE) {
             let withdrawn_equity = calc_value(
                 I80F48::from_num(collateral_amount),
                 price,


Cargo.toml
```diff
@@ -49,7 +49,7 @@ pyth-solana-receiver-sdk = "0.6.0"
 switchboard-on-demand = "0.3.5"
 
 borsh = "1.5.7"
-bytemuck = "1.22.0"
+bytemuck = { version = "1.22.0" }
```
IliaZyrin: Nitpick but I would revert.

programs/drift-mocks/src/macros.rs
```diff
@@ -0,0 +1,29 @@
+#[macro_export]
+macro_rules! assert_struct_size {
```
IliaZyrin, line 2: I feel like we should depend on the type crate's version of these to not copy them for every integration. Or maybe we should have a utils crate for such stuff.

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,446 @@
+use crate::{constants::*, math_error, DriftMocksError};
+use anchor_lang::prelude::*;
+
+// Account discriminators from Drift IDL
+pub const SPOT_MARKET_DISCRIMINATOR: [u8; 8] = [100, 177, 8, 107, 168, 65, 65, 39];
+pub const USER_DISCRIMINATOR: [u8; 8] = [159, 117, 95, 227, 239, 151, 58, 236];
+pub const USER_STATS_DISCRIMINATOR: [u8; 8] = [176, 223, 136, 27, 122, 79, 32, 227];
+
+// Size assertions based on actual Drift account sizes
+// assert_struct_size!(MinimalSpotMarket, 776);
```
IliaZyrin: These should be uncommented before merging, right?

Cargo.toml
IliaZyrin: Nitpick but I would revert this.

programs/marginfi/src/errors.rs
```diff
@@ -379,6 +504,7 @@ impl MarginfiError {
                 | MarginfiError::MissingPythOrBankAccount
                 | MarginfiError::PythPushInvalidWindowSize
                 | MarginfiError::OracleMaxConfidenceExceeded
+                | MarginfiError::KaminoInvalidOracleSetup
```
IliaZyrin: Should we add the same for Drift and Solend here?

programs/marginfi/src/instructions/drift/deposit.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        // Add oracle if provided (not needed if using oracle type QuoteAsset)
+        if let Some(oracle) = &self.drift_oracle {
+            remaining_accounts.push(oracle.to_account_info());
+        }
+
+        // Always add spot market account
+        remaining_accounts.push(self.drift_spot_market.to_account_info());
+
+        // Always add token mint (needed for Token-2022 support)
```
IliaZyrin, line 283: Above it says:

`// 3. Token mint (if needed for Token-2022)`

but here it's added unconditionally. Shouldn't we check if the token is really token2022 here?

programs/marginfi/src/instructions/drift/harvest_reward.rs
```diff
@@ -0,0 +1,189 @@
...
+        };
+
+        let bump = self.bank.load()?.liquidity_vault_authority_bump;
+        let signer_seeds: &[&[&[u8]]] =
+            bank_signer!(BankVaultType::Liquidity, self.bank.key(), bump);
+
+        let cpi_ctx = CpiContext::new_with_signer(program, accounts, signer_seeds)
+            .with_remaining_accounts(remaining_accounts.to_vec());
+
+        // Try u32::MAX as u64 to avoid overflow
```
IliaZyrin, line 163: Why not just use `u64::MAX`?

programs/marginfi/src/instructions/drift/init_user.rs
```diff
@@ -0,0 +1,264 @@
...
+            authority: self.fee_payer.to_account_info(),
+            mint: self.mint.to_account_info(),
+        };
+        let cpi_ctx = CpiContext::new(program, accounts);
+        let decimals = self.mint.decimals;
+        transfer_checked(cpi_ctx, amount, decimals)?;
+        Ok(())
+    }
+
+    pub fn cpi_drift_deposit(
```
IliaZyrin, line 207: Can we reuse this one (and the above one) from `deposit.rs`?

programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -0,0 +1,499 @@
...
+        if let Some(reward_spot_market) = &self.drift_reward_spot_market {
+            remaining_accounts.push(reward_spot_market.to_account_info());
+        }
+
+        // Add second reward spot market if provided (backup for multiple rewards)
+        if let Some(reward_spot_market_2) = &self.drift_reward_spot_market_2 {
+            remaining_accounts.push(reward_spot_market_2.to_account_info());
+        }
+
+        // Always add main token mint (needed for Token-2022 support)
```
IliaZyrin, line 489: Same as for deposit: shouldn't we actually check for token2022 here?

programs/marginfi/src/instructions/solend/deposit.rs
```diff
@@ -0,0 +1,276 @@
...
+    let authority_bump: u8;
+    {
+        let marginfi_account = ctx.accounts.marginfi_account.load()?;
+        let bank = ctx.accounts.bank.load()?;
+        authority_bump = bank.liquidity_vault_authority_bump;
+
+        validate_asset_tags(&bank, &marginfi_account)?;
+        validate_bank_state(&bank, InstructionKind::FailsIfPausedOrReduceState)?;
+
+        check!(
```
IliaZyrin, line 55: This can be moved to constraints.

programs/marginfi/src/instructions/solend/init_obligation.rs
```diff
@@ -0,0 +1,209 @@
...
+        let cpi_ctx = CpiContext::new_with_signer(
+            self.solend_program.to_account_info(),
+            accounts,
+            signer_seeds,
+        );
+        init_obligation(cpi_ctx)?;
+        Ok(())
+    }
+
+    pub fn cpi_transfer_user_to_liquidity_vault(&self, amount: u64) -> MarginfiResult {
```
IliaZyrin, line 166: As with Drift, I think this and the below function can be removed and directly used from the `deposit.rs` file.

programs/marginfi/src/state/drift.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub oracle: Pubkey,
+    pub asset_weight_init: WrappedI80F48,
+    pub asset_weight_maint: WrappedI80F48,
+    pub deposit_limit: u64,
+    /// Either `DriftPythPull` or `DriftSwitchboardPull`
+    pub oracle_setup: OracleSetup,
+    /// Bank operational state - allows starting banks in paused state
+    pub operational_state: BankOperationalState,
+    /// Risk tier - determines if assets can be borrowed in isolation
+    pub risk_tier: RiskTier,
```
IliaZyrin, line 25: I think we can remove this since it's not gonna have liabs anyway.

programs/marginfi/src/state/solend.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub oracle: Pubkey,
+    pub asset_weight_init: WrappedI80F48,
+    pub asset_weight_maint: WrappedI80F48,
+    pub deposit_limit: u64,
+    /// Either `SolendPythPull` or `SolendSwitchboardPull`
+    pub oracle_setup: OracleSetup,
+    /// Bank operational state - allows starting banks in paused state
+    pub operational_state: BankOperationalState,
+    /// Risk tier - determines if assets can be borrowed in isolation
+    pub risk_tier: RiskTier,
```
IliaZyrin, line 25: Same as with Drift - can be omitted.

programs/solend-mocks/src/constants.rs
```diff
@@ -0,0 +1,31 @@
+use fixed::types::I80F48;
+use fixed_macro::types::I80F48;
+
+/// Powers of 10 for decimal calculations
+pub const MAX_EXP_10_I80F48: usize = 24;
+pub const EXP_10_I80F48: [I80F48; MAX_EXP_10_I80F48] = [
```
IliaZyrin: Can we use the ones from `type-crate`?

programs/solend-mocks/src/state.rs
```diff
@@ -0,0 +1,537 @@
...
+// Account sizes
+// Solend's official reserve size is 619 bytes (includes 1-byte version field)
+// Since Anchor adds its own 1-byte discriminator, our struct is 618 bytes
+// Total size when loaded: 1 (discriminator) + 618 (struct) = 619 bytes
+pub const RESERVE_LEN: usize = 619;
+// Obligation size constant for manual validation (Solend's official size)
+pub const OBLIGATION_LEN: usize = 1300;
+pub const LENDING_MARKET_LEN: usize = 290;
+
+// assert_struct_size!(MinimalReserve, RESERVE_LEN);
```
IliaZyrin: Reminder to uncomment. But I guess with `618` instead of `619`?

programs/solend-mocks/src/state.rs
```diff
@@ -0,0 +1,537 @@
...
+        let total_liquidity = reserve.calculate_total_liquidity()?;
+
+        if reserve.collateral_mint_total_supply == 0 || total_liquidity == I80F48::ZERO {
+            // Use initial rate when no supply
+            Ok(CollateralExchangeRate(reserve.initial_exchange_rate()))
+        } else {
+            let mint_supply = I80F48::from_num(reserve.collateral_mint_total_supply);
+
+            let rate = mint_supply
+                .checked_div(total_liquidity)
```
IliaZyrin, line 394: I think we can use non-checked version here since we explicitly checked for zero above.

programs/solend-mocks/src/state.rs
```diff
@@ -0,0 +1,537 @@
...
+pub struct CollateralExchangeRate(pub I80F48);
+
+impl CollateralExchangeRate {
+    /// Create from reserve state
+    pub fn from_reserve(reserve: &SolendMinimalReserve) -> Result<Self> {
+        let total_liquidity = reserve.calculate_total_liquidity()?;
+
+        if reserve.collateral_mint_total_supply == 0 || total_liquidity == I80F48::ZERO {
+            // Use initial rate when no supply
+            Ok(CollateralExchangeRate(reserve.initial_exchange_rate()))
```
IliaZyrin, line 388: Should we error if `initial_exchange_rate()` happened to be zero?

If so, then we could use non-checked maths in conversion functions (`collateral_to_liquidity`, `liquidity_to_collateral`).

type-crate/src/types/bank.rs
```diff
@@ -125,7 +125,18 @@ pub struct Bank {
...
+    /// Drift banks only, otherwise Pubkey default
+    pub drift_user: Pubkey,
+    /// Drift banks only, otherwise Pubkey default
+    pub drift_user_stats: Pubkey,
+    /// Solend banks only, otherwise Pubkey default
+    pub solend_reserve: Pubkey,
+    /// Solend banks only, otherwise Pubkey default
+    pub solend_obligation: Pubkey,
+
+    pub _padding_1: [[u64; 2]; 5], // 8 * 2 * 5 = 80B (reduced from 15 to account for 5 more Pubkeys = 160 bytes)
```
IliaZyrin: This comment is not needed.

```suggestion
    pub _padding_1: [[u64; 2]; 5], // 8 * 2 * 5 = 80B
```

tests/utils/solend-utils.ts
```diff
@@ -0,0 +1,359 @@
...
+  "SOLEND_USDC_RESERVE_SEED_0000000"
+);
+export const SOLEND_TOKENA_RESERVE_SEED = Buffer.from(
+  "SOLEND_TOKENA_RESERVE_SEED_00000"
+);
+
+/**
+ * Layout sizes
+ */
+export const LENDING_MARKET_SIZE = SDK_LENDING_MARKET_SIZE; // Using size from SDK
```
IliaZyrin: I think we can just not rename it in the export line [above](https://github.com/mrgnlabs/marginfi-v2/pull/444/files#diff-62633c46edfa041a0dd8aff65bdedcec9b2a33d8e2e26a6943984fa4f04885bbR16). The reserve and obligation sizes are also already defined [above](https://github.com/mrgnlabs/marginfi-v2/pull/444/files#diff-62633c46edfa041a0dd8aff65bdedcec9b2a33d8e2e26a6943984fa4f04885bbR28), so let's use them to not bloat the exported constts.

programs/drift-mocks/src/macros.rs
Henry-E: Good point, we can fix this in a separate PR.

programs/drift-mocks/src/state.rs
Henry-E: They were commented out because of issues with Drift using unsafe and things being misaligned. Jon originally added these assertions but I don't have a full understanding of what they're meant to catch or what might fail on the backend if they're not asserted. Deleted them to avoid confusion.

programs/marginfi/src/errors.rs
Henry-E: I think we already added these in the 0.1.6 merge.

programs/marginfi/src/instructions/drift/deposit.rs
Henry-E: Updated the comment to clarify: required for Token-2022, harmless to include for regular mints.

programs/marginfi/src/instructions/drift/harvest_reward.rs
Henry-E: Drift's withdraw is a bit weird. If there's more than u32::MAX tokens, the instruction can just be called a second time.

programs/marginfi/src/instructions/drift/init_user.rs
Henry-E: There are a few account differences between the contexts so they can't easily share these functions.

programs/marginfi/src/instructions/solend/deposit.rs
Henry-E: Agree this could be in constraints, but the main instruction functions do this check in the function body so following that convention.

programs/marginfi/src/state/drift.rs
Henry-E: We made risk_tier configurable on Kamino at Jon's request (1c43b2f6), keeping it here for consistency.

programs/solend-mocks/src/constants.rs
Henry-E: You're right, this is one of those things that got missed when merging 0.1.5, thanks for spotting. Trying to catch as much of this kind of thing as possible.

programs/solend-mocks/src/state.rs
Henry-E: Same as Drift, deleted these.

programs/solend-mocks/src/state.rs
Henry-E: Let's add this in a separate PR, maybe as part of the CU optimization work.

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Core adjustment method using u128 for intermediate calculations.
+    /// Applies Drift's exchange rate: adjusted = raw * cumulative_interest / precision
+    fn adjust_oracle_value_u128(&self, raw: u128) -> Result<u128> {
+        let cumulative_interest = self.cumulative_deposit_interest;
+
+        let adjusted_value = raw
+            .checked_mul(cumulative_interest)
+            .ok_or_else(math_error!())?
+            .checked_div(SPOT_CUMULATIVE_INTEREST_PRECISION)
```
IliaZyrin, line 222: Everywhere in the comments above we mention using of `SPOT_BALANCE_PRECISION` (which is 10^9), but eventually divide by `SPOT_CUMULATIVE_INTEREST_PRECISION` (which is 10^10). It is confusing and we also have no usages of `SPOT_BALANCE_PRECISION` and `SPOT_BALANCE_PRECISION_U64` in code.

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+    pub oracle: Pubkey,
+    /// The token mint of the market
+    pub mint: Pubkey,
+    /// The vault used to store the market's deposits
+    pub vault: Pubkey,
+
+    pub padding_1: [u8; 296],
+
+    /// All the fields we need for testing (stored as raw bytes for simplicity)
+    pub deposit_balance: u128,
```
IliaZyrin, line 60: Copying the comment from draft to make sure it's resolved: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2294895383

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+    pub padding_3: [u8; 49],
+
+    pub pool_id: u8,
+    /// Padding to reach 776 bytes total
+    pub padding_4: [u8; 40],
+}
+
+/// Minimal representation of Drift's User account
+/// Only includes the fields we actually need
+#[account(zero_copy(unsafe), discriminator = &USER_DISCRIMINATOR)]
```
IliaZyrin, line 85: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2294891257

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Adjust oracle price using Drift's internal exchange rate
+    /// For Drift integration, we need to apply the exchange rate to the oracle price
+    /// so that MarginFi's valuation calculations are correct.
+    ///
+    /// The goal is: valuation = scaled_balance * adjusted_oracle_price
+    /// Where: valuation = token_amount * oracle_price
+    /// And: token_amount = scaled_balance * cumulative_interest / SPOT_BALANCE_PRECISION
+    ///
+    /// Therefore: adjusted_oracle_price = oracle_price * cumulative_interest / SPOT_BALANCE_PRECISION
```
IliaZyrin, line 199: I would add a couple more steps here for clarity, e.g.:
```suggestion
    /// By definition: valuation = oracle_price * token_amount
    /// Where: token_amount = scaled_balance * cumulative_interest / SPOT_BALANCE_PRECISION
    ///
    /// The goal here is to calculate such adjusted_oracle_price that:
    /// valuation = adjusted_oracle_price * scaled_balance
    ///
    /// Thus: adjusted_oracle_price * scaled_balance = oracle_price * scaled_balance * cumulative_interest / SPOT_BALANCE_PRECISION
    /// 
    /// Therefore: adjusted_oracle_price = oracle_price * cumulative_interest / SPOT_BALANCE_PRECISION
```

IliaZyrin: As I see this function is not used anywhere (except for the as-a-reference mention in the TS tests). And I assume that's intentional - to serve as a hint for the logic we apply here. But maybe let's just remove the function and leave the comment by itself? Or at least mark it as unused - by prefixing "_"

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    /// Convert scaled balance back to token amount for withdrawals
+    ///
+    /// # Parameters
+    /// * `scaled_balance` - Balance in Drift's internal scaled units (SPOT_BALANCE_PRECISION = 10^9)
+    ///
+    /// # Returns
+    /// * Token amount in native mint precision (mint_decimals)
+    pub fn get_withdraw_token_amount(&self, scaled_balance: u64) -> Result<u64> {
```
IliaZyrin, line 174: A reminder to add tests for this: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2284712846

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+            .iter()
+            .filter(|pos| pos.scaled_balance > 0 && pos.balance_type == SpotBalanceType::Deposit)
+            .map(|pos| pos.market_index)
+            .collect()
+    }
+
+    /// Check if Drift has bricked this account with excessive admin deposits
+    /// We support 1 main asset + up to 2 reward assets (3 total active deposits)
+    /// If Drift admin deposited more reward assets, the account cannot withdraw
+    pub fn validate_not_bricked_by_admin_deposits(&self) -> Result<()> {
```
IliaZyrin, line 298: A reminder to add tests for this: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2264545028

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+    /// Validate spot positions for marginfi integration
+    /// - USDC (market_index 0) must use position[0]
+    /// - All other assets must use position[1]
+    /// - Indices 2+ can have admin deposits
+    /// - Position must be deposit type
+    pub fn validate_spot_position(&self, market_index: u16) -> Result<()> {
+        let expected_position_index = if market_index == 0 { 0 } else { 1 };
+
+        // Only validate positions 0 and 1 for the standard pattern
+        for i in 0..2 {
```
IliaZyrin, line 365: Agreed with this: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2264548142

programs/drift-mocks/src/state.rs
```diff
@@ -0,0 +1,436 @@
...
+                            "Position {} has balance_type {:?} but expected Deposit",
+                            i,
+                            position.balance_type
+                        );
+                        return Err(DriftMocksError::InvalidBalanceType.into());
+                    }
+                }
+            } else {
+                // This position is empty - just ensure it's truly empty
+                if position.market_index != 0 {
```
IliaZyrin, line 394: I think it's better to check `cumulative_deposits` here.
https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2264548372

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -120,6 +121,7 @@ pub fn validate_instructions(
         COMPUTE_PROGRAM_KEY,
         id_crate::ID,
         kamino_mocks::kamino_lending::ID,
+        DRIFT_PROGRAM_ID,
```
IliaZyrin: I know that Drift is priority right now, but let's add SOLEND stuff here as well, to not get back to this later.

programs/marginfi/src/instructions/drift/add_pool.rs
```diff
@@ -0,0 +1,242 @@
...
+    bank.drift_user_stats = drift_user_stats.key();
+
+    log_pool_info(&bank);
+
+    group.add_bank()?;
+
+    bank.config.validate()?;
+    bank.config
+        .validate_oracle_setup(ctx.remaining_accounts, None, None, None)?;
+    bank.config.validate_oracle_age()?;
```
IliaZyrin, line 92: Let's also add the same check to Kamino's `add_pool`, while we're here.

And I [agree that it should be part](https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2282511298) of the general config `validate()` function.

programs/marginfi/src/instructions/drift/deposit.rs
```diff
@@ -0,0 +1,292 @@
...
+impl<'info> DriftDeposit<'info> {
+    pub fn cpi_update_spot_market_cumulative_interest(&self) -> MarginfiResult {
+        let accounts = UpdateSpotMarketCumulativeInterest {
+            state: self.drift_state.to_account_info(),
+            spot_market: self.drift_spot_market.to_account_info(),
+            oracle: self
+                .drift_oracle
+                .as_ref()
+                .map(|o| o.to_account_info())
+                .unwrap_or(self.system_program.to_account_info()), // USDC uses system program
```
IliaZyrin, line 225: Can you please point me to the code where Drift expects it? I was only able to find the exception for the [default](https://github.com/drift-labs/protocol-v2/blob/0c1bb266d1c903ccfc4fd1a3c017a09c14366352/programs/drift/src/state/oracle_map.rs#L301) key, not the system_program one.

programs/marginfi/src/instructions/drift/harvest_reward.rs
```diff
@@ -0,0 +1,189 @@
...
+///
+/// Remaining accounts should be passed in the order required by Drift's withdraw instruction:
+/// 1. Oracle accounts (optional)
+/// 2. Spot market accounts (always required)
+/// 3. Token mint (optional for Token-2022)
+pub fn drift_harvest_reward<'info>(
+    ctx: Context<'_, '_, 'info, 'info, DriftHarvestReward<'info>>,
+) -> MarginfiResult {
+    let spot_market_index = {
+        let harvest_spot_market = ctx.accounts.harvest_drift_spot_market.load()?;
```
IliaZyrin, line 33: https://github.com/mrgnlabs/marginfi-v2-internal/pull/115/files#r2294899802

programs/marginfi/src/instructions/drift/harvest_reward.rs
IliaZyrin: I've checked [the Drift's logic](https://github.com/drift-labs/protocol-v2/blob/0c1bb266d1c903ccfc4fd1a3c017a09c14366352/programs/drift/src/instructions/user.rs#L915) and don't see why we can't use u64 here. Can you please elaborate on why you think it is weird?
I think I'm missing some details.

programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -0,0 +1,506 @@
...
+                scaled_decrement = drift_spot_market.get_scaled_balance_decrement(token_amount)?;
+            }
+
+            bank_account.withdraw(I80F48::from_num(scaled_decrement))?;
+
+            (token_amount, scaled_decrement)
+        };
+
+        // Track withdrawal limit for risk admin during deleverage
+        if ctx.accounts.authority.key() == group.risk_admin {
```
IliaZyrin: We [now](https://github.com/mrgnlabs/marginfi-v2/pull/459/files) use:
```suggestion
        if marginfi_account.get_flag(ACCOUNT_IN_DELEVERAGE) {
```

programs/marginfi/src/instructions/solend/withdraw.rs
```diff
@@ -0,0 +1,387 @@
...
+
+        let collateral_amount = if withdraw_all {
+            bank_account.withdraw_all()?
+        } else {
+            bank_account.withdraw(I80F48::from_num(amount))?;
+            amount
+        };
+
+        // Track withdrawal limit for risk admin during deleverage
+        if ctx.accounts.authority.key() == group.risk_admin {
```
IliaZyrin: We [now](https://github.com/mrgnlabs/marginfi-v2/pull/459/files) use:
```suggestion
        if marginfi_account.get_flag(ACCOUNT_IN_DELEVERAGE) {
```

tests/09_withdraw_repay.spec.ts
```diff
@@ -274,6 +310,10 @@ describe("Withdraw funds", () => {
     const user = users[0];
     const userAccKey = user.accounts.get(USER_ACCOUNT);
 
+    // Note: In bankrun we don't advance the clock here since the previous test already
+    // advanced it and claimed emissions. The withdrawEmissionsIx in this transaction
```
IliaZyrin, line 314: It is confusing. If we already claimed emissions after the last advancing, then we should not need to call `withdrawEmissionsIx` here. @jgur-psyops

tests/e04_emodeLiquidation.spec.ts
```diff
@@ -537,7 +538,7 @@ describe("Emode liquidation", () => {
       SystemProgram.transfer({
         fromPubkey: liquidator.wallet.publicKey,
         toPubkey: bankrunProgram.provider.publicKey,
-        lamports: 5678,
+        lamports: 5538,
```
IliaZyrin, line 541: Was it necessary? Just curious.

programs/marginfi/src/instructions/drift/withdraw.rs
IliaZyrin: That and the below comment was pending for weeks, forgot to push it. In any case I'll address all my comments here and above.

programs/drift-mocks/src/state.rs
IliaZyrin: It looks like the correct divisor is indeed [10^10](https://github.com/drift-labs/protocol-v2/blob/master/sdk/src/constants/numericConstants.ts#L42). So I will update the comments here.

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
IliaZyrin: Will not enable Solend for now.

programs/drift-mocks/src/state.rs
IliaZyrin: @imef-femi will add it as part of https://github.com/mrgnlabs/marginfi-v2/pull/473

programs/marginfi/src/instructions/drift/harvest_reward.rs
IliaZyrin: Not a problem as long as Drift has [this check](https://github.com/drift-labs/protocol-v2/blob/983ac0efb65ec4593a5194caffb634719d135263/programs/drift/src/controller/token.rs#L109-L122) in place

PR #446, opened 2025-11-26
Internal/price in bank cache

IMEF-FEMI:
PR for the issue #417 

Authored by @sqhell 

cc: @jgur-psyops

### DIFF
--- a/programs/marginfi/tests/user_actions/bank_cache.rs
+++ b/programs/marginfi/tests/user_actions/bank_cache.rs
@@ -1,7 +1,7 @@
 use fixed::types::I80F48;
 use fixed_macro::types::I80F48;
 use fixtures::{assert_eq_noise, marginfi_account::MarginfiAccountFixture, prelude::*};
-use marginfi_type_crate::types::{BankCache, BankConfigOpt};
+use marginfi_type_crate::types::BankConfigOpt;
 use solana_program_test::tokio;
 
 async fn setup_borrow_with_price_cache(
@@ -80,7 +80,19 @@ async fn bank_cache_resets_after_full_repay() -> anyhow::Result<()> {
         .await?;
 
     let bank_after_repay = sol_bank.load().await;
-    assert_eq!(bank_after_repay.cache, BankCache::default());
+    // No change (uses the price from the prior borrow price!)
+    let expected_price: I80F48 = I80F48::from_num(sol_bank.get_price().await);
+    let cached_price: I80F48 = bank_after_repay.cache.last_oracle_price.into();
+    assert_eq_noise!(cached_price, expected_price);
+    assert_eq!(
+        bank_after_borrow.cache.last_oracle_price_timestamp,
+        bank_after_repay.cache.last_oracle_price_timestamp
+    );
+
+    // The rate fields are reset when liabs are 0
+    assert_eq!(bank_after_repay.cache.base_rate, 0);
+    assert_eq!(bank_after_repay.cache.lending_rate, 0);
+    assert_eq!(bank_after_repay.cache.borrowing_rate, 0);
 
     Ok(())
 }
@@ -118,9 +130,13 @@ async fn bank_cache_pulse_refreshes_price_from_oracle() -> anyhow::Result<()> {
 
     let bank_after_pulse = sol_bank_f.load().await;
     let cached_price_after_pulse: I80F48 = bank_after_pulse.cache.last_oracle_price.into();
-    let expected_price = I80F48::from_num(sol_bank_f.get_price().await);
+    let cached_conf_after_pulse: I80F48 =
+        bank_after_pulse.cache.last_oracle_price_confidence.into();
+    let expected_price: I80F48 = I80F48::from_num(sol_bank_f.get_price().await);
 
     assert_eq_noise!(cached_price_after_pulse, expected_price);
+    // Note: No confidence bands in the Rust tests
+    assert_eq_noise!(cached_conf_after_pulse, I80F48::ZERO);
 
     Ok(())
 }


programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -181,27 +181,37 @@ pub fn kamino_withdraw<'info>(
             amount: collateral_amount,
             close_balance: withdraw_all,
         });
+    }
```
IliaZyrin: Nitpick but just wanted to ask why this change? This way we load the marginfi account twice.

type-crate/src/types/bank_cache.rs
```diff
@@ -35,7 +35,24 @@ pub struct BankCache {
...
+    /// * Only updated when instruction modifies shares AND uses oracle price
+    /// * Not updated for operations that don't require prices (e.g., deposit, repay)
+    /// * Price in USD (or quote currency) per token, as I80F48 (includes any bias applied)
+    /// * Zero if never updated
+    pub last_oracle_price: WrappedI80F48,
+
+    /// Unix timestamp (seconds) when last_oracle_price was last updated
+    /// * Used to determine staleness of cached price
+    /// * Zero if never updated
+    pub last_oracle_price_timestamp: i64,
```
IliaZyrin, line 48: I think we can just use existing Bank's `last_update` field instead of introducing the new one.
But we should then set it in Kamino withdraw (and maybe deposit - for consistency), because it's not updated there at the moment. It's only updated in `accrue_interest()` afaik, which is not called from Kamino instructions.

programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -181,27 +181,37 @@ pub fn kamino_withdraw<'info>(
...
-            );
-            risk_result?;
-            health_cache.program_version = PROGRAM_VERSION;
-
-            health_cache.set_engine_ok(true);
-            marginfi_account.health_cache = health_cache;
+    marginfi_account.lending_account.sort_balances();
+
+    // Note: during liquidating, we skip all health checks until the end of the transaction.
+    if !marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP) {
```
IliaZyrin: I think we should update the price cache in the "receivership" cases (liquidation/deleveraging) as well. It won't work for `withdraw_all` for now, but at least for the regular - yes.

On a side note, when [this](https://github.com/mrgnlabs/marginfi-v2/pull/446/files#diff-b0e8565f180e5eb656433fca060a9af25a6ea7406c5d3bd99248948387106be5L107) is done, we may also re-use the same price, for `withdrawn_equity` calculations and price cache updates, to not fetch it twice.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -379,10 +379,6 @@ pub fn lending_account_liquidate<'info>(
                 .ok_or(MarginfiError::MathError)?
                 .into();
 
-        asset_bank.update_bank_cache(group)?;
```
IliaZyrin, line 382: Not a big deal but I would not move these and instead used the `risk_engine` from [here](https://github.com/mrgnlabs/marginfi-v2/pull/446/files#diff-0784a7c2bddc126f67aa6e98cb8f72639d8e74289f0c7725c5d0288eba045ccdR165).

E.g. replace that part with:
```
let risk_engine = RiskEngine::new(&liquidatee_marginfi_account, liquidatee_remaining_accounts)?;
    let (pre_liquidation_health, _, _) = risk_engine
            .check_pre_liquidation_condition_and_get_account_health(
                Some(&ctx.accounts.liab_bank.key()),
                &mut None,
                false,
            )?;
            
```

programs/marginfi/src/instructions/marginfi_group/handle_bankruptcy.rs
```diff
@@ -175,8 +182,6 @@ pub fn lending_pool_handle_bankruptcy<'info>(
     )?
     .repay(bad_debt)?;
 
-    bank.update_bank_cache(group)?;
```
IliaZyrin, line 178: This must not be moved from here because it must be called after all manipulations with the bank such as interest accrual and loss socialization.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -604,6 +606,29 @@ impl<'info> RiskEngine<'_, 'info> {
...
+
+        let bank = bank_account.bank.load()?;
+        let (price_feed_res, _) = bank_account.try_get_price_feed();
+        let price_feed = price_feed_res?;
+
+        let price = price_feed.get_price_and_confidence_of_type(
+            OraclePriceType::RealTime,
+            bank.config.oracle_max_confidence,
+        )?;
+        check!(price.price > I80F48::ZERO, MarginfiError::ZeroAssetPrice);
```
IliaZyrin: Not sure we should check for zero here. In theory nothing should prevent you from withdrawing your zero-priced assets from the bank.

programs/marginfi/src/state/price.rs
```diff
@@ -634,21 +656,36 @@ impl PriceAdapter for SwitchboardPullPriceFeed {
         let price = self.get_price()?;
 
         match bias {
+            None => Ok(price),
```
IliaZyrin: Nitpick (at least for the future) but I would not change such things because it bloats the PR and takes reviewers' time to check it and only see that nothing has changed except for cosmetics.

programs/marginfi/src/utils/general.rs
```diff
@@ -363,6 +366,70 @@ pub fn fetch_asset_price_for_bank<'info>(
     Ok(price)
 }
 
+/// Fetch liability price for a given bank from a properly structured remaining accounts slice.
+/// Uses High bias for conservative liability valuation. Errors if the bank is not found or the price is zero.
+pub fn fetch_liability_price_for_bank<'info>(
```
IliaZyrin: It's not used anywhere, right? Let's remove it then.

type-crate/src/types/bank_cache.rs
IMEF-FEMI: we might have a challenge with this as the bank cache updates (for rates) and the price updates usually does not necessarily happen at the same time. as sometimes we update the rates without updating the cached price... like in deposit function

IliaZyrin: True, makes sense.

programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -96,8 +97,12 @@ pub fn kamino_withdraw<'info>(
 
         let in_receivership = marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP);
         let price = if in_receivership {
-            let price =
-                fetch_asset_price_for_bank(&bank_key, &bank, &clock, ctx.remaining_accounts)?;
+            let price = fetch_asset_price_for_bank_low_bias(
```
IliaZyrin, line 100: I think ideally we want a non-biased price here. It's not critical at all but if it also helps to reduce the number of `fetch_price_*` functions, let's do it.

programs/marginfi/src/state/bank.rs
```diff
@@ -560,6 +568,30 @@ impl BankImpl for Bank {
...
+    fn update_cache_price(
+        &mut self,
+        oracle_price: Option<OraclePriceWithConfidence>,
+    ) -> MarginfiResult<()> {
+        if let Some(price_with_confidence) = oracle_price {
+            self.cache.last_oracle_price = price_with_confidence.price.into();
+            self.cache.last_oracle_price_confidence = price_with_confidence.confidence.into();
+            self.cache.last_oracle_price_timestamp = Clock::get()?.unix_timestamp;
+        } else {
+            msg!("price cache update failed");
```
IliaZyrin: I think this message is confusing given that [there are cases](https://github.com/mrgnlabs/marginfi-v2/pull/446/files#diff-823904705faab762f8a1fe01b204e4161589a8e3bc6bc24a6a85d4e4e33b45deR219) where we explicitly expect the price to be `None`. So there is nothing to _fail_ here, it's an expected behavior.

On the other hand, in Kamino Withdraw we simply [do not call](https://github.com/mrgnlabs/marginfi-v2/pull/446/files#diff-b0e8565f180e5eb656433fca060a9af25a6ea7406c5d3bd99248948387106be5R210) this if the price is `None`. If we apply the same approach to all the other "maybe" cases, then we can leave the message as it is.

programs/marginfi/tests/user_actions/bank_cache.rs
```diff
@@ -0,0 +1,312 @@
...
+        sol_bank_post.last_update >= sol_bank_pre.last_update,
+        "Bank's last_update should be updated after bankruptcy"
+    );
+
+    Ok(())
+}
+
+/// Test that bank's last_update field is set when bank is modified
+/// This verifies that last_update can be used to determine price staleness
+/// (since we removed the dedicated last_oracle_price_timestamp field)
```
IliaZyrin, line 304: You didn't remove `last_oracle_price_timestamp` though: https://github.com/mrgnlabs/marginfi-v2/pull/446/files#diff-6abc5d6d44fd41e73ed6d1d7c6318c6442f7122712cc1ea4c29c791335ee3b16R48

tests/utils/user-instructions.ts
```diff
@@ -572,6 +578,29 @@ export const healthPulse = (
     .instruction();
 };
 
+/**
+ * Creates a bank price pulse instruction. This ix refreshes the cached oracle price
+ * for a given bank without modifying user positions.
+ * * `remaining` - pass the oracle accounts required for this bank
+ *   in the same order as used for other oracle reads for that bank.
+ */
+export const bankPricePulse = (
```
IliaZyrin: Nitpick but I would name it `pulseBankPrice` for consistency here.

tests/z02_lastOraclePrice.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    const bankAfter = await bankrunProgram.account.bank.fetch(banks[1]);
+    const afterFields = readCacheFields(bankAfter.cache);
+
+    // update_bank_cache was called with None, so price/confidence should be unchanged
+    assertI80F48Equal(afterFields.price, beforeFields.price);
+    assertI80F48Equal(afterFields.conf, beforeFields.conf);
+  });
+
+  it("(user 0) full repay resets bank cache when liabilities go to zero", async () => {
```
IliaZyrin: Just curious what was the reason for this logic?

programs/marginfi/src/instructions/kamino/withdraw.rs
jgur-psyops: I think we did low here in case price > 0 but price - confidence  = 0. Not sure if that's actually possible tho. We need this "low bias" one for liquidation (it's actually the original, I think, and the others were written later).

programs/marginfi/src/state/bank.rs
jgur-psyops: Fair point, I'll just silence the message, it's a waste of compute anyways.

programs/marginfi/tests/user_actions/bank_cache.rs
jgur-psyops: I think let's just leave it. There are these weird edge cases where the rest of the cache updates and the price cache doesn't, so this does give you some extra utility.

tests/z02_lastOraclePrice.spec.ts
jgur-psyops: Good point, maybe we shouldn't. I think when we originally wrote `update_bank_cache`

```
        if (total_assets_amount == I80F48::ZERO) || (total_liabilities_amount == I80F48::ZERO) {
            self.cache = BankCache::default();
            return Ok(());
        }
```

this reset to zero made sense. Now that there's other stuff in the cache what we really want to do is just reset the rates, not the price cache. I'll make it do that.

programs/marginfi/src/instructions/kamino/withdraw.rs
IliaZyrin: I don't think it's possible, but even if so, we still want that "> 0" price. And yeah, we probably just reused the function which was already available, makes sense.

PR #450, opened 2025-11-30
 Admin transfer/freeze account (for KYC/AML compliance)

IMEF-FEMI:
Summary:

  - Introduce an ACCOUNT_FROZEN flag on MarginfiAccount plus a new admin-only instruction marginfi_account_set_freeze that toggles it,
    emitting MarginfiAccountFreezeEvent.
  - Enforce frozen-state semantics across all major account operations (deposit, withdraw, borrow, repay, close balance, close account,
    emissions, flashloans, account transfers, Kamino deposit/withdraw): account authorities are blocked with AccountFrozen (6103) while
    the group admin remains authorized to act.

### DIFF
--- a/.github/actions/base-setup/action.yaml
+++ b/.github/actions/base-setup/action.yaml
@@ -38,9 +38,12 @@ runs:
     if: ${{ inputs.include-anchor == 'true' }}
     uses: ./.github/actions/setup-anchor-cli/
 
-  - name: Cache Anchor build output
+  - name: Cache Anchor build artifacts
     if: ${{ inputs.include-anchor == 'true' }}
     uses: actions/cache@v3
     with:
-      path: target/
-      key: anchor-build-${{ runner.os }}-${{ env.ANCHOR_CLI_VERSION }}-${{ env.ANCHOR_SHA }}-${{ hashFiles('programs/**/Cargo.lock') }}
+      path: |
+        target/deploy
+        target/idl
+        target/types
+      key: anchor-build-slim-v1-${{ runner.os }}-${{ env.ANCHOR_CLI_VERSION }}-${{ env.ANCHOR_SHA }}-${{ hashFiles('programs/**/Cargo.lock') }}


programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -57,6 +57,32 @@ pub trait MarginfiAccountImpl {
...
+    }
+
+    if marginfi_account.authority == signer {
+        return true;
+    }
+
+    marginfi_account.get_flag(ACCOUNT_FROZEN) && group_admin == signer
+}
+
+#[inline(always)]
```
jgur-psyops: nit: probably doesn't need to be inlined

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -57,6 +57,32 @@ pub trait MarginfiAccountImpl {
...
+) -> bool {
+    if allow_receivership && marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP) {
+        return true;
+    }
+
+    if marginfi_account.authority == signer {
+        return true;
+    }
+
+    marginfi_account.get_flag(ACCOUNT_FROZEN) && group_admin == signer
```
jgur-psyops: it makes sense that the group admin can only sign after first freezing! 

nit: I think it's a little odd that the "not frozen" case is handled above should probably be like if(frozen) check group admin (else not frozen) check authority

programs/marginfi/src/errors.rs
```diff
@@ -208,6 +208,8 @@ pub enum MarginfiError {
     DailyWithdrawalLimitExceeded,
     #[msg("Cannot set daily withdrawal limit to zero")] // 6102
     ZeroWithdrawalLimit,
+    #[msg("Account is frozen")] // 6103
```
IliaZyrin: Nitpick, but I would add a bit more context here. At least:
```suggestion
    #[msg("Account is frozen by the group admin")] // 6103
```

Otherwise it looks like if it could be caused by the user.

programs/marginfi/src/state/marginfi_account.rs
IliaZyrin: I agree with the nit and I think it will actually allow to get rid of the below function. This should handle all cases if I'm not mistaken:
```
pub fn is_signer_authorized(
    marginfi_account: &MarginfiAccount,
    group_admin: Pubkey,
    signer: Pubkey,
    allow_receivership: bool,
) -> bool {
    if allow_receivership && marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP) {
        return true;
    }
    
    if marginfi_account.get_flag(ACCOUNT_FROZEN) {
        return group_admin == signer;
    }

    marginfi_account.authority == signer
}
```

programs/marginfi/src/instructions/marginfi_account/freeze.rs
```diff
@@ -0,0 +1,58 @@
...
+    check_eq!(
+        group.admin,
+        ctx.accounts.admin.key(),
+        MarginfiError::Unauthorized
+    );
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    if frozen {
+        marginfi_account.set_flag(ACCOUNT_FROZEN, true);
+    } else {
+        marginfi_account.unset_flag(ACCOUNT_FROZEN, false);
```
IliaZyrin: We don't want to print message when the account is unfrozen?  Just making sure it's intended.

tests/15_accountFreeze.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+    const mintTx = new Transaction();
+    mintTx.add(
+      createMintToInstruction(
+        ecosystem.usdcMint.publicKey,
+        freezeUser.usdcAccount,
+        wallet.publicKey,
+        200 * 10 ** ecosystem.usdcDecimals
+      )
+    );
+    mintTx.add(
```
IliaZyrin, line 97: Duplicated ix

programs/marginfi/src/instructions/marginfi_account/freeze.rs
IMEF-FEMI: oh sorry that was an oversight 😁

tests/15_accountFreeze.spec.ts
IMEF-FEMI: They are actually both different instructions:
1. mint for freeze user
2. mint for group admin

Anchor.toml
```diff
@@ -36,7 +36,7 @@ wallet = "~/.config/solana/id.json"
 
 # (remove RUST_LOG= to see bankRun logs)
 [scripts]
-test = "RUST_LOG= yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/*.spec.ts --exit --require tests/rootHooks.ts"
+# test = "RUST_LOG= yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/*.spec.ts --exit --require tests/rootHooks.ts"
```
IliaZyrin: Not a big deal but I would revert the changes in this file so that by default we still run the whole suite of tests.

IliaZyrin (approved):
LGTM, but we will need to apply the same checks to Drift and Solend ixes as well. Not sure what will be merged first though.

PR #457, opened 2025-12-04
0.1.6 minor audit changes

jgur-psyops:
Minor audit fixes for 0.1.6

## P1 (Multi-point curves)
P1-L-01: fixed, now validates curve during `lending_pool_configure_bank_interest_only`

P1-I-01/02: acknowledged, there aren't any banks with these problematic utilization rates in prod, and since we will remove the migrate_curve ix in 1.7 after we migrate all prod banks, we don't think it will matter.

P1-I-03: fixed comment

## P2 (Fixed price banks)

P2-L-01: acknowledged. In most cases we will not set the fixed price to 0, however there are occasionally assets that become genuinely worthless (i.e., they have become completely illiquid). In that case, it could be appropriate to give borrowers a windfall and allow them to treat the position as worthless. We might do this when a bank has a nominal borrowed balance of e.g. $100, like some isolated banks do, that wouldn't merit the manual effort to deleverage it first.

P2-I-01: fixed. We are trying to cut back on the edge-case error codes, having over 100, but Unauthorized probably works better here!

P2-I-02: fixed, removed

## P3 (Deleverage)

P3-M-01: there's no great way to block an attacker from doing this on purpose by spamming e.g. thousands of accounts with balances below `ZERO_AMOUNT_THRESHOLD`, but this is merely annoying as we can still unwind these with manual repays. We expect that in all cases when we want to tokenlessly deleverage a bank, it has first been in reduce-only for a long time (which blocks new borrows), so an attacker would have to plan this some months in advance. Increased the threshold to 10x `ZERO_AMOUNT_THRESHOLD` to account for the  likely scenario with some accidentally left-behind dust. Will hold off on mitigating the troll scenario for now, if someone were to actually waste their time and rent to do it, we would probably just read the balance that the bank actually ended up with and update this ix to just set the threshold to that amount!

P3-L-01: Added the check to deposit, although we assume that banks in `TOKENLESS_REPAYMENTS_ALLOWED` will always be in reduce only mode already.

P3-I-01: Fixed. To avoid having to pass the liquidation account to withdraw, adds a new unique flag to the deleverage process which can be checked during withdraw.

## P4 (Other Scope)

P4-I-01: fixed, now uses the `program_fees_enabled` helper. There are no arena banks in prod currently.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/deposit.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/deposit.rs
@@ -16,8 +16,9 @@ use anchor_lang::solana_program::clock::Clock;
 use anchor_lang::solana_program::sysvar::Sysvar;
 use anchor_spl::token_interface::{TokenAccount, TokenInterface};
 use fixed::types::I80F48;
-use marginfi_type_crate::types::{
-    Bank, MarginfiAccount, MarginfiGroup, ACCOUNT_DISABLED, ACCOUNT_IN_RECEIVERSHIP,
+use marginfi_type_crate::{
+    constants::TOKENLESS_REPAYMENTS_ALLOWED,
+    types::{Bank, MarginfiAccount, MarginfiGroup, ACCOUNT_DISABLED, ACCOUNT_IN_RECEIVERSHIP},
 };
 
 /// 1. Accrue interest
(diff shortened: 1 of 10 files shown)


IliaZyrin (approved):
Looks good!

IMEF-FEMI:
Everything looks good from here too 😁

PR #458, opened 2025-12-06
Emode copy instruction

jgur-psyops:
A simple convenience instruction to copy the emode settings from one bank to another, so we can e.g. copy the emode settings from some other LST when we create a new one.

### DIFF
--- a/programs/marginfi/tests/admin_actions/setup_bank.rs
+++ b/programs/marginfi/tests/admin_actions/setup_bank.rs
@@ -23,6 +23,7 @@ use marginfi_type_crate::{
 };
 use pretty_assertions::assert_eq;
 use solana_program_test::*;
+use solana_sdk::signature::Keypair;
 use solana_sdk::signer::Signer;
 use solana_sdk::{
     clock::Clock, instruction::Instruction, pubkey::Pubkey, transaction::Transaction,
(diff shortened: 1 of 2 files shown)


programs/marginfi/src/instructions/marginfi_group/emode_clone.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    check!(
+        ctx.accounts.signer.key() == group.admin || ctx.accounts.signer.key() == group.emode_admin,
+        MarginfiError::Unauthorized
+    );
+
+    let source_bank = ctx.accounts.copy_from_bank.load()?;
+    let mut destination_bank = ctx.accounts.copy_to_bank.load_mut()?;
+
+    destination_bank.emode = source_bank.emode;
```
IMEF-FEMI, line 17: I'm thinking we might need a validation here using say maybe 

`destination_bank.emode.validate_entries_with_liability_weights(
      &destination_bank.config,
      group.emode_max_init_leverage,
      group.emode_max_maint_leverage,
  )?;`

 to ensure the destination banks weight stays within the max leverage limits.... or what do you think?

programs/marginfi/src/lib.rs
```diff
@@ -200,7 +200,8 @@ pub mod marginfi {
         marginfi_group::lending_pool_set_fixed_oracle_price(ctx, price)
     }
 
-    /// (emode_admin only)
+    /// (emode_admin only) Copies emode settings from one bank to another. Useful when applying
```
IliaZyrin: The comment should be moved below, I believe

IliaZyrin (approved):

programs/marginfi/src/instructions/marginfi_group/emode_clone.rs
IMEF-FEMI: @jgur-psyops 👋

jgur-psyops: I think it's fine, the only way a source bank can have some emode setting is if it initially passed the validation in the first place!

PR #460, opened 2025-12-12
forward-allocating bump heap allocator

IMEF-FEMI:
Summary:
  Introduces a custom forward-allocating bump heap allocator with memory recycling capabilities, increasing MAX_KAMINO_POSITIONS from 8 to 16.

  Key Changes:
  - New allocator.rs: Forward-allocating bump allocator with heap_pos()/heap_restore() for memory recycling between oracle price loads
  - Increased position limit: MAX_KAMINO_POSITIONS bumped from 8 → 16
  - Updated liquidation instructions: liquidate, liquidate_start, liquidate_end, and handle_bankruptcy now use heap-efficient health calculations
  - New test k18: Validates liquidation works correctly with 16 total positions (15 Kamino deposits + 1 borrow)

  Why:
  Previously, accounts with many Kamino positions risked becoming unliquidatable due to CU/heap memory exhaustion. The custom allocator recycles heap memory (~3KB per position) so 16 positions fit within 32KB without requiring requestHeapFrame.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
@@ -29,11 +29,17 @@ pub fn log_pool_info(bank: &Bank) {
         conf.asset_tag
     );
     msg!(
-        "oracle conf {:?} age: {:?} flags: {:?}",
-        conf.oracle_max_confidence,
+        "oracle setup: {:?} conf: {:.4} age: {:?} flags: {:?}",
+        conf.oracle_setup as u8,
+        u32_to_centi(conf.oracle_max_confidence).to_num::<f64>(),
         conf.oracle_max_age,
         bank.flags as u8
     );
+    for (i, key) in conf.oracle_keys.iter().enumerate() {
+        if *key != Pubkey::default() {
+            msg!("  oracle key {}: {:?}", i, key);
+        }
+    }
     let interest = conf.interest_rate_config;
     msg!(
         "Insurance fixed: {:?} ir: {:?}",


Henry-E:
There's a more comprehensive set of realistic tests of combinations of positions and mints and other things that @IliaZyrin identified that we need to test and get a realistic view of first before doing something like this imo.

programs/marginfi/src/allocator.rs
```diff
@@ -0,0 +1,247 @@
...
+Code (simplified):
+    pos = self.start + self.len;        // Start at top (0x300008000)
+    pos = pos.saturating_sub(size);     // Move down
+    pos &= !(align - 1);                // Align
+```
+
+## Why Backward Allocation Breaks with Extended Heap
+
+If we simply compile program with increased HEAP_LENGTH without changing direction
+and or not using `requestHeapFrame` (for functions that fit in 32 KiB), we get this situation:
```
jgur-psyops: this is interesting I didn't know the forward allocation was incompatible with requestheapframe, is there a demo or something to show this?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -822,6 +823,640 @@ impl<'info> RiskEngine<'_, 'info> {
...
+///
+/// - `marginfi_account`: The account to calculate health for
+/// - `remaining_ais`: Remaining accounts containing banks and oracles
+/// - `requirement_type`: Initial, Maintenance, or Equity requirement
+/// - `health_cache`: Optional cache to populate with results
+///
+/// ## Returns
+///
+/// (total_assets, total_liabilities) weighted according to requirement_type
+pub fn get_health_components_with_heap_reuse<'info>(
```
jgur-psyops: why not heap restore the default behavior instead of its own function, unless there is some reason we would ever call the non heap restoring version ?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -822,6 +823,640 @@ impl<'info> RiskEngine<'_, 'info> {
...
+        let end_idx = oracle_ai_idx + num_accounts - 1;
+        require_gte!(
+            remaining_ais.len(),
+            end_idx,
+            MarginfiError::WrongNumberOfOracleAccounts
+        );
+        let oracle_ais = &remaining_ais[oracle_ai_idx..end_idx];
+
+        // Create oracle adapter (heap allocation happens here)
+        let price_adapter_result = OraclePriceFeedAdapter::try_from_bank(&bank, oracle_ais, &clock);
```
jgur-psyops, line 426: did we experiment at all with how much heap the oracle call here actually uses in the kamino (like did we confirm it's 3kish bytes?)

maybe there is a way to cut back the heap usage?

programs/marginfi/src/allocator.rs
IMEF-FEMI: I've updated the doc to explain this better... but a similar implementation can be seen here:
[Squads allocator.rs](https://github.com/Squads-Grid/external-signature-program/blob/main/src/allocator.rs)

programs/marginfi/src/state/marginfi_account.rs
IMEF-FEMI: i'm going updated other instructions to use the heap versions shortly... wanted to first test the outcome for the current and ensure no breaking changes and vulnerabilities

programs/marginfi/src/state/marginfi_account.rs
IMEF-FEMI: the ~3k ish bytes is just an over-exaggeration (worst case)... after testing with logs, the usage usually doesn't exceed ~128 bytes (and thats for the switchboard oracles)

Henry-E:
It's not a solution as such. It's more like there's a lot more factors to consider than just memory issues when scaling to >8 integration assets.

Things that would be a better use of your time than this
- strip out anything that doesn't use bankrun and replace it with bankrun in the current typescript testing framework. The setup takes 40 - 60 seconds each time partly because of mocks program txes and then also some other tests that still use non-bank run testing. It would be much appreciated if you could just strip out and replace all that with bankrun transactions.
- Once that's done and everything in the testing framework is on bankrun, that's when we can sit down and discuss what a proper comprehensive testing framework for scaling past 8 assets looks like. As mentioned earlier there's a few different aspects which need to also be tested which aren't being covered, e.g. different mints per bank. And also none of the integrations have tests with switchboard or the new liquidation approach. This is something we should sit down and discuss a plan for though.

programs/marginfi/src/state/marginfi_account.rs
jgur-psyops: yeh it's fine if you're documenting worst case, just want to confirm the size at worst for our napkin math of memory usage.

If you can print the heap pointer before/after should be able to precisely doc the worst case.

programs/marginfi/src/allocator.rs
```diff
@@ -0,0 +1,317 @@
...
+  - Holding references to heap data across restore boundary
+  - Accessing Box/Vec/String contents after restore
+  - Storing heap pointers in struct fields that survive restore
+
+Example of SAFE usage:
+```rust
+let checkpoint = heap_pos();
+{
+    let temp_vec: Vec<u8> = vec![1, 2, 3];          // Heap allocation
+    let sum: u64 = temp_vec.iter().sum();           // Copy to stack!
```
IliaZyrin: This is confusing comment, because we're not _copying_ anything here, but rather putting a _new value_ (`sum`) on stack.

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -195,16 +196,18 @@ pub fn lending_account_withdraw<'info>(
...
             &mut Some(&mut health_cache),
-        );
-        risk_result?;
+        )?;
         health_cache.program_version = PROGRAM_VERSION;
 
-        // Note: in flashloans, risk_engine is None, and we skip the cache price update.
-        maybe_price = risk_engine.and_then(|e| e.get_unbiased_price_for_bank(&bank_pk).ok());
+        // Fetch unbiased price for cache update
+        let bank = bank_loader.load()?;
```
IliaZyrin: So the logic is now exactly the same as in the `else` below. I suggest we just move it out the scope to not duplicate.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -846,6 +847,653 @@ impl<'info> RiskEngine<'_, 'info> {
...
+
+        // Log heap usage per position for measurement/debugging
+        // Measured results: Pyth ~64 bytes, Switchboard ~128 bytes per position
+        #[cfg(target_os = "solana")]
+        {
+            let heap_after_oracle = heap_pos();
+            let heap_used = heap_after_oracle.saturating_sub(heap_checkpoint);
+            msg!(
```
IliaZyrin: We should remove this before merging. Or, which is probably better, change it to `debug!()`.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -846,6 +847,653 @@ impl<'info> RiskEngine<'_, 'info> {
...
+/// Check bankruptcy condition with heap reuse optimization.
+///
+/// This is equivalent to `RiskEngine::check_account_bankrupt` but uses heap reuse
+/// to process positions one at a time.
+pub fn check_account_bankrupt<'info>(
+    marginfi_account: &MarginfiAccount,
+    remaining_ais: &'info [AccountInfo<'info>],
+    health_cache: &mut Option<&mut HealthCache>,
+) -> MarginfiResult {
+    let (equity_assets, equity_liabs) = get_health_components(
```
IliaZyrin, line 582: Missing check for flashloan here? Also, should we just put it directly inside `get_health_components` if it's only used by the callers of this function?

scripts/build-program.sh
```diff
@@ -13,9 +13,9 @@ fi
 if [ "$cluster" = "mainnet" ]; then
     features="--features mainnet-beta"
```
IliaZyrin: Not using custom-heap in mainnet?

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -183,8 +182,10 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
     // Note: Kamino deposits sans-LUT are severely limited.
     const depositsPerTx = 2;
 
-    // Only deposit into the first MAX_KAMINO_POSITIONS banks
-    for (let i = 0; i < MAX_KAMINO_POSITIONS; i += depositsPerTx) {
+    // NOTE: We only deposit into 8 banks for this test suite since we don't use a LUT.
```
IliaZyrin: Let's just rename the const above to be `KAMINO_POSITIONS` and use it everywhere (with the value of `8`, as before). Otherwise it creates confusion since `MAX_KAMINO_POSITIONS` is not even used.

tests/k17_kaminoLimitsMultiDeposit.spec.ts
```diff
@@ -980,9 +980,9 @@ describe("k17: Limits test - 8 Kamino + 7 regular TOKEN_A deposits, liquidation
       state: lutState,
     });
 
-    // Add compute budget
+    // Add compute budget
```
IliaZyrin: formatting

tests/k17_kaminoLimitsMultiDeposit.spec.ts
```diff
@@ -90,8 +90,8 @@ import Decimal from "decimal.js";
 import { wrappedI80F48toBigNumber } from "@mrgnlabs/mrgn-common";
 import { assert } from "chai";
 
-const MAX_KAMINO_DEPOSITS = 8; // Maximum Kamino positions per account
-const NUM_KAMINO_BANKS_FOR_TESTING = 9; // Create 9 banks to test liquidator limit
+
+const NUM_KAMINO_BANKS_FOR_TESTING = 17;
```
IliaZyrin: Do we really need 17 here? 15 is not enough?

Please update/delete the outdated comments throughout this file. E.g. they mention 9 banks and non-actual [expectations](https://github.com/mrgnlabs/marginfi-v2/pull/460/files#diff-5e27b489de57debc4feb34b209f53ca66ad6c4ae619538aa41011c9b02e00c76R10).

tests/k18_sixteenKaminoLiquidation.spec.ts
```diff
@@ -0,0 +1,1006 @@
...
+} from "@kamino-finance/klend-sdk";
+import { TOKEN_PROGRAM_ID } from "@solana/spl-token";
+import { ComputeBudgetProgram } from "@solana/web3.js";
+import Decimal from "decimal.js";
+import { wrappedI80F48toBigNumber } from "@mrgnlabs/mrgn-common";
+import { assert } from "chai";
+import { CONF_INTERVAL_MULTIPLE, ORACLE_CONF_INTERVAL } from "./utils/types";
+
+/** Maximum Kamino positions per account - now 16 with custom allocator */
+const MAX_KAMINO_POSITIONS = 16;
```
IliaZyrin: 1. This is unused. Let's delete it.
2. I don't think we need a [separate const](https://github.com/mrgnlabs/marginfi-v2/pull/460/files#diff-f8ac6f724a787aea5589f693d6e23290a910a7036c0e7ea0b0b1411708db2326R173) for Kamino, given that the maximum Kamino positions is now equal to the maximum positions in general. But up to you, if you want to keep it for documentation purposes.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -846,6 +847,653 @@ impl<'info> RiskEngine<'_, 'info> {
...
+}
+
+/// Check pre-liquidation condition with heap reuse optimization.
+///
+/// This is equivalent to `RiskEngine::check_pre_liquidation_condition_and_get_account_health`
+/// but uses heap reuse to process positions one at a time, enabling support for accounts
+/// with up to 16 positions.
+///
+/// Returns (account_health, assets, liabilities) if the account is liquidatable.
+pub fn check_pre_liquidation_condition_and_get_account_health<'info>(
```
IliaZyrin, line 509: General comment about all these new functions. Should we remove their non-heap-optimized versions now? Or is there a reason for both variants to stay?

programs/marginfi/src/state/marginfi_account.rs
IMEF-FEMI: was gonna add to top level instructions but skipped lol... i'll put it back here for now

programs/marginfi/src/state/marginfi_account.rs
IMEF-FEMI: just did a bit of cleanup... we now have just the new functions... all non heap versions have been removed

Thanks

programs/marginfi/src/errors.rs
```diff
@@ -234,8 +234,8 @@ pub enum MarginfiError {
     KaminoReserveValidationFailed, // 6210
     #[msg("Invalid oracle setup: only KaminoPythPush and KaminoSwitchboardPull are supported")]
     KaminoInvalidOracleSetup, // 6211
-    #[msg("Maximum Kamino positions limit exceeded (max 8 positions per account)")]
-    KaminoPositionLimitExceeded, // 6212
+    #[msg("Deprecated: Kamino position limit no longer enforced separately")]
+    KaminoPositionLimitExceeded, // 6212 - kept for error code stability
```
IliaZyrin: We usually just rename it to smth like "VacatedError1" or "DummyError1" and then when the time comes to introduce a new error, we put it in place.

scripts/build-program.sh
```diff
@@ -11,11 +11,11 @@ if [ -z "$program_lib_name" ] || [ -z "$cluster" ]; then
 fi
 
 if [ "$cluster" = "mainnet" ]; then
-    features="--features mainnet-beta"
+    features="--features,mainnet-beta,custom-heap"
```
IliaZyrin:
```suggestion
    features="--features mainnet-beta,custom-heap"
```

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -303,11 +237,13 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
...
     const withdrawBank = kaminoBanks[0];
-    const replacementBank = kaminoBanks[MAX_KAMINO_POSITIONS];
+    // Use bank 8 (index 8) as replacement - this is the first unused bank
+    // (we only deposited into banks 0-7)
+    const replacementBank = kaminoBanks[8];
 
     // Remaining accounts exclude the bank being closed
     const remainingPositions = [];
-    for (let i = 1; i < MAX_KAMINO_POSITIONS; i++) {
+    for (let i = 1; i < 8; i++) {
```
IliaZyrin:
```suggestion
    for (let i = 1; i < KAMINO_POSITIONS; i++) {
```

tests/k14_limitsWithKamino.spec.ts
```diff
@@ -303,11 +237,13 @@ describe("k14: Limits on number of accounts, with Kamino and emode", () => {
     const farmState = farmAccounts.get(A_FARM_STATE);
 
     const withdrawBank = kaminoBanks[0];
-    const replacementBank = kaminoBanks[MAX_KAMINO_POSITIONS];
+    // Use bank 8 (index 8) as replacement - this is the first unused bank
+    // (we only deposited into banks 0-7)
+    const replacementBank = kaminoBanks[8];
```
IliaZyrin:
```suggestion
    const replacementBank = kaminoBanks[KAMINO_POSITIONS];
```

IliaZyrin (approved):
LGTM, just a few nitpicks.

PR #461, opened 2025-12-13
Additional tests for drift

IliaZyrin:
This one:
- fixes wtihdraw_all behavior (WARNING: change in the program logic!),
- removes some redundant/non-testing-anything tests,
- improves comments around tricky Drift-related logic,
- adds more checks in existing tests and makes them precise where possible,
- adds a few new tests.

### DIFF
--- a/tests/d04_userDeposits.spec.ts
+++ b/tests/d04_userDeposits.spec.ts
@@ -25,7 +25,6 @@ import { deriveSpotMarketPDA } from "./utils/pdas";
 import {
   getSpotMarketAccount,
   getUserPositions,
-  calculateUtilizationRate,
   calculateInterestRate,
   isDriftPositionBorrow,
   isDriftPositionDeposit,
@@ -34,6 +33,7 @@ import {
   DRIFT_UTILIZATION_PRECISION,
   TOKEN_A_SCALING_FACTOR,
   USDC_SCALING_FACTOR,
+  calculateUtilization,
 } from "./utils/drift-utils";
 
 describe("d04: Drift - User Deposits and Borrows", () => {
@@ -339,25 +339,23 @@ describe("d04: Drift - User Deposits and Borrows", () => {
     assert.ok(isDriftPositionBorrow(usdcPosition));
     assert.ok(isDriftPositionDeposit(tokenAPosition));
 
-    const usdcUtilization = calculateUtilizationRate(
-      usdcMarketAfter.depositBalance,
-      usdcMarketAfter.borrowBalance
+    const usdcUtilization = calculateUtilization(
+      usdcMarketAfter
     );
     console.log(
       "USDC utilization: " +
-        (usdcUtilization / DRIFT_UTILIZATION_PRECISION) * 100 +
+        (usdcUtilization.mul(new BN(100)).div(DRIFT_UTILIZATION_PRECISION).toNumber()) +
         "%"
     ); // 10%
 
     const usdcInterestRate = calculateInterestRate(
-      usdcUtilization,
-      usdcMarketAfter.optimalUtilization,
-      usdcMarketAfter.optimalBorrowRate,
-      usdcMarketAfter.maxBorrowRate
+      usdcMarketAfter,
+      new BN(0),
+      usdcUtilization
     );
     console.log(
       "USDC InterestRate: " +
-        (usdcInterestRate / DRIFT_UTILIZATION_PRECISION) * 100 +
+        (usdcInterestRate.mul(new BN(100)).div(DRIFT_UTILIZATION_PRECISION).toNumber()) +
         "%"
     ); // 1%
   });
@@ -453,25 +451,23 @@ describe("d04: Drift - User Deposits and Borrows", () => {
     assert.ok(isDriftPositionDeposit(usdcPosition));
     assert.ok(isDriftPositionBorrow(tokenAPosition));
 
-    const tokenAUtilization = calculateUtilizationRate(
-      tokenAMarketAfter.depositBalance,
-      tokenAMarketAfter.borrowBalance
+    const tokenAUtilization = calculateUtilization(
+      tokenAMarketAfter
     );
     console.log(
       "Token A utilization: " +
-        (tokenAUtilization / DRIFT_UTILIZATION_PRECISION) * 100 +
+        (tokenAUtilization.mul(new BN(100)).div(DRIFT_UTILIZATION_PRECISION).toNumber()) +
         "%"
     ); // 10%
 
     const tokenAInterestRate = calculateInterestRate(
-      tokenAUtilization,
-      tokenAMarketAfter.optimalUtilization,
-      tokenAMarketAfter.optimalBorrowRate,
-      tokenAMarketAfter.maxBorrowRate
+      tokenAMarketAfter,
+      new BN(0),
+      tokenAUtilization
     );
     console.log(
       "Token A InterestRate: " +
-        (tokenAInterestRate / DRIFT_UTILIZATION_PRECISION) * 100 +
+        (tokenAInterestRate.mul(new BN(100)).div(DRIFT_UTILIZATION_PRECISION).toNumber()) +
         "%"
     ); // 400%
   });
--- a/tests/d12_driftHarvestReward.spec.ts
+++ b/tests/d12_driftHarvestReward.spec.ts
@@ -524,7 +524,10 @@ describe("d12: Drift Harvest Reward", () => {
       false
     );
 
-    const userTokenBAfter = await getTokenBalance(bankRunProvider, destinationTokenAccount);
+    const userTokenBAfter = await getTokenBalance(
+      bankRunProvider,
+      destinationTokenAccount
+    );
     assertBNEqual(depositBAmount, userTokenBAfter);
   });
 
--- a/tests/d13_oraclePriceConversionTest.spec.ts
+++ b/tests/d13_oraclePriceConversionTest.spec.ts
@@ -32,11 +32,12 @@ import {
   makePulseHealthIx,
   USDC_MARKET_INDEX,
   TOKEN_A_MARKET_INDEX,
-  calculateUtilizationRate,
   calculateInterestRate,
   DRIFT_UTILIZATION_PRECISION,
   USDC_SCALING_FACTOR,
   TOKEN_A_SCALING_FACTOR,
+  ONE_YEAR,
+  ONE,
 } from "./utils/drift-utils";
 import { refreshPullOraclesBankrun } from "./utils/bankrun-oracles";
 import { makeUpdateSpotMarketCumulativeInterestIx } from "./utils/drift-sdk";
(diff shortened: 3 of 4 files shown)


programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -94,20 +94,24 @@ pub fn drift_withdraw<'info>(
...
-            // Sanity check
-            require_gte!(
-                scaled_balance,
-                actual_scaled_balance_decrememt,
-                MarginfiError::MathError
-            );
+            // Handle edge case where expected_scaled_balance_change is exactly 1 more than scaled_decrement
+            // This happens when depositing and immediately withdrawing with no interest earned
+            // due to Drift rounding up the scaled decrement. We round down manually.
+            if expected_scaled_balance_change != scaled_balance + 1 {
```
IliaZyrin: This was necessary to fix [withdraw_all](https://github.com/mrgnlabs/marginfi-v2/pull/461/files#diff-435b4a3a6400a42270e00d0683f9db1241719bf4d125869647afca8f7f6064dfR47).

cc @jgur-psyops @Henry-E

programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -116,13 +120,13 @@ pub fn drift_withdraw<'info>(
             // This happens when depositing and immediately withdrawing with no interest earned
             // due to Drift rounding up the scaled decrement. We round down manually.
             if scaled_decrement == asset_shares + 1 {
-                token_amount = drift_spot_market.get_withdraw_token_amount(asset_shares)?;
```
IliaZyrin, line 119: This was necessary to fix the ["withdraw_all" case](https://github.com/mrgnlabs/marginfi-v2/pull/461/files#diff-435b4a3a6400a42270e00d0683f9db1241719bf4d125869647afca8f7f6064dfR113) when `withdraw_all` flag is `false`, e.g. when you withdraw 200 USDC and that's your **whole** deposit.

cc @jgur-psyops @Henry-E 

I also think that this whole chunk was a **no-op**, i.e. it was not changing anything (at least most of the time, but probably always, still trying to come up with the values where it's meaningful):
```
token_amount = drift_spot_market.get_withdraw_token_amount(asset_shares)?;
scaled_decrement = drift_spot_market.get_scaled_balance_decrement(token_amount)?;
```

IliaZyrin: Ok, I see now, **it is useful** but only if cumulative interest is not round, which is why it was passing [here](https://github.com/mrgnlabs/marginfi-v2/pull/461/files#diff-afa6e3c00b336a2dcf20a901cea5e2e1c93783da75442af391b1db17f982353eL915). But it was not enough for the "clean" bank state.

tests/d08_driftWithdraw.spec.ts
```diff
@@ -301,40 +247,43 @@ describe("d08: Drift Withdraw Tests", () => {
...
+      sharesBefore.sub(scaledBalanceDecrease)
+    );
   });
 
-  it("(user 0) Try to withdraw more than deposited - should fail", async () => {
+  it("(user 0) Tries to withdraw more than deposited - should fail", async () => {
     const user = users[0];
-    const excessiveAmount = new BN(1000 * 10 ** ecosystem.tokenADecimals);
+    // Due to Drift rounding on previous withdrawal (2 Token A => 2.00000001), there is now only (3 * 10 ** ecosystem.tokenADecimals - 1) tokens left.
+    // TODO: I don't like this though!
```
IliaZyrin: To me it seems buggy when e.g. user deposits 5 tokens, then withdraws 2 of them and then **cannot** withdraw 3. But maybe it's not even possible/very rare in the real life, given the interest generated. What do you think @Henry-E ?

cc @jgur-psyops

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: skipping the sanity check does not seem like the correct solution here, will come back to this.

Can you be clear in your explanation of what the issue with withdraw_all was and why this was needed to fix it. That would be helpful. I can figure it out from first principles but it would have been quicker to read a short explanation.

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: I feel like I've talked to @jgur-psyops before about this and it is intended behaviour that withdrawing the full amount without calling `withdraw_all=True` shouldn't necessarily close the bank deposit on the marginfi account. 

I also don't like blanket withdrawing all asset shares here. It's always token amounts.

Maybe we can add this behaviour in a later update when it's getting audited

programs/marginfi/src/instructions/drift/withdraw.rs
IliaZyrin: I wrote a test to showcase what was not working and referred it [above](https://github.com/mrgnlabs/marginfi-v2/pull/461/files#diff-435b4a3a6400a42270e00d0683f9db1241719bf4d125869647afca8f7f6064dfR47).

Basically the plain and simple `withdraw_all` didn't work precisely because of this unconditional sanity check - it was failing because `scaled_balance` was less than `expected_scaled_balance_change` (which is now turned into condition: `if expected_scaled_balance_change != scaled_balance + 1`). That is because of the Drift rounding, which leads to trying to withdraw 1 unit more than the total shares you have. I hope it's clearer now.

programs/marginfi/src/instructions/drift/withdraw.rs
IliaZyrin: It's not that it doesn't close the account, it's just that it didn't work at all (i.e. the transaction was failing) - the test proving that was referred [above](https://github.com/mrgnlabs/marginfi-v2/pull/461/files#diff-435b4a3a6400a42270e00d0683f9db1241719bf4d125869647afca8f7f6064dfR113).

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: So yeh, the point of the sanity check is literally that we're not giving back the user more tokens / scaled balance than they have ownership over because of drift's scaled balance rounding up. If you skip this check then this is also a loss of funds of the same variant as the loss of funds bug introduced below.

will discuss both changes in a single follow up message. But both changes can probably be reverted.

programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -93,29 +93,51 @@ pub fn drift_withdraw<'info>(
...
+            let mut token_amount = drift_spot_market.get_withdraw_token_amount(scaled_balance)?;
+            let mut expected_scaled_balance_change =
                 drift_spot_market.get_scaled_balance_decrement(token_amount)?;
 
+            // If rounding would require more scaled balance than we have, reduce the withdraw
+            // amount by 1 base unit to keep the init deposit buffer intact. In practice, this means
+            // if you deposit and immediately withdraw_all, you will lose one lamport, which will be
+            // trapped in the Drift init buffer forever.
+            if expected_scaled_balance_change > scaled_balance && token_amount > 0 {
+                token_amount = token_amount.saturating_sub(1);
```
jgur-psyops, line 105: Here is the substantive change: we now short change you a lamport if necessary to preserve the buffer. Your balance still closes as expected so accounting should still make sense. If you look at d08, you will see the result is that now each withdraw_all is shortchanged by 1 lamport

tests/d08_driftWithdraw.spec.ts
```diff
@@ -398,56 +340,68 @@ describe("d08: Drift Withdraw Tests", () => {
...
-      spotPositionBefore.scaledBalance.toString()
+    const marginfiBalanceBefore =
+      marginfiAccountBefore.lendingAccount.balances.find(
+        (b) => b.bankPk.equals(driftTokenABank) && b.active === 1
+      );
+
+    const assetSharesBefore = new BN(
+      wrappedI80F48toBigNumber(marginfiBalanceBefore.assetShares).toString()
+    );
+    const maxTokenAmount = scaledBalanceToTokenAmount(
```
jgur-psyops, line 378: note that we no longer need the "subtract 1 for no reason" kludge in `scaledBalanceToTokenAmount` and can now naturally withdraw the computed max (which is just 1 lamport short of your actual balance).

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: Why not just copy the existing maths that's done below.

```
else if scaled_decrement == asset_shares + 1 {
                token_amount = drift_spot_market.get_withdraw_token_amount(asset_shares)?;
                scaled_decrement = drift_spot_market.get_scaled_balance_decrement(token_amount)?;
            }
```

In the specific edge case where someone attempts to withdraw their full deposit amount and no interest has been earned, drift will round up by 1 unit of scaled balance (not decrease it by a full token unit / lamport).

- important to limit to specifically only the case where `scaled_decrement` is `asset_shares + 1` and no other cases
- `get withdraw token amount( asset shares)` does the rounding down for you. Maybe this could communicated better with function naming or something but the flow is `asset shares` -> `floored token amount` -> `scaled balance decrement`. The rounding down is done automatically and we go back from token amount to scaled balance / asset shares to give the actual amount of scaled decrement that occurs. Unless this maths is somehow wrong and the auditors missed it, in which case we should be subtracting lamports here as well then by default.
- It just matches the drift maths 1:1 this way, rather than a) arbitrarily subtracting a token unit b) in a broad set of cases `scaled balance decrement > asset shares` vs. more exactly when `scaled balance decrement == asset shares + 1`

Aside from that I think we're on the same page. 
- Dust behaviour in regular full withdrawal case continues as before
- @IliaZyrin correctly identified that immediate withdraw all with no interest earned case wasn't covered and should be fixed

tests/d08_driftWithdraw.spec.ts
Henry-E: What's changed in the code base such it's no longer necessary to subtract by 1 to safely withdraw? If the tests work fine with that change now, I'm guessing it would have worked fine with that prior to this as well, just because I don't know where edge case was catching. It's not that important I guess but just didn't notice anything else changing in the code base in this PR that makes this change make sense. But still better to remove it regardless.

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: If we reuse the existing maths + keep the sanity check that we're not returning more scaled balance than asset shares available, then I guess I would approve of this change. But I still think we need to get every rust change audited. It's a weird edge case for sure and it's not expected behaviour. But it is an edge case that won't actively block people unless they create a fresh deposit + call withdraw all in the same transaction. Assuming that we only list banks that are actively earning interest each block.

jgur-psyops: it will be extremely difficult to get this audited due to the confusing diff structure and our delayed timeline so we should be confident of this change internally.

In the withdraw all path we don't need to recompute token amount because it was already computed from the market exchange rate and shares earlier, unlike the default path where it is user-input by default, so we can just sub 1. in theory the computed amount is at most 1 lamp greater in drifts math. in the default path we must handle bad user inputs (exceeds scaled bal by more than one) but in the withdraw all path we do not since we computed the token amount from shares already.

feel free to try the +1 conditional instead of the > if the tests pass I'm good with either.

This edge case must be fixed so the financial math checks out AND withdraw is always enabled.

tests/d08_driftWithdraw.spec.ts
jgur-psyops: I'm referring to comment that was above the subtraction. if it was for no reason, it must be removed. this one-subtraction was what enabled various withdraws to work with the above bug

Henry-E: Right but that makes no sense because
- `scaled balance to token amount` returns token amount
- `withdraw all` never uses the token amount arg when processing a withdrawal

So there's no overlap in places where fixing the `withdraw all` bug would overlap with checking the correct amount was returned. And because the bug was causing the entire transaction to fail, the entire amount would have been 0, not off by 1.

programs/marginfi/src/instructions/drift/withdraw.rs
Henry-E: > In the withdraw all path we don't need to recompute token amount because it was already computed from the market exchange rate and shares earlier, unlike the default path where it is user-input by default, so we can just sub 1.

That's kind of weak logic tbh. "We don't need to recalculate because we already made an inaccurate calculation earlier". There's no reason for us not match the maths for calculating exactly what the scaled decrement is for a given token amount. Just aribtrarily subbing by 1 token unit even if it's directly correct feels wrong.

> feel free to try the +1 conditional instead of the > if the tests pass I'm good with either.

 Ok great, i'm going to leave this for @IliaZyrin to update with since this is his PR and hopefully this helps with getting a better intuition of what's going on in the underlying maths.

tests/d08_driftWithdraw.spec.ts
IliaZyrin: Yes, Henry, you're correct here, this change is unrelated to withdraw_all fixes. Jon just moved "-1 expectation" on the level of a caller in two places where this function is used. But that does look nicer, so let's keep it.

PR #464, opened 2025-12-19
unit tests for drift and solend integrations

IMEF-FEMI:
This PR adds local unit tests for the drift and solend integrations

### DIFF
--- a/programs/marginfi/src/instructions/drift/local_tests.rs
+++ b/programs/marginfi/src/instructions/drift/local_tests.rs
@@ -118,7 +118,21 @@ mod tests {
         assert_eq!(get_precision_increase(6).unwrap(), EXP_10[13]); // USDC
         assert_eq!(get_precision_increase(9).unwrap(), EXP_10[10]); // SOL
         assert_eq!(get_precision_increase(8).unwrap(), EXP_10[11]); // BTC
+    }
+
+    #[test]
+    fn precision_increase_boundary_cases() {
+        // Maximum supported decimals (19)
+        assert_eq!(
+            get_precision_increase(DRIFT_PRECISION_EXP).unwrap(),
+            EXP_10[0]
+        );
+
+        assert_eq!(get_precision_increase(0).unwrap(), EXP_10[19]); // 10^19
+
         assert!(get_precision_increase(DRIFT_PRECISION_EXP + 1).is_err());
+        assert!(get_precision_increase(20).is_err());
+        assert!(get_precision_increase(100).is_err());
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


programs/marginfi/src/instructions/drift/local_tests.rs
```diff
@@ -0,0 +1,474 @@
...
+
+    fn overflow_raw_for_i64_exact(market: &MinimalSpotMarket) -> i64 {
+        largest_safe_raw_for_i64_exact(market).saturating_add(1)
+    }
+
+    /// Find the largest u64 token amount that won't overflow in scaled_balance_increment.
+    /// Formula: scaled = amount * precision_increase / interest
+    /// Safe when: amount <= u64::MAX * interest / precision_increase
+    fn largest_safe_amount_for_scaled_increment(market: &MinimalSpotMarket) -> u64 {
+        let precision_increase = get_precision_increase(market.decimals).unwrap();
```
jgur-psyops, line 59: I suppose one interesting edge case we might document in the unit tests is that Drift cannot support tokens with more decimals than DRIFT_PRECISION_EXP (19) because there is a checked sub in `get_precision_increase`

jgur-psyops: We can probably optimize the precision increase like so to make it more explicit: 
```
pub fn get_precision_increase(token_decimals: u32) -> Result<u128> {
    if token_decimals > DRIFT_PRECISION_EXP {
        return Err(DriftMocksError::MathError.into());
    }
    Ok(EXP_10[(DRIFT_PRECISION_EXP - token_decimals) as usize])
}
```

jgur-psyops: One thing I am curious about is why we are re-computing what is essentially a static value so many times, why aren't we either caching this on the market or creating a const value like `EXP_10` for precision conversions?

programs/marginfi/src/instructions/drift/local_tests.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    #[test]
+    fn withdraw_token_amount_zero_returns_zero() {
+        let market = spot_market(6, SPOT_CUMULATIVE_INTEREST_PRECISION);
+        assert_eq!(market.get_withdraw_token_amount(0).unwrap(), 0u64);
+    }
+
+    #[test]
+    fn round_trip_deposit_then_withdraw_at_same_interest() {
```
jgur-psyops, line 233: Let's hit this and `round_trip_with_interest_accrual` with some more test cases. Values around u64 max, very small ~zero, etc.

I believe `get_scaled_balance_decrement` has a special case at 0 (it's 0) so `immediate_withdraw_decrement_exceeds_increment_by_one` can demonstrate that as well

programs/marginfi/src/instructions/drift/local_tests.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    #[test]
+    fn integer_division_floors_correctly() {
+        let market = spot_market(6, 12_000_000_000u128); // 1.2x
+
+        assert_eq!(market.adjust_i64(5).unwrap(), 6); // 5 * 1.2 = 6 (exact)
+        assert_eq!(market.adjust_i64(1).unwrap(), 1); // 1 * 1.2 = 1.2, floors to 1
+        assert_eq!(market.adjust_i64(4).unwrap(), 4); // 4 * 1.2 = 4.8, floors to 4
+        assert_eq!(market.get_scaled_balance_increment(1).unwrap(), 833); // 833.33... floors
```
jgur-psyops, line 567: here we can show decrement would be 1 larger (again)

PR #467, opened 2025-12-28
Fix oracle max age display by keeping u16 instead of casting as u8

### LINKED ISSUE #402: Oracle max age display bugged
Case as u8 in `log_pool_info`, causing values over 256 to read incorrectly, should be a u16.

engineerhead:
Fixes #402


jgur-psyops:
Completed in #474

PR #469, opened 2026-01-04
Add support for user orders

abelmarnk:
This PR adds support for user orders as described in #-420

**Relevant Specification**
- Orders are of two kinds take-profit & stop-loss or both.
- A user can open an order containing assets and liabilities and orders currently support only one of both. 
- The first time a balance is used in any order it is assigned a tag which is present in the balance until the balance is closed.
- The order identifies it's balances with those tags.
- When a balance is closed, orders that contains the tag for that balance can also be closed.
- A single balance can be present in several orders at once, when one of them gets executed it may offset other orders and cause them to be closed(e.g in the current implementation the executor/keeper is expected to close the liability(s) involved in the order and once any balance in an order(an order different from the one just executed) is closed then that order is eligible to be closed). 
- The user can also close orders themselves.
- In the case where the user has several orders instead of closing all of them they can mark them to be closed in a single instruction by setting the balance tags for those orders to `0`, the executors/keepers can then close them and claim the rent & order fee(The order fee is not yet implemented).
- Executors execute orders in a similar way to the start-end liquidation flow
- The execution of the order would cause the liabilities to be closed and the asset balance to be decreased while staying above or equal to the trigger price
- Before the order is executed, it is checked that the account is still healthy if not, then it should be liquidated instead(It is possible for the account to be in an unhealthy state, which would be caused by other liabilities not in that order).
- Of course it is also checked that the trigger condition(SL / TP) is satisfied.
- Additionally it is checked that the balances that are not related to the order remain unchanged.
- In the case of a take-profit, if the executor sees the order as `net value > trigger price`, the executor can keep `net - tp` at most, ensuring that value still stays above or equal to the trigger price, it is also checked that this does not make the account unhealthy.

**Scope**
- Update Marginfi Program
	- State
	- Instructions(new)
	- Instructions(old) for compatibility e.g repay, withdraw
- Update CLI
- Update User Guide
- Update Observability IDLs
- Update Test Utils
- Update Tests

cc: @-jgur-psyops

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3633,9 +3633,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.95"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "02b3e5e68a3a1a02aad3ec490a98007cbc13c37cbe84a3cd7b8e406d76e7f778"
+checksum = "5ee95bc4ef87b8d5ba32e8b7714ccc834865276eab0aed5c9958d00ec45f49e8"
 dependencies = [
  "unicode-ident",
 ]


abelmarnk:
Wrong branch.

Issue #420, opened 2025-10-23
Trigger events for position changes


jgur-psyops:
Reminder to contributors: Please base off the latest version branch, and target the consolidated update with your PR, not main. The main branch reflects the latest version on mainnet, and may be 1-2 months stale compared to what's in active developement. Right now, that's 0.1.8-main, and https://github.com/mrgnlabs/marginfi-v2/pull/493

# Summary 

Some integrators would like the ability for users to set a "stop loss" or "take profit": if a position grows to a certain value, allow it to be unwound permissionlessly, returning the funds to the user. 

The implementation might look like this:
* User deposits A worth $100 and borrows B worth $50, a net value of $50.
* Users sets a take-profit at $100
* If EITHER 1) The price of A increases  to $150, 2), the price of B decreases to 0, 3) some combination such that A - B >= $100, the take-profit is activated.
* In (1), we withdraw $50 of A, swap to B, repay the B debt, and the user now has $100 in A
* In (2), we repay the now-worthless B debt.
* In (3), we withdraw as much as we need to zero out the $B debt, otherwise same as (1).

Once activated, anyone can permissionlessly crank the user's account to 1) Repay the B debt, 2) Return A-B to their wallet. It may also make more sense if (2) simply leaves the remaining deposit active, or in a future implementation, even converts A-B into a stable of the user's choosing.

The cranker should earn some fee for their service (a small percentage of profit). Also note that since the cranker is responsible for the A->B swap, they might earn more if they can find favorable slippage.

abelmarnk:
Hey @jgur-psyops, could i take this on?

jgur-psyops:
> Hey [@jgur-psyops](https://github.com/jgur-psyops), could i take this on?

It's relatively challenging but feel free to take a shot! Please use a 0.1.7-main branch as your base, we are expecting to complete this around our 1.8 update (early to mid Feburary)

abelmarnk:
Hey  @jgur-psyops sorry to bother you. 

Could I get your feedback on this approach:-

The approach generalizes the idea a bit it and allows for combinations of assets/liabilities to be tied to a single order (not just a single asset and a liability) as long as the combinations are distinct, it would still work if we allowed just one asset & liability in one combination, is this necessary or should it be one asset and one liability?

**High level(instructions)**
- Add take profit
- Remove take profit
- Add stop loss
- Remove stop loss
- Execute take profit (liquidator)
- Execute stop loss (liquidator)

[...]

PR #470, opened 2026-01-05
Add support for user orders

abelmarnk:
This PR adds support for user orders as described in #420

**Relevant Specification**
- Orders are of three kinds take-profit & stop-loss or both.
- In the case of both the `SL` must be less than the `TP`
- A user sets a slippage in BPS when creating an order.
- A user can open an order containing assets and liabilities and orders currently support only one of both. 
- The first time a balance is used in any order it is assigned a non-zero tag which is present in the balance until the balance is closed.
- The order identifies it's balances with those tags.
- The tags are used as an alternative to storing public keys in the balance which would require 32 bytes, and the tags using 2 bytes each.
- When a balance is closed, orders that contains the tag for that balance can also be closed.
- A single balance can be present in several orders at once, when one of them gets executed it may offset other orders and cause them to be closed(e.g in the current implementation the executor/keeper is expected to close the liability(s) involved in the order and once any balance in an order(an order different from the one just executed) is closed then that order is eligible to be closed). 
- The user can also close orders themselves.
- In the case where the user has several orders instead of closing all of them they can mark them to be closed in a single instruction by setting the balance tags for those orders to `0`, the executors/keepers can then close them and claim the rent & order fee.
- Executors execute orders in a similar way to the start-end liquidation flow
- The execution of the order would cause the liabilities to be closed and the asset balance to be decreased taking the slippage into account.
- The slippage is taken into account differently for both `SL` and `TP`.
- For `TP` the slippage is taken into account by ensuring that the resulting value left is not less than `(TP * (1 - (slippage/MAX_BPS))`
- There also exists a max fee which is in percentage, the max fee allows the keeper to bring the resulting value down to at most  `((V_A - V_L) * (1 - max_fee)`, where `V_A` and `V_L` are the values of assets and liabilities present in the order at the start or execution.
- In the case where the resulting value from the max fee is less than the resulting value from the slippage, the value from the slippage is used instead.
- For the `SL` case, the slippage is enforced by using it on the difference between the value of the assets and the value of the liabilities as seen when starting the execution of the order, i.e it ensures that the value left is not less than `((V_A - V_L) * (1 - (slippage/MAX_BPS))`, where `V_A` and `V_L` are the values of assets and liabilities present in the order at the start or execution.
- In both the `SL` and `TP` case above the executor is allowed to keep whatever is left between the captured value(what they see at the start of execution) and the minimum resulting value which is calculated as described above.
- After the order is executed, it is checked that the account health is as good as or better than it was at the start of execution, and if it is not then the account should at least be healthy, otherwise the order is reverted.
- Additionally it is checked that the balances that are not related to the order remain unchanged.

**Scope**
- Update Marginfi Program
	- State
	- Instructions(new)
	- Instructions(old) for compatibility e.g repay, withdraw
- Update CLI
- Update Test Utils
- Update Tests

**Notes**
Support for the Kamino and other integrations are not added in this PR, it requires very few changes along with tests and can be added in a separate PR.


programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,640 @@
...
+            marginfi_account_authority: marginfi_account.authority,
+            marginfi_group: marginfi_account.group,
+        },
+        order: order_loader.key(),
+    });
+
+    Ok(())
+}
+
+pub fn keeper_close_order(ctx: Context<KeeperCloseOrder>) -> MarginfiResult {
```
abelmarnk, line 193: We might want to relax this so it works even after the marginfi account has been closed.

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,662 @@
...
+pub struct PlaceOrder<'info> {
+    #[account(
+        constraint = (!group.load()?.is_protocol_paused()) @ MarginfiError::ProtocolPaused
+    )]
+    pub group: AccountLoader<'info, MarginfiGroup>,
+
+    #[account(
+        mut,
+        has_one = group @ MarginfiError::InvalidGroup,
+        has_one = authority @ MarginfiError::Unauthorized
```
jgur-psyops, line 549: In 1.7 (merging this week) we added a "frozen" account state, frozen accounts should be unable to place or execute orders.

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,662 @@
...
+        MarginfiError::OrderLiabilityNotClosed
+    );
+
+    health_cache.set_healthy(true); // We have checked the account to be healthy
+    marginfi_account.health_cache = health_cache;
+
+    let net = order_assets_in_equity;
+
+    // Check that the liquidator did not over-withdraw.
+    match order.trigger {
```
jgur-psyops, line 454: I believe only checking net here means the keeper can keep the "excess profit" here, in take-profit instances, correct?

Example:
* User has $150 asset, $50 liability, sets a take-profit at $90
* Net is currently above $90 so the order is triggered

### Friendly Keeper:
Repays $50 in debt, withdraws $50. User is left with $100 in asset, no liabilities

### Profit-seeking Keeper:
Repays $50 in debt, withdraws $60. User is left with $90 in asset, no liabilities, keeper makes $10

Here perhaps we may want to enforce a maximum profit (my gut says a flat amount in dollars) using the health difference before/after.

I don't think there is a equivalent on the stop-loss side, only on the take-profit/both path.

abelmarnk: That makes sense, my previous intention was to make it similar to order matching.
For this we would enforce the change based on:- 
- The health difference like you said(i.e the account does not become less healthy than it was)
- And also capping the profit in dollars that the keeper would get to keep, even if the account were left in a state as healthy as it was before the order.

What would be a reasonable cap for the profit, since orders may be frequent i'm guessing it shouldn't be high like maybe something from $0.125 to $0.75?

An example based on everything so far using $0.25 as a cap:-

***Asset A (Equity = $10, Maintenance = $9.5)***

***Liab B (Equity = $5, Maintenance = $6.5)***

***TP = $3***

The user's expected net would be $3, the changes would then lead to the remaining assets left being worth $4.75, because of the cap.

programs/marginfi/src/instructions/marginfi_account/order.rs
abelmarnk: Noted.

programs/marginfi/src/instructions/marginfi_account/order.rs
jgur-psyops: We might not even need to enforce "healthy" just "at least equal or healthier than at start", as we would prefer an unhealthy account to execute an order that made them healthier vs being liquidated. 

The max profit can be configurable in the same as `liquidation_max_fee`. This would probably some flat amount in $.

But also, because of slippage constraints, when swapping from withdrawn/repaid asset, even the "Friendly Keeper" isn't going to bring your net to exactly your intended level, e.g. when he swaps the $50 he withdrew from you he'll probably get $49.9 to repay, to repay your whole $50 he'd need to withdraw maybe $50.1, so there probably needs to be a bps component. As slippage varies per-asset the slippage tolerance component might be something the user specifies on the `Order` when they make it.

abelmarnk: Sounds good!

abelmarnk: Made some slight changes to the approach, it may not be in all cases that the account is as healthy as it was or better after execution, while also keeping it profitable for the keeper, e.g in the case where the maintenance weights for both the assets and the liabilities are 1, then such a reduction cannot happen(while still allowing the keeper to keep a bit for fees or as allowed by slippage), though the only case where it can't happen is when the weights for assets and liabilities are equal to 1(Also the changes the changes are still guided by the the slippage the user set as well as the max fee, i.e it cannot go lower than what the slippage allows).
wdyt?

jgur-psyops:
Hey @abelmarnk , we're going to move this internally to complete it next week, if you have any other changes you'd like to commit please push them by Monday Feb 02 2026. Because this is a larger change, it will move an internal branch, we'll edit for a bit, merge into the consolidated update, and then it will ship to prod with 1.8 in ~late Feb.

PR #471, opened 2026-01-08
Drift deposit limit native

IMEF-FEMI:
Summary: Normalize Drift Bank Deposit Limit Configuration

The updates ensures Drift bank deposit limits can be configured using the same pattern as Kamino, Solend, and regular marginfi banks—in native token decimals. Previously, Drift banks required admins to manually convert deposit limits to 9-decimal scaled balance format, creating an inconsistent configuration experience.

  Goal: All marginfi bank integrations should have a consistent admin experience:

  | Integration  | Before This Change          | After This Change |
  |--------------|-----------------------------|-------------------|
  | Regular Bank | Native decimals ✓           | Native decimals ✓ |
  | Kamino       | Native decimals ✓           | Native decimals ✓ |
  | Solend       | Native decimals ✓           | Native decimals ✓ |
  | Drift        | 9-decimal scaled balance ❌ | Native decimals ✓ |

  Drift was the outlier because it internally uses 9-decimal precision for all scaled balances, and marginfi stores these directly with mint_decimals = 9. This required admins to "think differently" when configuring Drift banks.

  Potential Admin Mistake Without This Change

  If an admin configured a Drift USDC bank without realizing the 9-decimal requirement:

  // Admin intends: 10 million USDC limit
  // Admin sets (thinking in 6-decimal like other USDC banks):
  deposit_limit = 10_000_000_000_000  // 10M × 10^6

  // What actually happens:
  // Bank interprets as 10^13 scaled balance units
  // = 10,000 "whole units" in 9-decimal
  // = 10,000 USDC effective limit

  // Result: 1000x LOWER than intended

  Consequences:
  - Users would hit deposit limits almost immediately
  - Difficult to diagnose since the misconfiguration wouldn't cause errors at setup time—only when users tried to deposit

  After This Change

  Admins configure all bank types identically:
  // For ANY USDC bank (regular, Kamino, Solend, or Drift):
  deposit_limit_native = 10_000_000_000_000  // 10M USDC in 6-decimal

  // The program handles internal conversion automatically


programs/marginfi/src/state/drift.rs
```diff
@@ -16,7 +18,11 @@ pub struct DriftConfigCompact {
     pub oracle: Pubkey,
     pub asset_weight_init: WrappedI80F48,
     pub asset_weight_maint: WrappedI80F48,
-    pub deposit_limit: u64,
+    /// Deposit limit in **native token decimals** (e.g., for USDC with 6 decimals,
+    /// 1_000_000_000_000 = 1M USDC). This is automatically converted to 9-decimal
+    /// scaled balance units when creating the bank, since Drift uses fixed 9-decimal
+    /// precision for all spot market balances.
+    pub deposit_limit_native: u64,
```
jgur-psyops: we can probably keep the original name since it now functions like all the other deposit limits do

programs/marginfi/src/state/drift.rs
```diff
@@ -62,8 +68,41 @@ impl DriftConfigCompact {
...
+        &self,
+        spot_market_key: Pubkey,
+        token_decimals: u8,
+    ) -> MarginfiResult<BankConfig> {
+        require!(
+            token_decimals <= DRIFT_PRECISION_EXP as u8,
+            MarginfiError::MathError
+        );
+
+        let deposit_limit = if token_decimals < 9 {
```
jgur-psyops: const instead of 9

programs/marginfi/src/state/drift.rs
```diff
@@ -62,8 +68,41 @@ impl DriftConfigCompact {
...
+            let decimal_adjustment = 10u64.pow((9 - token_decimals) as u32);
```
jgur-psyops: use our static pow array instead of computing pow

jgur-psyops: we have an adjust_decimals function somewhere already (I don't recall where offhand, maybe kamino utils?)

tests/d06_driftBankInit.spec.ts
```diff
@@ -143,10 +143,14 @@ describe("d06: Init Drift banks", () => {
...
-    assertBNEqual(
-      config.depositLimit,
-      new BN(100_000_000 * 10 ** ecosystem.usdcDecimals)
-    ); // 100mil USDC because big interest needed
+    // deposit_limit is stored in 9-decimal scaled balance units (converted from native decimals)
+    // Input: 100M USDC in native decimals = 100_000_000 * 10^usdcDecimals
+    // Stored: input * 10^(9 - usdcDecimals) = 100M in 9-decimal
+    const inputDepositLimit = new BN(100_000_000).mul(new BN(10 ** ecosystem.usdcDecimals));
+    const scaledDepositLimit = inputDepositLimit.mul(
+      new BN(10 ** (DRIFT_SCALED_BALANCE_DECIMALS - ecosystem.usdcDecimals))
```
IliaZyrin: Use `USDC_SCALING_FACTOR`

programs/marginfi/src/state/drift.rs
IliaZyrin: Afaik we have it in TS tests, but not in Rust. Can be added though. [Here](https://github.com/mrgnlabs/marginfi-v2/blob/drift-solend-integration/programs/drift-mocks/src/constants.rs#L62) is the best place probably, since there is already a similar conversion function.

programs/marginfi/src/state/drift.rs
IliaZyrin: Agreed with this.

Henry-E:
It's kind of a big change to go from tracking scaled balance exactly, which is what we do currently on all the integrations, to tracking something different from what's actually on the drift account. If you guys want to make this change go ahead but make sure you get the maths fully correct.

programs/drift-mocks/src/constants.rs
```diff
@@ -65,3 +66,30 @@ pub fn get_precision_increase(token_decimals: u32) -> Result<u128> {
...
+pub fn scale_drift_deposit_limit(deposit_limit: u64, mint_decimals: u8) -> Result<I80F48> {
+    let limit = I80F48::from_num(deposit_limit);
+
+    if mint_decimals == DRIFT_SCALED_BALANCE_DECIMALS {
+        return Ok(limit);
+    }
+
+    if mint_decimals < DRIFT_SCALED_BALANCE_DECIMALS {
+        let diff = (DRIFT_SCALED_BALANCE_DECIMALS - mint_decimals) as usize;
+        let scale = *EXP_10
```
IliaZyrin: Use `EXP_10_I80F48` instead.

programs/drift-mocks/src/constants.rs
```diff
@@ -65,3 +66,30 @@ pub fn get_precision_increase(token_decimals: u32) -> Result<u128> {
...
+            .get(diff)
+            .ok_or_else(|| error!(DriftMocksError::MathError))?;
+        return limit
+            .checked_mul(I80F48::from_num(scale as i128))
+            .ok_or_else(|| error!(DriftMocksError::MathError));
+    }
+
+    let diff = (mint_decimals - DRIFT_SCALED_BALANCE_DECIMALS) as usize;
+    let scale = *EXP_10
```
IliaZyrin: `EXP_10_I80F48`

Anchor.toml
```diff
@@ -38,7 +38,7 @@ wallet = "~/.config/solana/id.json"
 
 # (remove RUST_LOG= to see bankRun logs)
 [scripts]
-test = "RUST_LOG= yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/*.spec.ts --exit --require tests/rootHooks.ts"
+# test = "RUST_LOG= yarn run ts-mocha -p ./tsconfig.json -t 1000000 tests/*.spec.ts --exit --require tests/rootHooks.ts"
```
IliaZyrin: Please revert changes to this file.

programs/marginfi/src/instructions/drift/add_pool.rs
```diff
@@ -51,20 +49,20 @@ pub fn lending_pool_add_bank_drift(
...
     let fee_vault_authority_bump = ctx.bumps.fee_vault_authority;
 
+    let config = bank_config.to_bank_config(spot_market_key);
+
     *bank = Bank::new(
         ctx.accounts.group.key(),
         config, // Use the modified BankConfig directly instead of converting from BankConfigCompact
         bank_mint.key(),
-        DRIFT_SCALED_BALANCE_DECIMALS, // Use fixed 9 decimals for all Drift banks
+        bank_mint.decimals,
```
jgur-psyops, line 70: nit: let's sanity check this is less than 10^19 (or whatever the max is for drift before decimal adjustment gets wacky)

PR #472, opened 2026-01-09
JupLend integration

Henry-E:
## Summary

Add JupLend integration.

### Deployed to Alt-Staging (mainnet-beta)
- **Program**: `5UDghkpgW1HfYSrmEj2iAApHShqU44H6PKTAar9LL9bY`
- **Group**: `ERBiJdWtnVBBd4gFm7YVHT3a776x5NbGbJBR5BDvsxtj`
- **Banks**: WSOL, USDC, USDT, USDG, USDS (seed=600, asset_tag=6)


Henry-E:
Things to be aware of
- [ ] Juplend's public source doesn't match what's on chain, specifically related to about 3 different accounts, two of which are supposed to be ATAs on chain but in their source that's not the case. And another which is supposed to be optional (and which we would not pass in) but is currently non-optional on chain.
- [ ] There's an ops/ folder that can be migrated to the mrgn-ts-scripts repo if needs be, it's just here for now for simplicity
- [ ] Similarly the branch is forked from the `remove-non-bankrun-tests` PR, which allows tests to be run when developing in <8 seconds vs 1m+ previously which greatly helped with development speed.  There's a `patch` file in there which could be applied to different branch to recreate the same changes. Only including it just so it's known that it's a possibility.

PR #474, opened 2026-01-14
Final tweaks on Drift implementation

IliaZyrin:
Also contains oracles-related fixes from https://github.com/mrgnlabs/marginfi-v2/pull/438 and https://github.com/mrgnlabs/marginfi-v2/pull/467


type-crate/src/constants.rs
```diff
@@ -29,7 +29,8 @@ pub const DAILY_RESET_INTERVAL: i64 = 24 * 60 * 60; // 24 hours
 
 /// Due to real-world constraints, oracles using an age less than this value are typically too
 /// unreliable, and we want to restrict pools from picking an oracle that is effectively unusable
-pub const ORACLE_MIN_AGE: u16 = 30;
+/// Switchboard oracles are cranked on demand, so we can use a lower value (10 seconds)
+pub const ORACLE_MIN_AGE: u16 = 10;
 pub const MAX_PYTH_ORACLE_AGE: u64 = 60;
 pub const MAX_SWB_ORACLE_AGE: u64 = 3 * 60;
```
jgur-psyops: this one can be removed now, or ORACLE_MIN_AGE becomes this one, whatever makes sense to you.  I agree 10s is probably fine for swb.

programs/drift-mocks/src/state.rs
```diff
@@ -54,35 +63,41 @@ pub struct MinimalSpotMarket {
...
 
-    pub padding_2b: [u8; 104],
+    pub _padding4: [[u8; 32]; 3],
+    pub _padding5: [u8; 8],
 
     pub decimals: u32,
     pub market_index: u16,
 
-    pub padding_3: [u8; 49],
+    pub _padding6: [[u8; 16]; 3],
```
jgur-psyops: nit: This is an amusing way to reach 49 bytes! I usually do:

```
  pub _padding6: [u64; 6]; 
  pub _padding7: [u8; 1]
```
or
```
  pub _padding6: [u8; 32]; 
  pub _padding7: [u8; 16];
  pub _padding8: [u8; 1]
```

What you have here is fine obviously, just pointing out that it's a curious thing.

programs/drift-mocks/src/state.rs
```diff
@@ -361,49 +390,41 @@ impl MinimalUser {
     pub fn validate_spot_position(&self, market_index: u16) -> Result<()> {
         let expected_position_index = if market_index == 0 { 0 } else { 1 };
```
jgur-psyops, line 409: Probably should add a comment here because it's quite confusing. If I recall correctly, the logic is that index = 0 means it's the quote token market (USDC), and it will be at index 0, otherwise it's always at index 1? Something to that effect.

jgur-psyops:
LGTM

programs/drift-mocks/src/state.rs
IliaZyrin: Yeah, I initially did it with the "powers-of-two" way everywhere but then for this specific type I needed `Default` trait and it required < 40 bytes arrays because the bigger ones do not define `Default` trait. So this type is done in this weird way but the rest is more elegant!

Good point about using u64 though, somehow forgot about it.

programs/drift-mocks/src/state.rs
IliaZyrin: But it's already in the comment above:
```
/// - USDC (market_index 0) must use position[0]
/// - All other assets must use position[1]
```

PR #476, opened 2026-01-15
fix: added emode leverage limit validation to the lending_pool_config…

IMEF-FEMI:
Added Emode entries validation to the lending_pool_configure_bank_emode instruction 
And added test for it


IliaZyrin (approved):
LGTM

PR #486, opened 2026-01-21
Human-readable rate points

### LINKED ISSUE #482: ratepoints human-readable in add/configure bank
Currently rate points output in their raw u32 form, which is not human-readable without conversion. Display the f32 or f64 "actual" rate for easier tx preview.

chambaz:
Add u32_to_milli and u32_to_centi conversion functions to type-crate
as inverses of the existing milli_to_u32 and centi_to_u32 functions.
Update log_pool_info to display interest rate curve points as readable
percentages instead of raw u32 values.

Closes #482

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
@@ -1,6 +1,6 @@
 use crate::utils::wrapped_i80f48_to_f64;
 use anchor_lang::prelude::*;
-use marginfi_type_crate::types::Bank;
+use marginfi_type_crate::types::{u32_to_centi, u32_to_milli, Bank};
 
 /// Echo the information used to create banks to the log output. Useful for at-a-glance debugging
 /// bank creation txes in explorer. Note: costs a lot of CU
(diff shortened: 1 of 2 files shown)


programs/marginfi/src/instructions/marginfi_group/add_pool_common.rs
```diff
@@ -47,9 +47,18 @@ pub fn log_pool_info(bank: &Bank) {
...
+        "Rate at 0: {:.2}% rate at 100: {:.2}%",
+        u32_to_milli(interest.zero_util_rate).to_num::<f64>() * 100.0,
+        u32_to_milli(interest.hundred_util_rate).to_num::<f64>() * 100.0,
     );
+    for (i, p) in interest.points.iter().enumerate() {
+        if p.util != 0 {
+            msg!(
+                "  Point {}: util={:.2}% rate={:.2}%",
+                i,
+                u32_to_centi(p.util).to_num::<f64>() * 100.0,
```
jgur-psyops: nit: float math eats a lot of compute, might as well omit the *100.0 and just print without mul.

Issue #485, opened 2026-01-21
Daily withdraw/borrow limit


jgur-psyops:
We want to impose a reasonable per-bank limit on daily withdraws/borrows. It suffices to impose a 24-hour limit that reset to zero at some fixed time each day, though a sliding-24-hour window would be cooler.

* Denominated in native token
* Permission-less reset (admin doesn't have to do anything to reset the window)
* Should exclude flashloan, liquidation, and deleverage traffic - all other kinda of financial txes increment/decrement the window and will fail when it's exceeded until it resets
* Deposits offset withdraws, same with repays/borrows, so spamming deposit/withdraw doesn't consume any of the window

Nice to haves:
* global (per group) windows on a dollar basis, e.g. $10M in withdraws/day
* Hourly and Daily windows, with different limits

jgur-psyops:
Closed in #489

Issue #488, opened 2026-01-22
Risk engine account packing normalization


jgur-psyops:
Reminder to contributors: Please base off the latest version branch, and target the consolidated update with your PR, not main. The main branch reflects the latest version on mainnet, and may be 1-2 months stale compared to what's in active developement. Right now, that's Right now, that's 0.1.8-main, and https://github.com/mrgnlabs/marginfi-v2/pull/493

# Summary 

Partially blocks #485 until completed

In order to support borrow/withdraw windows, we must force the caller to pass risk accounts for all balances, including the one being withdraw/repaid in withdraw-all and repay-all paths. Currently, we let the caller omit those.

jgur-psyops:
Closed in #489

PR #489, opened 2026-01-27
rate limiter feature and risk account enforcement for repay all and withdraw all. issues: #485 and #488

IMEF-FEMI:
Summary of changes
                                                                                                                                                                                
  - Implements a sliding window rate limiter system at both bank and group levels to protect against rapid fund extraction                                                      
  - Enforces risk account requirements for all balances in withdraw-all and repay-all paths                                                                                     
  - Resolves issues #485 and #488                                                                                                                                               
                                                                                                                                                                                
  Features                                                                                                                                                                      
                                                                                                                                                                                
  1. Sliding Window Rate Limiter                                                                                                                                                
                                                                                                                                                                                
  A new rate limiting system that tracks net outflow (withdraws + borrows - deposits - repays) with configurable hourly and daily windows:                                      
                                                                                                                                                                                
  Bank-Level Rate Limiting (native tokens)                                                                                                                                      
  - Configurable hourly and daily max outflow limits per bank                                                                                                                   
  - Tracks net outflow in native token amounts                                                                                                                                  
  - Deposits and repays offset withdraws and borrows                                                                                                                            
                                                                                                                                                                                
  Group-Level Rate Limiting (USD)                                                                                                                                               
  - Configurable hourly and daily max outflow limits for the entire group                                                                                                       
  - Tracks aggregate net outflow in USD using oracle prices                                                                                                                     
  - Provides protocol-wide protection against coordinated extraction                                                                                                            
                                                                                                                                                                                
  Key Implementation Details:                                                                                                                                                   
  - Uses weighted blend of previous and current windows for smooth capacity transitions                                                                                         
  - Automatically skipped during flashloan, liquidation, and deleverage operations                                                                                              
  - Inflows (deposits/repays) reduce net outflow, releasing capacity for subsequent outflows                                                                                    
  - Setting limit to 0 disables rate limiting for that window                                                                                                                   
                                                                                                                                                                                
  2. Risk Account Enforcement for Close-Balance Operations                                                                                                                      
                                                                                                                                                                                
  Forces callers to pass risk accounts for ALL balances in withdraw-all and repay-all paths

### DIFF
--- a/programs/marginfi/fuzz/src/lib.rs
+++ b/programs/marginfi/fuzz/src/lib.rs
@@ -629,6 +629,7 @@ impl<'state> MarginfiFuzzContext<'state> {
                     MarginfiError::RiskEngineInitRejected.into(),
                     MarginfiError::NoAssetFound.into(),
                     MarginfiError::BankAccountNotFound.into(),
+                    MarginfiError::InvalidBankAccount.into(),
                     MarginfiError::AccountDisabled.into(),
                 ]
                 .contains(&error),


programs/marginfi/src/instructions/drift/withdraw.rs
```diff
@@ -158,6 +173,30 @@ pub fn drift_withdraw<'info>(
             (token_amount, scaled_decrement)
         };
 
+        // Rate limiting tracks net outflow; skip for flashloan/liquidation/deleverage flows.
+        let rate_limit_amount = if withdraw_all { token_amount } else { amount };
```
IliaZyrin: You should always just use `token_amount` here since it can potentially differ from `amount` even if [withdraw_all is false](https://github.com/mrgnlabs/marginfi-v2/pull/489/changes#diff-99bb30dfa38177622ee870b8c9eaebdf431f37703e4f09a90ea515e91723a6bbR167).

programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -129,6 +145,35 @@ pub fn kamino_withdraw<'info>(
...
+                    .try_record_outflow(rate_limit_amount, clock.unix_timestamp)?;
+            }
+
+            // Group-level rate limiting (USD) - use fresh oracle price
+            if group_rate_limit_enabled {
+                check!(price > I80F48::ZERO, MarginfiError::InvalidRateLimitPrice);
+                let usd_value = calc_value(
+                    I80F48::from_num(rate_limit_amount),
+                    price,
+                    bank.get_balance_decimals(),
```
IliaZyrin: Nitpick but I would suggest to use bank.mint_decimals to explicitly state that in this scenario we always use actual mint_decimals (even in case of Drift), so that this wrapper function is only used where necessary (i.e. where it conveys the meaning of "the similar logic in other venues does give different results").


```suggestion
                    bank.mint_decimals,
```

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -139,6 +157,30 @@ pub fn lending_account_withdraw<'info>(
...
+                    .try_record_outflow(rate_limit_amount, clock.unix_timestamp)?;
+            }
+
+            // Group-level rate limiting (USD) - use fresh oracle price (fetched above)
+            if group_rate_limit_enabled {
+                check!(price > I80F48::ZERO, MarginfiError::InvalidRateLimitPrice);
+                let usd_value = calc_value(
+                    I80F48::from_num(rate_limit_amount),
+                    price,
+                    bank.get_balance_decimals(),
```
IliaZyrin: Same as with [Kamino Withdraw](https://github.com/mrgnlabs/marginfi-v2/pull/489/changes#r2738732534):

```suggestion
                    bank.mint_decimals,
```

programs/marginfi/src/instructions/solend/withdraw.rs
```diff
@@ -121,6 +135,34 @@ pub fn solend_withdraw<'info>(
...
+                    .try_record_outflow(rate_limit_amount, clock.unix_timestamp)?;
+            }
+
+            // Group-level rate limiting (USD) - use fresh oracle price
+            if group_rate_limit_enabled {
+                check!(price > I80F48::ZERO, MarginfiError::InvalidRateLimitPrice);
+                let usd_value = calc_value(
+                    I80F48::from_num(rate_limit_amount),
+                    price,
+                    bank.get_balance_decimals(),
```
IliaZyrin: Same as above:

```suggestion
                    bank.mint_decimals,
```

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -33,6 +33,142 @@ pub fn get_remaining_accounts_per_bank(bank: &Bank) -> MarginfiResult<usize> {
     }
 }
 
+/// Validate remaining accounts for active balances, optionally skipping a specific bank.
+///
+/// When `skip_bank` is provided, the balance for that bank is excluded from validation.
+/// This is used when closing a balance via withdraw_all - the remaining accounts only need
+/// to cover the balances that will remain active after the close.
+pub fn validate_remaining_accounts_for_balances<'info>(
```
IliaZyrin: This is unused, let's remove.

programs/marginfi/src/state/rate_limiter.rs
```diff
@@ -0,0 +1,348 @@
...
+
+        // Total net outflow = weighted_prev + cur_window_outflow
+        let total_net_outflow = weighted_prev.saturating_add(self.cur_window_outflow);
+
+        // Remaining capacity = max_outflow - total_net_outflow
+        // If total_net_outflow is negative (more inflows), we have extra capacity
+        (self.max_outflow as i64).saturating_sub(total_net_outflow)
+    }
+
+    fn record_flow(&mut self, amount: i64, current_timestamp: i64) -> MarginfiResult<()> {
```
IliaZyrin: This function seems redundant. I guess it was planned to be used for both inflows and outflows but currently is only used by the outflow one.

So I suggest just moving the logic directly into `try_record_outflow()` and simplify it accordingly.

programs/marginfi/src/state/rate_limiter.rs
```diff
@@ -0,0 +1,348 @@
...
+        self.hourly
+            .initialize(max_outflow, HOURLY_WINDOW_DURATION, current_timestamp);
+    }
+
+    fn configure_daily(&mut self, max_outflow: u64, current_timestamp: i64) {
+        self.daily
+            .initialize(max_outflow, DAILY_WINDOW_DURATION, current_timestamp);
+    }
+
+    fn try_record_outflow(&mut self, amount: u64, current_timestamp: i64) -> MarginfiResult<()> {
```
IliaZyrin: Most of the logic here duplicates what is in the [wrapper](https://github.com/mrgnlabs/marginfi-v2/pull/489/changes#diff-ebb7085f3dff3ba7da349650cab43de35be0370ccf6e433853645f14ee97279eR284) function. I assume it can be safely removed.

test-utils/src/marginfi_account.rs
```diff
@@ -264,7 +264,7 @@ impl MarginfiAccountFixture {
         bank: &BankFixture,
         ui_amount: T,
         withdraw_all: Option<bool>,
-        is_liquidate: bool,
+        _is_liquidate: bool,
```
IliaZyrin: Just remove it if it's not used anymore.

tests/09_withdraw_repay.spec.ts
```diff
@@ -474,34 +491,100 @@ describe("Withdraw funds", () => {
...
     const balancesAfter = userAccAfter.lendingAccount.balances;
     assert.equal(bankAfter.lendingPositionCount, 0);
 
-    // This balance is now inactive
-    assert.equal(balancesAfter[1].active, 0);
+    // The SOL balance should now be inactive (either not found or active=0)
+    const solBalanceAfter = balancesAfter.find((b) =>
+      b.bankPk.equals(bankKeypairSol.publicKey)
+    );
+    // After withdraw_all, balance is either cleared (not found) or marked inactive
```
IliaZyrin: It is always cleared -> won't be found. You could also just leave the existing check.

tests/09_withdraw_repay.spec.ts
```diff
@@ -474,34 +491,100 @@ describe("Withdraw funds", () => {
...
+    if (solBalanceAfter) {
+      assert.equal(solBalanceAfter.active, 0);
+    }
+    // If not found, that's also correct - the balance was fully closed
+  });
+
+  it("(user 1) withdraws all Token A balance - cleanup for liquidation test", async () => {
```
IliaZyrin: How was it working before without it?

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+
+const RATE_LIMIT_ACCOUNT = "rate_limit_account";
+const HOURLY_WINDOW_SECONDS = 60 * 60;
+const DAILY_WINDOW_SECONDS = 24 * 60 * 60;
+
+
+const toNative = (amount: number, decimals: number): BN =>
+  new BN(amount).mul(new BN(10).pow(new BN(decimals)));
+
+const usdcNative = (amount: number): BN =>
```
IliaZyrin: I think it would be good to move those to a common place so that we can use them in the other tests as well. Looks neat.

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+    );
+
+    await repayUsdc(usdcNative(5));
+
+    await borrowUsdc(usdcNative(5));
+
+    const group = await program.account.marginfiGroup.fetch(
+      marginfiGroup.publicKey
+    );
+    assert.ok(
```
IliaZyrin: I think we can be more precise in this check, e.g.:

```suggestion
      assertBNEqual(group.rateLimiter.hourly.curWindowOutflow, usdcNative(15));
```

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+      bankHourlyLimit
+    );
+
+    await expectFailedTxWithMessage(async () => {
+      await borrowUsdc(new BN(1));
+    }, "Bank hourly rate limit exceeded");
+  });
+
+  it("(user 2) bank hourly limit blocks excess withdraw", async () => {
+    const bankHourlyLimit = usdcNative(2);
```
IliaZyrin: Two very subjective nitpicks:
- I would not declare one-time-used constants like this, in the beginning of the test, because you have to periodically scroll up instead of just naturally following the test flow.
- Is it possible to not create a new user here? IMHO it's simpler to read the whole test suite from top to bottom if the context is persistent (but clean on startup) and the tests are small and focused.

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+    ).toNumber();
+    const maxExpectedOutflow = Math.floor(price * 10);
+    const actualOutflow =
+      groupAfter.rateLimiter.daily.curWindowOutflow.toNumber();
+    assert.ok(
+      actualOutflow > 0,
+      "Group daily outflow should be recorded after borrow"
+    );
+    assert.ok(
+      actualOutflow <= maxExpectedOutflow,
```
IliaZyrin: I think they should be just equal in this case. The price is unbiased here, right?

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+          bank: bankKeypairUsdc.publicKey,
+          tokenAccount: rateLimitUser.usdcAccount,
+          remaining: usdcOnlyRemainingAccounts(),
+          amount: withdrawAmount,
+        })
+      )
+    );
+
+    const bank = await program.account.bank.fetch(bankKeypairUsdc.publicKey);
+    const expectedNetOutflow = depositAmount.neg().add(withdrawAmount);
```
IliaZyrin: Nitpick but I think this would read easier=)

```suggestion
    const expectedNetOutflow = withdrawAmount.sub(depositAmount);
```

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+    assertBNEqual(bank.rateLimiter.hourly.curWindowOutflow, expectedNetOutflow);
+  });
+
+  it("(admin) partial updates preserve existing windows", async () => {
```
IliaZyrin: I think this can be combined with the first test since it's a sub-scenario of configuration testing.

tests/17_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+    await setRateLimits({
+      bankHourly: bankHourlyLimit,
+      bankDaily: new BN(0),
+      groupHourly: new BN(0),
+      groupDaily: new BN(0),
+    });
+
+    await borrowUsdc(bankHourlyLimit);
+
+    await setRateLimits({
```
IliaZyrin, line 514: I suggest either adding a failing borrow before this OR moving this "unlimited" logic check to some of the above tests. Otherwise it doesn't really test anything: within the context of this test there is no sign there was some "enforcement", since both borrows just passed.

tests/16_rateLimiter.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+    // Reset all rate limits to disabled state
+    await setRateLimits({
+      bankHourly: new BN(0),
+      bankDaily: new BN(0),
+      groupHourly: new BN(0),
+      groupDaily: new BN(0),
+    });
+
+    // Also reset Token A bank limits
+    await groupAdmin.mrgnProgram.provider.sendAndConfirm(
```
IliaZyrin: It seems unnecessary.

type-crate/src/constants.rs
```diff
@@ -26,6 +26,8 @@ pub const LIQUIDATION_INSURANCE_FEE: I80F48 = I80F48!(0.025);
 
 pub const SECONDS_PER_YEAR: I80F48 = I80F48!(31_536_000);
 pub const DAILY_RESET_INTERVAL: i64 = 24 * 60 * 60; // 24 hours
+pub const HOURLY_WINDOW_DURATION: u64 = 60 * 60; // 1 hour in seconds
+pub const DAILY_WINDOW_DURATION: u64 = 24 * 60 * 60; // 24 hours in seconds
```
IliaZyrin: I think you can just reuse the existing constant above (feel free to rename it to make more sense).

tests/09_withdraw_repay.spec.ts
IMEF-FEMI: It's actually not useful... i added it prior as i was having issues testing the (risk account enforcement)

programs/marginfi/src/instructions/marginfi_account/repay.rs
```diff
@@ -84,6 +97,34 @@ pub fn lending_account_repay<'info>(
...
+    if !should_skip_rate_limit(marginfi_account.account_flags) {
+        // Bank-level rate limiting (native tokens)
+        if bank.rate_limiter.is_enabled() {
+            bank.rate_limiter
+                .record_inflow(repay_amount_post_fee, clock.unix_timestamp);
+        }
+
+        // Group-level rate limiting (USD) - prefer live price, fall back to cached.
+        if group_rate_limit_enabled {
+            let rate_limit_price = fetch_rate_limit_price_for_inflow(
```
jgur-psyops: Moved this here to avoid errors from the fallback when the limit check is disabled anyways (`liquidate_receiver_repay_without_oracles_should_succeed` checks fails previously and validates this now)

programs/marginfi/src/utils/general.rs
```diff
@@ -390,6 +391,38 @@ pub fn fetch_unbiased_price_for_bank<'info>(
...
+        {
+            if price > I80F48::ZERO {
+                return Ok(price);
+            }
+        }
+    }
+
+    let cached_price: I80F48 = bank.cache.last_oracle_price.into();
+    let cached_ts = bank.cache.last_oracle_price_timestamp;
+    let max_age = bank.config.get_oracle_max_age() as i64;
```
jgur-psyops, line 403: We should tolerate more staleness here, the window is not super price sensitive, and having deposits or repays blocked by oracle cranking seems like a bad idea. I lean towards these always being non-blocking on deposit/repay, no matter how stale the cache is, but in the alternative maybe we do e.g. some number of hours or 10x oracle age. Thoughts @imef-femi @IliaZyrin @dawid-sw

IliaZyrin: Good point, I vote for "always being non-blocking on deposit/repay".

programs/marginfi/src/utils/general.rs
```diff
@@ -388,6 +389,28 @@ pub fn fetch_unbiased_price_for_bank<'info>(
...
+) -> MarginfiResult<Option<I80F48>> {
+    // Fall back to cached price if fresh enough
+    let cached_price: I80F48 = bank.cache.last_oracle_price.into();
+    let cached_ts = bank.cache.last_oracle_price_timestamp;
+    let max_age = bank.config.get_oracle_max_age() as i64;
+    let age = clock.unix_timestamp.saturating_sub(cached_ts);
+
+    if cached_price <= I80F48::ZERO || cached_ts == 0 || age < 0 || age > max_age {
+        // No valid price available - return None instead of erroring
+        return Ok(None);
```
jgur-psyops, line 408: Let me summarize the two options and see if we're making the best choice here. 

- Option 1 (now): mark a $0 if the cache is stale and do not increase the deposit window at all until untracked funds are computed later (e.g. on pulse). 
- Option 2: mark the price immediately even if the cache is arbitrarily old and increase the window.

Option 1

* Attacker loops withdraw and stale-deposits (which don't raise the window yet) until the window is exceeded, blocking legitimate withdraws. This is low risk (it's just a troll) but fairly trivial to execute (just withdraw a ton of money, wait until the cache is stale, and deposit it). You have to have a lot of money, and be dedicated to being annoying.
* In the extreme case, someone rich with actual_funds >= window_size can block the window every day until anyone sends a `pulse_bank_price_cache` (withdraw all their funds, deposit again once stale, wait until someone pulses, repeat the next day).

Option 2

* Price is $1 stale-cached, actually $100, attacker can't do anything bad in this scenario. Deposits under-count when increasing the withdraw window, which means it may be reached sooner.
* Price is $100 stale-cached, actually $1, attacker deposits $1 to increase the window by $100, enabling them to withdraw $99 more dollars unfairly elsewhere. This is bad because you can loop this to arbitrarily raise the window. In the extreme case they increase the window infinitely, but they are dependent on real price differences to execute, which is quite limiting. Here we could haircut the deposit price by ~10% to mitigate in most practical circumstances.

I lean towards the current implementation (1) because that troll is impractical and pointless whereas in (2) if the attacker can manipulate prices it opens up a lot of possibility. Just documenting this so the design choice is purposeful!

cc @IMEF-FEMI @IliaZyrin @dawid-sw @princedotrs

dawid-sw: I'm voting for 1st option as well.

Comparing to 2nd option, the 1st one is more defensive - in the worst case scenario, new inflows will not count - but it can be easily fixed later on with permissionless `pulse_bank_price_cache`. Also scenario is unlikely, someone needs to move a lot of funds, the loop is not that trivial as a withdrawal requires "fresh" price, so each loop he needs to wait until price goes stale - so it's not instant and requires some time, and can be broken by any withdrawal or pulse. So it's rather hard to perform, especially as there is no real financial gain.

princedotrs: I agree with @jgur-psyops and @dawid-sw on going with Option 1.
In this trade-off, we should prioritize Solvency over Liveness.

Option 2 Risk: If we trust stale prices, a crash + stale oracle allows attackers to bypass limits using "phantom value," draining the protocol. This is catastrophic.

Option 1 Downside: A blocking scenario is just a Liveness failure, which is permissionlessly fixable via pulse_bank_price_cache.

Fail-closed is the only safe option here.

IMEF-FEMI: i also think i'd go with the current implementation 😁
Till we think of something better in the future...

IliaZyrin: We have a consensus here - option 1!

PR #496, opened 2026-01-30
Fixed price Kamino/Drift

### LINKED ISSUE #484: Fixed prices with kamino/drift/etc
Reminder to contributors: Please base off the latest version branch, and target the consolidated update with your PR, not main. The main branch reflects the latest version on mainnet, and may be 1-2 months stale compared to what's in active developement. Right now, that's 0.1.8-main, and https://github.com/mrgnlabs/marginfi-v2/pull/493

# Summary 

While the fixed-price feature "works" for Kamino/Drift/etc, it doesn't account for interest that accumulates on those banks, which makes it impractical. We should have a fixed-price oracle type for those asset types that (1) fixes the "base" price that we normally get from Pyth/Swb/etc, (2) continues to multiply by the exchange rate for that venue's underlying Reserve/Market/etc.

jgur-psyops:
Likely needs audit in 1.8

## Summary

Closes #484 

Adds support for Fixed prices in Kamino/Drift banks. Major refactor of functions in price.rs. Adds tests for Fixed-price functionality.

### Breaking change
`lending_pool_set_fixed_oracle_price` [now expects integration "oracles"](https://github.com/mrgnlabs/marginfi-v2/pull/496/changes#diff-0b4bcb97cd3d20c81c66abe4b4928d9bdf7997119bed36d4e53ae57e1a5fa422R56) (reserve - for Kamino, spot market - for Drift) in `remaining_accounts` so that it can set the proper `OracleSetup` right away, without the need to call `lending_pool_configure_bank_oracle` afterwards. Regular banks do not require passing any remaining accounts, as before.

Thanks to https://github.com/AadityaAnand for starting the feature

### DIFF
--- a/type-crate/src/types/bank.rs
+++ b/type-crate/src/types/bank.rs
@@ -20,9 +20,12 @@ assert_struct_align!(Bank, 8);
 #[cfg_attr(not(feature = "anchor"), derive(Zeroable))]
 #[derive(Debug)]
 pub struct Bank {
+    /// The SPL token mint this bank manages
     pub mint: Pubkey,
+    /// Number of decimals of the `mint`. Must be < 24.
     pub mint_decimals: u8,
 
+    /// The `MarginfiGroup` this bank belongs to
     pub group: Pubkey,
 
     // Note: The padding is here, not after mint_decimals. Pubkey has alignment 1, so those 32
@@ -40,12 +43,18 @@ pub struct Bank {
     /// * Initially 1
     pub liability_share_value: WrappedI80F48,
 
+    /// The SPL token account holding deposited liquidity
     pub liquidity_vault: Pubkey,
+    /// PDA bump for the liquidity vault
     pub liquidity_vault_bump: u8,
+    /// PDA bump for the liquidity vault authority
     pub liquidity_vault_authority_bump: u8,
 
+    /// The SPL token account holding insurance fund tokens
     pub insurance_vault: Pubkey,
+    /// PDA bump for the insurance vault
     pub insurance_vault_bump: u8,
+    /// PDA bump for the insurance vault authority
     pub insurance_vault_authority_bump: u8,
 
     pub _pad1: [u8; 4], // 4x u8 + 4 = 8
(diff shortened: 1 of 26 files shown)


programs/marginfi/src/state/price.rs
```diff
@@ -182,29 +256,13 @@ impl OraclePriceFeedAdapter {
...
-                    // * Note: Typically price updates are owned by `pyth_solana_receiver_sdk` and the oracle
-                    // feed account itself is owned by PYTH ID. On localnet, the mock program may own both for
-                    // simplicity.
-                    let owner_ok = account_info.owner.eq(&PYTH_ID)
-                        || account_info.owner.eq(&pyth_solana_receiver_sdk::id());
-                    check!(owner_ok, MarginfiError::StakedPythPushWrongAccountOwner);
-                }
+                check_primary_oracle_key(bank_config, account_info)?;
 
+                // On localnet, allow the mock program ID OR the real one (for regression tests against
```
IliaZyrin: The comment is no longer relevant since you removed the respective logic.

type-crate/src/types/bank.rs
```diff
@@ -208,6 +210,8 @@ impl OracleSetup {
             10 => Some(Self::DriftSwitchboardPull),
             11 => Some(Self::SolendPythPull),
             12 => Some(Self::SolendSwitchboardPull),
+            14 | 15 => Some(Self::FixedKamino),
```
jgur-psyops: 15 should be FixedDrift and this should just be 14: Kamino 15: Drift rather than 14 | 15 Kamino

jgur-psyops: 13 (and Drift 14)

programs/marginfi/src/state/price.rs
```diff
@@ -574,6 +463,73 @@ impl OraclePriceFeedAdapter {
...
                 Ok(OraclePriceFeedAdapter::SwitchboardPull(price_feed))
             }
+            OracleSetup::FixedDrift => {
+                // Fixed base price + Drift spot market exchange rate
+                // Requires: Drift spot market (no oracle needed)
+                check!(ais.len() == 1, MarginfiError::WrongNumberOfOracleAccounts);
+
+                let spot_market_info = &ais[0];
+
+                let spot_market_loader = load_drift_spot_market(bank_config, spot_market_info)?;
```
jgur-psyops, line 473: Just to document this for future reference: we actually don't have to check Reserve/SpotMarket for Drift or Kamino banks because they enforce reserve/market validation when they are first created, they cannot be created with a FixedPrice, and the reserve/market they are created with can never change.

Nonetheless, a sanity check is always welcome, in case any of the above ever ceases to be accurate!

PR #498, opened 2026-01-31
Receivership CU reduction via pricing cache

jgur-psyops:
Scheduled for audit in 1.8

## Summary

Receivership liquidation exceeds its CU budget with a large number of integration accounts (mostly Kamino) primarily because it must load price data twice: once at Start, once at End. However, since they are in the same tx, the price inherently does not change between Start/End, so this is quite wasteful.

This implementation caches the prices in Start Liquidation and consumes them in End, avoiding this wasteful reload, which enables the more extreme scenarios in test m03 to pass with sufficient CU.

Another notable finding in this development has been that using spot/twap pricing and loading the confidence interval currently accounts for a considerable portion of the risk engine budget, switching to spot only with no confidence adjustments would save at least 30% of the current CU budget used in loading the risk engine (at least when all involved accounts are Pyth-based).

## Breaking Changes
Start Liquidation - Banks passed in remaining accounts must now be marked writeable
End Liquidation - Now only requires Banks to be passed in remaining accounts, omit all oracles/reserves/etc. Banks must be marked writeable.

Note: dupes a ton of liquidator code. See trait-based-risk-engine for alternative attempt to reconcile: it's not pretty :(


tests/utils/user-instructions.ts
```diff
@@ -659,6 +643,79 @@ export const composeRemainingAccounts = (
...
+  });
+
+  return banksAndOracles.map((accs) => ({
+    pubkey: accs[0],
+    isSigner: false,
+    isWritable: true,
+  }));
+};
+
+const toWritableAccountMetas = (
```
IliaZyrin, line 702: I'm not getting why we need this one.

PR #499, opened 2026-02-04
Limit orders feature

jgur-psyops:
# Summary

Introduces "Orders". An `Order` is a stop-loss and/or take-profit trigger that a "Keeper"  can permissionlessly execute. For example, a user is lending $100 in SOL and borrowing $50 in BONK. They set a take-profit at $70. If SOL goes to $120, the Keeper can execute, closing their BONK position and leaving them with $70 in SOL. A Keeper can also execute if BONK falls to $30, likewise leaving the user with $70 in SOL. 

Project Zero will run Keepers initially upon feature public launch (ETA Q2/Q3 2026), but any third-party can run a Keeper. Users configure their max slippage tolerance when setting the order, Keepers are permitted to keep whatever is leftover after completing the order execution as profit, and they also get to keep the rent from the Order (currently worth about $0.25). If you are an integrator expecting to use this feature, you are strongly recommended to run your own Keepers. Keepers are permissionless, any wallet can be a Keeper.

## New Accounts

* `Order` - tracks information about a single take-profit and/or stop-loss order for an asset/liability pair on the user's account.
* `ExecuteOrderRecord` - an ephemeral account that is always closed in the same TX it is opened in, used to pass information between the start and end of order execution. None of these should exist in production. 

## New Instructions 
* `PlaceOrder` (user) - Place a new Stop Loss, Take Profit, or Both type Order on a pair of balances the user currently holds.
* `StartExecuteOrder` (Keeper) - Keepers run this to begin the execution of an Order. Must be at the start of the tx, and `EndExecuteOrder` must appear last in the tx. Withdraw/Repay of the involved balances typically follows this ix. Requires a risk check of just the balances involved in the Order.
*  `EndExecuteOrder` (Keeper) - Must be the last tx in executing an Order. Requires a risk check of just the balances involved in the Order.
* `CloseOrder` (user) - Clear an unwanted Order, user gets their rent back. 
* `SetKeeperCloseFlags` (user) -   Enables the Keeper to close Orders via `KeeperCloserOrder`, typically use `CloseOrder` instead. 
* `KeeperCloserOrder` (Keeper) - Close an Order on an account where neither of the original positions exists or all the tags have been cleared by the user

## Notes
* Although orders apply to an asset and liability pair, the user can also have other positions on their account. For example, if a user has lending positions A, B, C, and borrowing positions D, E, F, the user might have orders on A/D, A/E, and C/F at the same time. 
* The borrow-side position is always closed in full. The lending position is never closed. It can be withdrawn down to $0, but must remain open. If the Balance is closed by the user (e.g. by withdraw_all), and the same asset is deposited later to re-open it, Orders created prior to the Balance being closed **will not work**. This means users are able to modify their accounts such that active Orders are orphaned and can no longer execute, it's up to users to make sure they do not close out positions involved with their Orders without updating the Orders too.
* If a user sets a take-profit on A/B and then separately sets a stop loss on A/E, then the A/B order executes such that they no longer have enough of asset A to fulfill the A/E stop loss, then the A/E stop loss will remain open but can't be executed. It's up to users to make sure their various orders do not interfere. This is consistent with e.g. most perps platforms, where executing a stop-loss or take-profit does not close the other open order.

## Other Bug Fixes
* Fixes an issue where the `test-program-remix` script corrupts metadata and requires an `anchor clean` when the program is updated, however the tests now build to a different target directory, which increases disk usage.
* Partially fixes a `BlockhashNotFound` some longer Rust tests occasionally run into. If you're still seeing the error, add a `refresh_blockhash()` just prior, and make sure the tx is using `get_latest_blockhash()` instead of `ctx.last_blockhash`

Continues #470 

Thank you to https://github.com/abelmarnk for starting the feature.

### DIFF
--- a/programs/marginfi/src/lib.rs
+++ b/programs/marginfi/src/lib.rs
@@ -32,13 +32,13 @@ pub mod marginfi {
     /// provided on every call. Emode leverage caps reset to defaults (15x/20x) if None is passed.
     pub fn marginfi_group_configure(
         ctx: Context<MarginfiGroupConfigure>,
-        new_admin: Pubkey,
-        new_emode_admin: Pubkey,
-        new_curve_admin: Pubkey,
-        new_limit_admin: Pubkey,
-        new_emissions_admin: Pubkey,
-        new_metadata_admin: Pubkey,
-        new_risk_admin: Pubkey,
+        new_admin: Option<Pubkey>,
+        new_emode_admin: Option<Pubkey>,
+        new_curve_admin: Option<Pubkey>,
+        new_limit_admin: Option<Pubkey>,
+        new_emissions_admin: Option<Pubkey>,
+        new_metadata_admin: Option<Pubkey>,
+        new_risk_admin: Option<Pubkey>,
         emode_max_init_leverage: Option<WrappedI80F48>,
         emode_max_maint_leverage: Option<WrappedI80F48>,
     ) -> MarginfiResult {
--- a/guides/ADMIN/DEPLOY_GUIDE.md
+++ b/guides/ADMIN/DEPLOY_GUIDE.md
@@ -2,48 +2,55 @@
 
 Marginfi program authority is managed by squads (https://app.squads.so/squads/J3oBkTkDXU3TcAggJEa3YeBZE5om5yNAdTtLVNXFD47/home) and uses verified builds.
 
-
 First you will need:
-* Solana tools 2.1.20 or later (`agave-install init 2.1.20`)
-* solana-verify (`cargo install solana-verify`)
-* Docker (https://docs.docker.com/engine/install/ubuntu/)
-* A wallet with at least 15 SOL (this guide will assume your wallet is at `~/keys/mainnet-deploy.json`). Verify the pubkey of your wallet with `solana-keygen pubkey ~/keys/mainnet-deploy.json` and verify you have at least 15 SOL with `solana balance -k ~/keys/mainnet-deploy.json`
-* An RPC provider connected to mainnet (`solana config set --url https://api.mainnet-beta.solana.com`). The Solana public api is usually not sufficient, and a custom rpc is suggested.
 
+- Solana tools 2.1.20 or later (`agave-install init 2.1.20`)
+- solana-verify (`cargo install solana-verify`)
+- Docker (https://docs.docker.com/engine/install/ubuntu/)
+- A wallet with at least 15 SOL (this guide will assume your wallet is at `~/keys/mainnet-deploy.json`). Verify the pubkey of your wallet with `solana-keygen pubkey ~/keys/mainnet-deploy.json` and verify you have at least 15 SOL with `solana balance -k ~/keys/mainnet-deploy.json`
+- An RPC provider connected to mainnet (`solana config set --url https://api.mainnet-beta.solana.com`). The Solana public api is usually not sufficient, and a custom rpc is suggested.
 
 Steps:
-* Make sure you are on the appropriate release tag branch and you have pulled latest.
-* Run `./scripts/build-program-verifiable.sh marginfi mainnet`. Other people signing on the multisig should also run this and validate that the hash matches. 
-* Deploy the buffer with `./scripts/deploy-buffer.sh marginfi <YOUR_RPC_ENDPOINT> ~/keys/mainnet-deploy.json`
-* Go to squads, developers, programs, pick marginfi. The buffer address is the output of the previous command. The buffer refund is the public key of the wallet you have used so far (`solana-keygen pubkey ~/keys/mainnet-deploy.json` if you don't know it). Click next.
-* Go back to your cli and paste the command Squads gave you in step 2. If this key is not the one used in your solana CLI, make sure it pass it with -k, e.g.:
+
+- Make sure you are on the appropriate release tag branch and you have pulled latest.
+- Run `./scripts/build-program-verifiable.sh marginfi mainnet`. Other people signing on the multisig should also run this and validate that the hash matches.
+- Deploy the buffer with `./scripts/deploy-buffer.sh marginfi <YOUR_RPC_ENDPOINT> ~/keys/mainnet-deploy.json`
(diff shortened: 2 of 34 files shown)


programs/marginfi/src/lib.rs
```diff
@@ -221,6 +221,40 @@ pub mod marginfi {
         marginfi_account::initialize_account_pda(ctx, account_index, third_party_id)
     }
 
+    pub fn marginfi_account_place_order(
+        ctx: Context<PlaceOrder>,
+        mint_keys: Vec<Pubkey>,
```
dawid-sw: should be bank_keys, kida important because it impacts the IDL generation.

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,785 @@
...
+    )]
+    pub group: AccountLoader<'info, MarginfiGroup>,
+
+    /// The account owning the order
+    #[account(
+        mut,
+        has_one = group @ MarginfiError::InvalidGroup,
+        constraint = {
+            let acc = marginfi_account.load()?;
+            !acc.get_flag(ACCOUNT_IN_ORDER_EXECUTION)
```
dawid-sw, line 671: shouldn't we also check for ACCOUNT_IN_RECEIVERSHIP and ACCOUNT_IN_DELEVERAGE? Its protected by `validate_instructions` - 1st transaction check, but still worth to consider as defense in-depth.

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,785 @@
...
+
+    pub system_program: Program<'info, System>,
+}
+
+#[derive(Accounts)]
+pub struct KeeperCloseOrder<'info> {
+    /// CHECK: This uses an unchecked account here so the instruction can be called even when the
+    /// marginfi account was closed.
+    /// The ownership check is checked in the handler or/and type checks are made in the handler.
+    #[account(mut)]
```
dawid-sw: Doesnt need to be mut

programs/marginfi/src/state/order.rs
```diff
@@ -0,0 +1,244 @@
...
+        let mut idx: usize = 0;
+        let mut inactive_count: u8 = 0;
+
+        for balance in marginfi_account.lending_account.balances.iter() {
+            if !balance.is_active() {
+                inactive_count += 1;
+                continue;
+            }
+
+            check!(
```
dawid-sw: Case:
- Account has all 16 balance slots active (8 lending, 8 borrowing)
- 2 balances are tagged for the order, the other 14 are not
- The 2 order balances have lower bank_pk values than all 14 non-order balances

Since balances are sorted descending by bank_pk, the non-order balances (higher pks) come first. After iterating all 14 of them, idx = 14. When the loop reaches the first order balance (lower pk, near the end).


fix - move the bounds check after the tag skip, so then we can iterate through all positions (16) and skip the 2 active, then counter check (14) should work as expected

dawid-sw (changes requested):
good implementation, only small notes, nothing serious

programs/marginfi/src/state/order.rs
jgur-psyops: Fixed and added test `execute_order_record_init_allows_order_balances_sorted_last`

programs/marginfi/src/instructions/marginfi_account/order.rs
jgur-psyops: Added these to PlaceOrder as well, and moved the checks from the handler body to the constraint

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,782 @@
...
+
+    // ORDER_ACTIVE_TAGS == 2
+    let bank_key_1 = &bank_keys[0];
+    let bank_key_2 = &bank_keys[1];
+
+    check!(bank_key_1 != bank_key_2, MarginfiError::DuplicateBalance);
+
+    let lending_account = &mut marginfi_account.lending_account;
+
+    let balance_index_1 = lending_account
```
IliaZyrin: Suggestion: let's make this a method of LendingAccount.

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,782 @@
...
+        && marginfi_account_info.data_is_empty()
+    {
+        (Pubkey::default(), Pubkey::default(), true)
+    } else {
+        // Deserialize manually using bytemuck to avoid lifetime issues
+        let data = marginfi_account_info.try_borrow_data()?;
+
+        // Check discriminator
+        require!(
+            data.len() >= 8 + std::mem::size_of::<MarginfiAccount>(),
```
IliaZyrin: Why not check for exact equality here?

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,782 @@
...
+    // For the TP case another constraint(the max fee constraint) we want to enforce is:-
+    // net >= (1 - max_fee) * (start health)
+    // It may be possible that the value (1 - max_fee) * (start health) is less than the
+    // min allowed value based on the user's slippage constraint alone, in that case
+    // we clamp it to the slippage, allowing for that much.
+    // In the case where the value was greater we use (1 - max_fee) * (start health) instead.
+
+    // Check that the liquidator did not over-withdraw.
+
+    let slippage_frac = || -> MarginfiResult<I80F48> {
```
IliaZyrin: Just replace with values?

programs/marginfi/src/instructions/marginfi_account/order.rs
```diff
@@ -0,0 +1,782 @@
...
+                .checked_mul((slippage_frac)()?)
+                .ok_or_else(math_error!())?;
+
+            let allowed_diff = (start_health)()
+                .checked_mul((max_fee_frac)()?)
+                .ok_or_else(math_error!())?;
+
+            check!(
+                ((net >= allowed_diff) && (allowed_diff >= allowed_tp))
+                    || ((net >= allowed_tp) && (allowed_tp >= allowed_diff)),
```
IliaZyrin:
```suggestion
                net >= allowed_diff && net >= allowed_tp,
```

PR #507, opened 2026-02-11
nit: minor cleanups && more descriptive Emode Leverage errors

IMEF-FEMI:
Added more descriptive errors for the emode leverage configuration. and a few minor cleanups


programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -1,11 +1,30 @@
...
-use crate::state::emode::{DEFAULT_INIT_MAX_EMODE_LEVERAGE, DEFAULT_MAINT_MAX_EMODE_LEVERAGE};
 use crate::state::marginfi_group::MarginfiGroupImpl;
 use crate::{MarginfiError, MarginfiResult};
 use anchor_lang::prelude::*;
 use fixed::types::I80F48;
 use marginfi_type_crate::types::{basis_to_u32, MarginfiGroup, WrappedI80F48};
 
+/// Validate and apply an optional emode leverage value. If `Some`, validates it is in [1, 100] and
+/// writes the basis-point encoding. If `None`, requires the current on-chain value is non-zero.
+fn apply_emode_leverage(new_value: Option<WrappedI80F48>, current: &mut u32) -> MarginfiResult {
```
jgur-psyops: nit: maybe call it validate_and_apply_emode_leverage

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -1,11 +1,30 @@
...
+    match new_value {
+        Some(wrapped) => {
+            let leverage: I80F48 = wrapped.into();
+            require!(leverage >= I80F48::ONE, MarginfiError::BadEmodeConfig);
```
jgur-psyops: if it fails one of these requirements, it should msg! on which value exceeded the limit, it was difficult when setting a bunch of configs at once to spot the one that was failing.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -1,11 +1,30 @@
...
+            require!(
+                leverage <= I80F48::from_num(100),
+                MarginfiError::BadEmodeConfig
+            );
+            *current = basis_to_u32(leverage);
+        }
+        None => {
+            require!(*current != 0, MarginfiError::BadEmodeConfig);
```
jgur-psyops: we can probably no-op here instead of erroring, or does setting a leverage of zero have some bad side effect? For example if you made a brand-new group (which defaults 0) and you don't care about emode, wouldn't this check block updating the admin?

programs/marginfi/src/state/emode.rs
```diff
@@ -112,13 +112,13 @@ impl EmodeSettingsImpl for EmodeSettings {
             let max_leverage_init = calculate_max_leverage(asset_init_w, liab_init_w)?;
             check!(
                 max_leverage_init <= max_allowed_init_leverage,
```
jgur-psyops, line 127: same here: would help if we msg! the `max_leverage_init` and `max_allowed_init_leverage` on errors so you can see by how much the limit was exceeded.

PR #508, opened 2026-02-11
Security Fix: Cap liquidation premium for small accounts

AdeshAtole:
### [HIGH] Uncapped Seized/Repaid Ratio for Small Accounts
**File:** `programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs:56`

**Issue:**
When `ignore_healthy` is true (accounts below $5 equity threshold), the fee cap check `seized <= repaid * max_fee` is completely skipped. This allows a liquidator to seize a disproportionate amount of assets relative to what they repaid.

**Attack:**
1. Find a small account (\<$5 equity) that's unhealthy
2. Repay a minimal amount of debt
3. Seize all remaining assets (no cap)
4. Profit from the difference

**Fix:** Added a relaxed 2x cap (`seized <= repaid * 2`) for small accounts. This is generous enough to allow efficient dust cleanup while preventing exploitation.

---
*Found during a Solana security audit*

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
@@ -52,10 +52,21 @@ pub fn end_liquidation<'info>(
         I80F48!(1) + LIQUIDATION_BONUS_FEE_MINIMUM,
     );
 
-    // Ensure seized asset‐value ≤ N% of repaid liability‐value, where N = 100% + the bonus fee
+    // Ensure seized asset-value ≤ N% of repaid liability-value, where N = 100% + the bonus fee.
+    // SECURITY FIX: Apply fee cap check even for small (ignore_healthy) accounts.
+    // Previously, accounts below LIQUIDATION_CLOSEOUT_DOLLAR_THRESHOLD ($5) had no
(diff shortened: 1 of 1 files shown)


jgur-psyops:
Thank you! We prefer a small dollar limit (~$5) to accommodate the operational overhead of liquidators. While we agree that the liquidator can get a disproportionate profit in cases where the user account is liquidated when worth less $5 net, it makes sense to do this because running a liquidator is very expensive (geyser/rpc, infra costs, etc), and we want to economically incentivize liquidators to continue indexing small-value accounts, which currently make up a large share of our 500,000+ accounts. 

Without a profit motive to do so, many liquidators will avoid indexing low-value accounts to save costs, which allows attackers to open many accounts to dodge liquidators.

Receivership liquidation also requires a small amount of non-recoverable rent, so the value threshold is set in anticipation of the liquidator recovering that amount.

AdeshAtole:
That makes total sense — appreciate the detailed explanation. The economics of incentivizing liquidators to index low-value accounts (especially with 500K+ accounts) is a valid reason to keep the threshold low. The rent recovery cost for receivership liquidation is a good point I hadn't considered.

Thanks for taking the time to explain the design rationale!

PR #509, opened 2026-02-12
feat: Risk-admin close for LiquidationRecord (#500)

### LINKED ISSUE #500: Risk-admin only close LiquidationRecord

# Summary

There is currently no way to recover rent from the LiquidationRecord, which is somewhat by design since it serves as a permanent record of receivership liquidations, but on a week like this one, we end up creating quite a lot of them, which really adds up!

The main blocker to closing LiquidationRecords has been coming up with a good criteria. For now, create a risk-admin only instruction to close LiquidationRecords, so we can recover rent from deleveraged or near-empty accounts. Creating new records is simply, so this is relatively low risk.

RedaRahmani:
## Summary of changes
- Added `risk_admin_close_liquidation_record` instruction to allow risk admin to close an idle `LiquidationRecord` PDA and reclaim rent.
- Enforced safety checks: risk-admin only, record must be idle (`liquidation_receiver == Pubkey::default()`), record must belong to the provided `MarginfiAccount`.
- Added pointer-consistency guard before clearing `MarginfiAccount.liquidation_record` to prevent clearing a mismatched pointer.
- Wired the instruction through the instruction module + program entrypoints.
- Added a test helper to build the close instruction in fixtures.

## Tests
- `close_liquidation_record_happy_path`: closes record, clears pointer, and proves rent reclaim via receiver balance delta (~record lamports minus fee tolerance).
- `close_liquidation_record_non_admin_fails`: non-admin fails with `Unauthorized`, record remains.
- `close_liquidation_record_active_record_fails`: active record fails with `IllegalAction`, record remains.

Closes  #500


programs/marginfi/src/instructions/marginfi_account/close_liq_record.rs
```diff
@@ -0,0 +1,73 @@
...
+    require!(
+        liq_record.marginfi_account == ctx.accounts.marginfi_account.key(),
+        MarginfiError::InvalidLiquidationRecord,
+    );
+
+    drop(liq_record);
+
+    // Only clear the pointer if it actually points to this record (or is already default).
+    // This prevents accidentally zeroing a pointer that was already reassigned.
+    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
```
jgur-psyops, line 108: Here we will likely impose a small subset of requirements on the admin (to avoid turning this into essentially a supplementary fee for liquidators). Either 
1. the marginfi account associated with this record is closed, or 
2. the record has not had activity within e.g. 90 days.

jgur-psyops: Our goal here is probably to allow the creator of the Record to close it after N (e.g. 60) days of inactivity, and allow the admin to close it after N+X (e.g. 90) days, to encourage Record creators to clean up their own rent, and allow the admin to claim if they don't bother.

RedaRahmani: Hey @jgur-psyops Thanks for the review, Your suggestion makes sense. I agree with this approach since it lets the record creator clean up first, and only falls back to admin later.

I’m going to implement it with these rules:
- After 60 days of no activity, the record creator can close the LiquidationRecord and reclaim the rent (SOL).
- After 90 days of no activity, the risk admin can close it only if the creator hasn’t already closed it.

I’ll proceed with this direction and keep the change minimal and consistent with existing patterns. If you spot any other changes you think are necessary, please let me know.

dawid-sw (changes requested):
Consider dropping the new `last_activity_ts` field and deriving the inactivity window from the existing `LiquidationEntry.timestamp` values instead - just take the max timestamp across all four entries.

This avoids a struct modification and sidesteps a migration concern: every `LiquidationRecord` that already exists on-chain will have `last_activity_ts == 0` (since the field is carved out of previously-zeroed reserved bytes), which means the creator can close it immediately - bypassing the intended 60-day wait.

There is one edge case with the entries-based approach: if a record has never been used for liquidation, all entry timestamps are zero, so the creator could close it immediately. In theory, an account owner could exploit this to front-run an incoming liquidation - create a record, then close it right before a liquidator tries to use it, temporarily blocking the liquidation. In practice this is unlikely, and if it becomes a concern, we can restrict closure of empty records to risk-admin only and handle it with off-chain logic.

jgur-psyops:
>  just take the max timestamp across all four entries.

They should always be chronological, so the most recent should suffice.

> every `LiquidationRecord` that already exists on-chain will have `last_activity_ts == 0` (since the field is carved out of previously-zeroed reserved bytes), which means the creator can close it immediately - bypassing the intended 60-day wait.

This is probably fine: we didn't have this close policy earlier, but we'll "play fair" OTC here and give record creators at least 30 days to get their rent (if they want it). There are only a few hundred records currently, so it's not a huge amount of rent.

> There is one edge case with the entries-based approach: if a record has never been used for liquidation, all entry timestamps are zero, so the creator could close it immediately.

That seems fair: if the record has never been used, you can close it instantly. This allows a liquidator to reclaim after a failed liquidation. While the owner could open/close spam to try to deny an account liquidation, the liquidator only needs to land one open to stop your denial. Once the liquidator lands one, the owner cannot create a new one nor close the old one, because they are not the record payer.

dawid-sw:
@jgur-psyops so you are voting for a new field or using the existing timestamp in the liquidation entries?

jgur-psyops:
superseded by #560

PR #516, opened 2026-02-21
JupLend cleanup

IliaZyrin:
Also:
- updates Kamino IDLs and fixes the instructions & tests accordingly,
- [forbids](https://github.com/mrgnlabs/marginfi-v2/pull/516/changes#diff-72cc8cf3980f03d6e102e5beef1db9a576e08f5c27c9656a742a5c2ea8bc585bR184) users to take receivership of their own accounts.

### DIFF
--- a/programs/marginfi/src/instructions/drift/deposit.rs
+++ b/programs/marginfi/src/instructions/drift/deposit.rs
@@ -1,5 +1,5 @@
 use crate::{
-    bank_signer, check,
+    bank_signer,
     constants::DRIFT_PROGRAM_ID,
     events::{AccountEventHeader, LendingAccountDepositEvent},
     state::{


programs/marginfi/src/instructions/juplend/deposit.rs
```diff
@@ -47,12 +50,6 @@ pub fn juplend_deposit(ctx: Context<JuplendDeposit>, amount: u64) -> MarginfiRes
 
         validate_asset_tags(&bank, &marginfi_account)?;
         validate_bank_state(&bank, InstructionKind::FailsIfPausedOrReduceState)?;
-
-        check!(
-            !marginfi_account.get_flag(ACCOUNT_DISABLED)
-                && !marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
```
jgur-psyops, line 53: regression: the receivership check is no longer enforced in constraints (same for Kamino)

IliaZyrin: It is enforced in `is_signer_authorized`

jgur-psyops: There's a small edge case with `is_signer_authorized` where the authority can call kamino deposit during receivership of themselves. 

Totally benign I think, maybe the best approach is just to add a check to `is_signer_authorized` that rejects if `marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP)` and the signer is the account's own authority to mitigate this edge case.

jgur-psyops: Alternatively leave is_signer_authorized as-is and add a defensive 

```    check!(
        !marginfi_account.get_flag(ACCOUNT_DISABLED)
            // Sanity check: liquidation doesn't allow the deposit ix, but just in case
            && !marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
        MarginfiError::AccountDisabled
    );
``` 

to the drift/kamino deposits like the regular deposit has.

programs/marginfi/src/instructions/juplend/withdraw.rs
dawid-sw, line 71: in kamino and drift we are double checking remaining accounts using `validate_remaining_accounts_for_balances_close_last`, without this check if wrong accounts are passed it will fail anyway because of checks in `check_account_init_health` -> `get_health_components`. 

so maybe worth to add this check as well to fail earlier?

IMEF-FEMI: this check has actually been removed in the incoming PR #525 . it was needed because we wanted to 
force the caller of the repay and withdraw instructions to pass the risk accounts for all balances in the lending account (including the ones being closed in withdraw-all and repay-all paths)

.vscode/tasks.json
IMEF-FEMI:
```suggestion
            "command": "./scripts/build-workspace.sh && yarn install && anchor build -p marginfi -- --no-default-features --features custom-heap && anchor build -p test_transfer_hook --no-idl && anchor build -p mocks && anchor test --skip-build",
```

PR #521, opened 2026-02-25
Remove  and cleanup emissions

dawid-sw:
Summary

- Remove on-chain emissions accounting (claim, settle, withdraw) - no longer used
- Add `lending_pool_reclaim_emissions_vault`: admin drains remaining vault tokens and disables emissions on a bank
- Add `lending_account_clear_emissions`: permissionlessly zero out `emissions_outstanding` on a balance after emissions are disabled
- Keep `marginfi_account_update_emissions_destination_account` for off-chain emissions distribution
- Remove `claim_emissions calls from deposit`, `withdraw`, `repay`, `borrow`, and `close_balance` paths
- Simplify `Balance::close()` - remove unused `check_emissions` parameter
- Remove emissions gate from limit order placement
- Clean up CLI, TS test utils, and Rust test helpers

From our research, outstanding emissions are old and not significant (dust), so we can remove them. And all banks on production have emissions turned off. So the plan would be:
1. Get outstanding dust emissions from banks.
2. Clear user `outstanding_emissions` - zeroing out this field so we can use it later on for other purposes.

The `marginfi_account_update_emissions_destination_account` IX stays as it is used by current integrators to point-out accounts that should receive off-chain emissions.

### DIFF
--- a/guides/DEVELOPERS_INTEGRATORS/GETTING_STARTED_DEV.md
+++ b/guides/DEVELOPERS_INTEGRATORS/GETTING_STARTED_DEV.md
@@ -90,7 +90,7 @@ Benchmarks:
 
 0.1.6
 
-| 9700X (8 threads)  | `[  27.718s] 373 tests run: 373 passed, 0 skipped`
+| 9700X (8 threads) | `[  27.718s] 373 tests run: 373 passed, 0 skipped`
 
 | 9700X (16 threads) | `[  19.343s] 373 tests run: 373 passed (3 flaky), 0 skipped`
 
@@ -165,7 +165,8 @@ Ensure your machine is not in Low Power battery mode (or in any other mode decre
 
 On slower machines, or in tests with many txes, this error can be consistent or sometimes intermittent. Try
 refreshing the blockhash in longer tests: `test_f.refresh_blockhash().await;` and switching usage of
-`ctx.last_blockhash` to 
+`ctx.last_blockhash` to
+
 ```
 let blockhash = {
     let banks_client = self.ctx.borrow().banks_client.clone();
@@ -176,6 +177,7 @@ let blockhash = {
 ## Validator Crashes at Startup
 
 Usually manifests as something like:
+
 ```
 Starting bankrun with pure bankrun setup...
 thread 'tokio-runtime-worker' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-program-test-1.18.0/src/lib.rs:716:17:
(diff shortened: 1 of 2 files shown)


programs/marginfi/src/instructions/marginfi_account/emissions.rs
```diff
@@ -1,365 +1,77 @@
...
+    /// the canonical ATA for each emissions mint.
     pub destination_account: AccountInfo<'info>,
 }
 
-/// Permissionlessly withdraw emissions to user emissions_destination_account
-pub fn lending_account_withdraw_emissions_permissionless<'info>(
-    ctx: Context<'_, '_, 'info, 'info, LendingAccountWithdrawEmissionsPermissionless<'info>>,
+/// Permissionlessly zero out `emissions_outstanding` on a balance after
+/// emissions have been disabled on the bank.
+pub fn lending_account_clear_emissions(
```
jgur-psyops, line 46: I think we need an instruction like this at the bank level as well, to zero `emissions_rate`, `emissions_remaining`, and `emissions_mint`. The flags `EMISSIONS_FLAG_BORROW_ACTIVE` and `EMISSIONS_FLAG_LENDING_ACTIVE` should also be zeroed and the naming set to SOMETHING_DEPRECATED

programs/marginfi/src/instructions/marginfi_account/emissions.rs
```diff
@@ -1,365 +1,77 @@
...
-) -> MarginfiResult {
-    let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
-
-    check!(
-        !marginfi_account.get_flag(ACCOUNT_DISABLED),
-        MarginfiError::AccountDisabled
-    );
-
-    check!(
-        !marginfi_account.get_flag(ACCOUNT_FROZEN),
```
jgur-psyops, line 200: Nit: we used to forbid this when the account is frozen. If the goal of account freezing is e.g. legal compliance might as well freeze this too.

programs/marginfi/src/instructions/marginfi_account/emissions.rs
jgur-psyops: Nvm vault withdraw already does this, minus the mint.

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -81,70 +81,71 @@ pub struct LendingPoolConfigureBank<'info> {
...
+            vault_balance,
+            ctx.accounts.emissions_mint.decimals,
+        )?;
+    }
 
-    bank.emissions_rate = emissions_rate;
-    bank.emissions_remaining = I80F48::from_num(total_emissions).into();
+    bank.emissions_remaining = I80F48::ZERO.into();
+    bank.emissions_rate = 0;
+    bank.flags &= !EMISSION_FLAGS;
```
jgur-psyops, line 144: reset bank.emissions_mint to pubkey default as well.

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -81,70 +81,71 @@ pub struct LendingPoolConfigureBank<'info> {
...
+            &[ctx.bumps.emissions_auth],
+        ]];
 
-    bank.override_emissions_flag(emissions_flags);
+        transfer_checked(
+            CpiContext::new_with_signer(
+                ctx.accounts.token_program.to_account_info(),
+                TransferChecked {
+                    from: ctx.accounts.emissions_vault.to_account_info(),
+                    to: ctx.accounts.destination_account.to_account_info(),
```
jgur-psyops, line 131: let's just set the destination to the canonical ATA of the global fee wallet and make this ix permissionless (it barely matters, I think we have like three figures in these vaults)

That would enable us to clone a few of these banks and user accounts from mainnet and write a small test on localnet for both the user and bank vault zeroing out.

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -81,70 +81,71 @@ pub struct LendingPoolConfigureBank<'info> {
...
+pub fn lending_pool_reclaim_emissions_vault(
+    ctx: Context<LendingPoolReclaimEmissionsVault>,
 ) -> MarginfiResult {
     let mut bank = ctx.accounts.bank.load_mut()?;
 
     check!(
-        bank.emissions_mint.eq(&Pubkey::default()),
-        MarginfiError::EmissionsAlreadySetup
+        bank.emissions_mint.ne(&Pubkey::default()),
+        MarginfiError::EmissionsUpdateError,
```
jgur-psyops: This can probably no-op and return early instead of erroring.

jgur-psyops (approved):
lgtm!

Note for mainnet when we operate this: 
(1) remember to create token indempotent for emissions_mint on the global_fee_wallet in each `lending_pool_reclaim_emissions_vault` tx
(2) `lending_pool_reclaim_emissions_vault` will fail if emissions never existed because `emissions_vault` would not exist, but it will no-op if emissions already sunset (i.e. we do not close `emissions_vault`). Just loop through all banks and try/catch sending `lending_pool_reclaim_emissions_vault`, ignore failures, and then validate that `emissions_mint` is pubkey default for all banks at conclusion and re-send any if not.

PR #522, opened 2026-02-26
Same bank emissions

abelmarnk:
Summary
- Adds a permissionless `lending_pool_emissions_deposit` instruction to allow anyone to deposit into a bank's liquidity vault and increase the asset share value.
- Add Rust & TS tests for the instruction

Notes:- 
- Some new helpers are added to the tests to allow for setting up emissions on a bank, since #521 would be removing all emissions related instructions and helpers.
- Denies banks that are non-standard from calling it.
- Denies bank mints with active transfer hooks or transfer fees from calling it.
- The changes only affects lenders.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
@@ -1,4 +1,3 @@
-use crate::{check, math_error, utils};
 use crate::events::{
     GroupEventHeader, LendingPoolBankConfigureEvent, LendingPoolBankConfigureFrozenEvent,
 };
@@ -6,7 +5,9 @@ use crate::prelude::MarginfiError;
 use crate::state::bank::BankImpl;
 use crate::state::emode::EmodeSettingsImpl;
 use crate::state::marginfi_group::MarginfiGroupImpl;
+use crate::utils::is_marginfi_asset_tag;
 use crate::MarginfiResult;
+use crate::{check, math_error, utils};
 use anchor_lang::prelude::*;
 use anchor_spl::associated_token::get_associated_token_address_with_program_id;
 use anchor_spl::token_2022::{transfer_checked, TransferChecked};
@@ -295,6 +296,8 @@ pub struct LendingPoolEmissionsDeposit<'info> {
         has_one = group @ MarginfiError::InvalidGroup,
         has_one = emissions_mint @ MarginfiError::InvalidEmissionsMint,
         has_one = liquidity_vault @ MarginfiError::InvalidLiquidityVault,
+        constraint = is_marginfi_asset_tag(bank.load()?.config.asset_tag)
+            @ MarginfiError::WrongAssetTagForStandardInstructions,
     )]
     pub bank: AccountLoader<'info, Bank>,
 
--- a/programs/marginfi/tests/misc/emissions_deposit.rs
+++ b/programs/marginfi/tests/misc/emissions_deposit.rs
@@ -10,7 +10,9 @@ async fn emissions_deposit_fails_when_bank_paused() -> anyhow::Result<()> {
 
     let usdc_bank = test_f.get_bank(&BankMint::Usdc);
 
-    usdc_bank.set_emissions_direct(usdc_bank.mint.key).await?;
+    usdc_bank
+        .set_emissions(usdc_bank.mint.key, 0, I80F48::ZERO, 0)
+        .await;
 
     let depositor = test_f.create_marginfi_account().await;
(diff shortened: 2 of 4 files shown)


programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -289,3 +347,37 @@ pub struct LendingPoolUpdateEmissionsParameters<'info> {
...
+            !group.load()?.is_protocol_paused()
+        ) @ MarginfiError::ProtocolPaused
+    )]
+    pub group: AccountLoader<'info, MarginfiGroup>,
+
+    #[account(
+        mut,
+        has_one = group @ MarginfiError::InvalidGroup,
+        has_one = emissions_mint @ MarginfiError::InvalidEmissionsMint,
+        has_one = liquidity_vault @ MarginfiError::InvalidLiquidityVault,
```
jgur-psyops, line 298: nit: let's also add constraints to forbid if the bank is in Paused or ReduceOnly or Killed states

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -254,6 +255,63 @@ pub fn lending_pool_update_emissions_parameters(
...
+pub fn lending_pool_emissions_deposit(
+    ctx: Context<LendingPoolEmissionsDeposit>,
+    amount: u64,
+) -> MarginfiResult {
+    if amount == 0 {
+        return Ok(());
+    }
+
+    let mut bank = ctx.accounts.bank.load_mut()?;
+
```
jgur-psyops, line 221: I think we must accrue interest before we apply the emission. 

Consider this example
```
Setup:
* total assets = 1,000,000
* total liabilities = 900,000
utilization rate U = liabilities / assets = 90%

Let's assume:
* at 90% util: borrow APR = 60%
* at 75% util: borrow APR = 30%

Interest for one day expected:
900,000 * 0.60 / 365 = 1,479.45

Now instead, someone calls `lending_pool_emissions_deposit` for 200,000
* total assets = 1,200,000
* total liabilities = 900,000
U is now 75%

900,000 * 0.30 / 365 = 739.73
```
Borrowers got a windfall of ~$700

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -254,6 +255,63 @@ pub fn lending_pool_update_emissions_parameters(
...
+                from: ctx.accounts.emissions_funding_account.to_account_info(),
+                to: ctx.accounts.liquidity_vault.to_account_info(),
+                authority: ctx.accounts.depositor.to_account_info(),
+                mint: ctx.accounts.emissions_mint.to_account_info(),
+            },
+        ),
+        amount,
+        ctx.accounts.emissions_mint.decimals,
+    )?;
+
```
jgur-psyops, line 264: Here we must either handle T22 fees and transfer hooks, or simply error if there are any transfer hooks enabled or a non-zero transfer fee. Since these have been pretty unpopular on Solana to date, the latter is probably fine until some asset comes up that actually uses these mint features.

Note: PyUSD uses fees, but the fee is zero, so we should still support Transfer Fees that are zero.

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
abelmarnk: Addressed [here](https://github.com/mrgnlabs/marginfi-v2/pull/522/changes/4f631bdb011c39050669dfdf4bee47bf0e863e2c#diff-20674d6656254a5f05074cbd86078be9115a26c4bf8379a0d1067ace9c6565c0R296).

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
abelmarnk: Addressed [here](https://github.com/mrgnlabs/marginfi-v2/pull/522/changes/4f631bdb011c39050669dfdf4bee47bf0e863e2c#diff-20674d6656254a5f05074cbd86078be9115a26c4bf8379a0d1067ace9c6565c0R279).

programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
abelmarnk: Resolved [here](https://github.com/mrgnlabs/marginfi-v2/pull/522/changes/4f631bdb011c39050669dfdf4bee47bf0e863e2c#diff-20674d6656254a5f05074cbd86078be9115a26c4bf8379a0d1067ace9c6565c0R272).

PR #526, opened 2026-03-06
Same bank emissions

jgur-psyops:
Thank you to @abelmarnk for the initial PR (#522)

Changes here:

- Now checks against just `mint` instead of `emissions_mint`, since the latter is being sunset.
- Added some tests for paused/T22/no deposits edge cases, notable demoing we do not support (or intend to) T22 fees with this feature.

Note: see commit [feccd9a](https://github.com/mrgnlabs/marginfi-v2/pull/526/commits/feccd9a559e0f828c1436814345b282ab69097a8) for changes since #522


programs/marginfi/src/instructions/marginfi_group/configure_bank.rs
```diff
@@ -200,3 +202,113 @@ pub struct LendingPoolReclaimEmissionsVault<'info> {
...
+        .ok_or_else(math_error!())?;
+
+    bank.asset_share_value = updated_total_assets
+        .checked_div(total_asset_shares)
+        .ok_or_else(math_error!())?
+        .into();
+
+    bank.update_bank_cache(&group)?;
+
+    msg!(
```
dawid-sw, line 272: is there any reason why msg! is used and not anchor event? event is easier for indexing

jgur-psyops: historically we just use these to make the ix easier review in squads, probably not a bad idea to make all of these config outputs into events.

dawid-sw: yeah worth to consider refactor to make indexing easier 👍

PR #527, opened 2026-03-07
Kamino Refresh (External/CPI)

abelmarnk:
## Summary

This PR adds three new instructions:- `deposit_with_refresh`, `withdraw_with_refresh` and `init_obligation_batch_refresh` that bundle the Kamino reserve/obligation refresh CPIs **inside** the Marginfi instruction itself.  The goal of adding the first two is to evaluate whether or not consolidating these refresh calls into a single CPI instruction is worth the trade-off versus the current pattern (separate external refresh instructions in the same transaction), for the third it is to demonstrate the difference between using the regular `refresh_reserve` and   `refresh_reserves_batch` 

This PR is not intended to be merged, this branch is purely for benchmarking and
comparison.

---

## New instructions

| Instruction | Description |
|---|---|
| `kamino_deposit_with_refresh` | Internally calls `refresh_reserves_batch` + `refresh_obligation` via CPI before depositing |
| `kamino_withdraw_with_refresh` | Internally calls `refresh_reserves_batch` + `refresh_obligation` via CPI before withdrawing |
| `init_obligation_batch_refresh` | It calls `refresh_reserves_batch` instead of `refresh_reserves` to skip price updates |

They all use the `refresh_reserves_batch` which saves more CU, particularly for the `init_obligation` instruction it saves ~47,000-52,000 CU(The exact amounts for the `deposit` and `withdraw` instructions were not measured since they do not call the refresh instructions internally) when compared to the simple `refresh_reserves` variant, the reason is that the regular variant forces us to update the price even though it is not necessary.

For deposit:- It is not necessary because a deposit only adds collateral so there is no need to evauluate health or similar metrics([Kamino does not check this](https://github.com/Kamino-Finance/klend/blob/master/programs/klend/src/handlers/handler_deposit_reserve_liquidity_and_obligation_collateral.rs#L94), also see [here](https://github.com/Kamino-Finance/klend/blob/master/programs/klend/src/lending_market/lending_operations.rs#L118) and [here](https://github.com/Kamino-Finance/klend/blob/master/programs/klend/src/lending_market/lending_operations.rs#L383)).

For withdraw:- It is not necessary because Marginfi uses integrations in such a way that it does not borrow from the pools, it only deposits and withdraws, kamino allows us to [withdraw without evaluating health](https://github.com/Kamino-Finance/klend/blob/master/programs/klend/src/handlers/handler_withdraw_obligation_collateral_and_redeem_reserve_collateral.rs#L100)(also see [here](https://github.com/Kamino-Finance/klend/blob/master/programs/klend/src/lending_market/lending_operations.rs#L493)) as long as we don't have any borrows, so it doesn't enforce that the price must be up to date.

For init-obligation:- A similar reasoning for deposit applies.

As a small note building the instruction **manually** instead of using Anchor's `CpiContext` saves an extra ~500-1000 CU(the current implementation does not do this).

---

## Benchmark results

Measured with solana-bankrun.  Clock slot and timestamp are advanced by the same amount before each measurement (External and CPI) so both instructions take similar paths after seeing stale values.

```
  ID   Grp       A CU       B CU       Δ-CU   Δ-CU%  A size  B size Δ-size A ix count B ix count Δix count  A / B
  S01  A      226,424    231,818     +5,394    2.4%    851B    823B   -28      4      2   -2  External / CPI
  S02  A      266,235    271,622     +5,387    2.0%    987B    959B   -28      4      2   -2  External / CPI
  S03  B      443,615    454,405    +10,790    2.4%    920B    864B   -56      7      3   -4  External / CPI
  S04  C          N/A        N/A        N/A     N/A  >1232B  >1232B   N/A      6      3   -3  External / CPI
  S05  C          N/A        N/A        N/A     N/A  >1232B  >1232B   N/A      6      3   -3  External / CPI
  S06  D      537,918    542,890     +4,972    0.9%   1119B   1091B   -28      5      3   -2  External / CPI
  S07  E       18,004     11,293     -6,711  -37.3%    319B    284B   -35      2      2   +0  Simple / Batch
  S08  F      335,703    288,278    -47,425  -14.1%    892B    856B   -36      2      2   +0  Simple / Batch
  S09  F      449,202    397,101    -52,101  -11.6%    956B    920B   -36      2      2   +0  Simple / Batch
```

**Groups A–D** (`A/Ext` = External, `B/CPI` = internal refresh instructions):
- CPI versions cost **~5,000** more (~2–3%) than the external pattern.
- CPI versions produce **smaller transactions** (−28 to −56 bytes) and require **fewer instructions** (−2 to −4), because the refresh calls are folded into the instruction itself rather than preceding it in the transaction.
- S05 (2-bank deposit+withdraw) exceeds the 1232-byte transaction size limit regardless of path; both variants fail to serialize.

**Some reasons CPI costs more CU:**
1. Cross-program invocation overhead:- The Solana runtime charges a flat cost for each CPI call, paid by the caller instruction.
2. On-chain account/instruction construction:- The Marginfi instruction builds the instruction accounts and data at runtime, whereas the external pattern assembles these off-chain and passes them in as ordinary instructions.

**Group E  `refresh_reserve` vs `refresh_reserves_batch` (S07):**
- A single `refresh_reserve` costs **18,004 CU** and 319 bytes.
- A single `refresh_reserves_batch` costs **11,293 CU** and 284 bytes.
- Batch is **~37% cheaper**, because `refresh_reserve` also uses oracle price data while `refresh_reserves_batch` with `skip_price_updates=true` skips that work entirely.
- This demonstrates that `refresh_reserves_batch` is strictly preferable whenever there is no need to refresh the price which is the case for the deposit and withdraw(same applies to the `init_obligation`, see #528). 

**Group F `init_obligation` vs `init_obligation_batch_refresh`**
- Uses `refresh_reserves_batch` instead of `refresh_reserves`
- Removes all oracle based accounts
- As per the benchmarks `init_obligation` costs 335,703 CU and 892 bytes without farms while `init_obligation_batch_refresh` costs  288,278 CU and 856 bytes, a difference of 47,425 CU, with farms `init_obligation` costs 449,202 CU and 956 bytes while `init_obligation_batch_refresh` costs  397,101 and 920 bytes, a difference of -52,101.
- One reason for the additional difference in CU is the handling of the optional accounts, both by Anchor(when they are passed) and in instruction handler.

## How to run the benchmarks

The program must be built first:

```bash
anchor build -p marginfi -- --no-default-features
```

Then run the prerequisite setup tests followed by the scenario benchmarks:

```bash
./scripts/run-kamino-scenarios.sh
```

Or equivalently:

```bash
yarn run ts-mocha -p ./tsconfig.json -t 1000000 \
  tests/k01_kaminoInit.spec.ts \
  tests/k02_kaminoUser.spec.ts \
  tests/k03_kaminoDeposit.spec.ts \
  tests/k04_kaminoMrgnUser.spec.ts \
  tests/k05_kaminoBankInit.spec.ts \
  tests/k06_MrgnKaminoDeposit.spec.ts \
  tests/k13_KfarmsHarvestReward.spec.ts \
  tests/k_scenarios.spec.ts \
  --exit --require tests/rootHooks.ts
```

The tests should not be run in the regular way using `anchor test --skip-build` unless they would fail, since the scenario test makes some assumptions about the current state of the test.

The full suite completes in about 5 seconds.

The results of the tests can be seen from the CI.


.github/workflows/test.yaml
```diff
@@ -25,151 +25,11 @@ concurrency:
...
 
 jobs:
-  lint:
-    name: Rust Lint
+  kamino-scenarios:
+    name: Build & Run Kamino Scenario Benchmarks
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: ./.github/actions/base-setup/
```
jgur-psyops, line 33: can restore this file to its original form (k_scenarios will run as part of the k-suite)

programs/marginfi/src/instructions/kamino/withdraw_with_refresh.rs
```diff
@@ -0,0 +1,551 @@
...
+        let cpi_ctx = CpiContext::new_with_signer(program, accounts, signer_seeds);
+        let decimals = self.reserve_liquidity_mint.decimals;
+        transfer_checked(cpi_ctx, amount, decimals)?;
+        Ok(())
+    }
+}
+
+impl Hashable for KaminoWithdrawWithRefresh<'_> {
+    fn get_hash() -> [u8; 8] {
+        get_discrim_hash("global", "kamino_withdraw_with_refresh")
```
jgur-psyops, line 549: should be added to the allowlist of `validate_ixes_exclusive` in `liquidate_start`.

jgur-psyops: similarly to the constants in `ix_discriminators` (this is a silly thing we do just to have them all the ones people might actually care about in one place for easy viewing).

programs/marginfi/src/instructions/kamino/init_obligation_batch_refresh.rs
```diff
@@ -0,0 +1,323 @@
...
+    init_obligation_farms_for_reserve, init_user_metadata, refresh_obligation,
+    refresh_reserves_batch,
+};
+use kamino_mocks::kamino_lending::types::InitObligationArgs;
+use kamino_mocks::state::MinimalReserve;
+use marginfi_type_crate::constants::LIQUIDITY_VAULT_AUTHORITY_SEED;
+use marginfi_type_crate::types::Bank;
+
+/// Initialize a Kamino obligation for a marginfi account using batch refresh
+pub fn kamino_init_obligation_batch_refresh(
```
jgur-psyops, line 28: While the rest of the this PR is very useful, we *probably* wouldn't update the init_obligation instruction since it only runs once-per-bank and thus we don't really care about CU performance. The benefit does not outweigh the audit bill.

programs/marginfi/src/instructions/kamino/deposit_with_refresh.rs
```diff
@@ -0,0 +1,376 @@
...
+        validate_bank_state(&bank, InstructionKind::FailsIfPausedOrReduceState)?;
+
+        check!(
+            !marginfi_account.get_flag(ACCOUNT_DISABLED)
+                && !marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
+            MarginfiError::AccountDisabled
+        );
+    }
+
+    ctx.accounts.cpi_refresh_reserve_batch()?;
```
jgur-psyops, line 78: It looks like the diff between classic deposit/withdraw and the novel approach is just these two lines, it would likely be better for us to simply add a new `refresh_mode: u8` or similar arg to the existing deposit/withdraw function and run these lines if the caller requested it (some callers will likely prefer to build their own refresh).

The change is breaking, but trivial for integrators, whereas maintaining two deposit/withdraw instructions will probably get rather annoying.

abelmarnk:
The external uses the same number of accounts and less CU and is generally more flexible, and in the case where multiple reserves are being refreshed they can be refreshed with a single instruction using the batch refresh, the CPI only allows a single one at a time and would use more CU. The only drawback from keeping things as they are is the slight increase in transaction size ~28-56. So i thought it would be better to just leave that as it is.

PR #528, opened 2026-03-07
Use batch refresh for `init_obligation`

abelmarnk:
**Summary**
This PR replaces the `refresh_reserves` with `batch_refresh_reserves` in the kamino `init_obligation` instruction. The replacement is done because the batch variant allows for skipping price updates which is not required for the deposit that is made when an obligation is created, it saves about [~47,000-52,000 CU](https://github.com/mrgnlabs/marginfi-v2/actions/runs/22818107641/job/66186388194?pr=527#step:11:203).

Some reasons for the difference include:-
- Skipping price updates
- Removing the handling of optional accounts

See #527 for some more details.


dawid-sw:
LGTM, the only note - would be nice to add some doc string to IX with explanation of remaining accounts that are needed, so its easier for integrators to reason about it.

PR #530, opened 2026-03-09
Scope liquidation cache lock clearing to receivership accounts

dawid-sw:
`withdraw_all`, `repay_all`, and `close_balance` now only clear the bank's `liq_cache_locked` flag when the account has `ACCOUNT_IN_RECEIVERSHIP` set. Previously, the lock was cleared unconditionally, which meant a non-receivership account sharing the same bank could inadvertently clear the lock during a liquidation transaction. close_balance had no lock-clearing logic at all.

  Changes
- `withdraw_all`, `repay_all`, `close_balance` take `in_receivership`: bool and gate lock clearing on it
- All callers pass `marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP)`
- Updated tests to assert that cross-account operations do not interfere with liquidation
- Updated `liq_cache_flags` doc comment

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
@@ -2,14 +2,14 @@ use anchor_lang::prelude::*;
 use fixed::types::I80F48;
 use marginfi_type_crate::{
     constants::{TOKENLESS_REPAYMENTS_COMPLETE, ZERO_AMOUNT_THRESHOLD},
-    types::{Bank, MarginfiAccount, MarginfiGroup},
+    types::{Bank, MarginfiAccount, MarginfiGroup, ACCOUNT_IN_RECEIVERSHIP},
 };
 
 use crate::{
     prelude::*,
     state::{
         bank::BankImpl,
-        marginfi_account::{BalanceImpl, LendingAccountImpl},
+        marginfi_account::{BalanceImpl, LendingAccountImpl, MarginfiAccountImpl},
     },
     utils::is_marginfi_asset_tag,
 };
@@ -20,6 +20,7 @@ pub fn lending_account_purge_delev_balance(
     let bank_pk = &ctx.accounts.bank.key();
     let mut bank = ctx.accounts.bank.load_mut()?;
     let mut marginfi_account = ctx.accounts.marginfi_account.load_mut()?;
+    let in_receivership = marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP);
     let lending_account = &mut marginfi_account.lending_account;
 
     let balance = lending_account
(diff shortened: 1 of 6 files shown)


programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
```diff
@@ -38,6 +39,9 @@ pub fn lending_account_purge_delev_balance(
     let asset_shares: I80F48 = balance.asset_shares.into();
     msg!("Balance had: {:?}", asset_shares.to_num::<f64>());
     balance.close()?;
+    if in_receivership {
```
jgur-psyops, line 42: Added this to be on the safe side, not that the admin is likely to call this by CPI during deleverage or anything but since we're covering edge cases I suppose we might as well!

programs/marginfi/tests/user_actions/liquidate_receiver.rs
```diff
@@ -1359,15 +1365,326 @@ async fn liquidate_receiver_other_account_repay_all_clears_bank_cache_lock() ->
...
+    let usdc_bank_state = usdc_bank.load().await;
+    assert!(!sol_bank_state.cache.is_liquidation_price_cache_locked());
+    assert!(!usdc_bank_state.cache.is_liquidation_price_cache_locked());
+    Ok(())
+}
+
+// Same lock-preservation invariant as above, but with a true second user (distinct authority signer).
+// The second user performs withdraw_all on a shared bank during A's receivership liquidation.
+#[tokio::test]
+async fn liquidate_receiver_second_user_signed_withdraw_all_does_not_clear_bank_cache_lock(
```
jgur-psyops, line 1527: I was curious if this would actually work and was somewhat surprised that it does. Not sure we actually want a second user interacting like this. The audit report suggests allowing it for flexibility reasons, but I'm debating if it adds to the attack surface without any real benefit, perhaps it's preferable to ban interact with banks that in the receivership flow so this test would fail instead. 

cc @IliaZyrin @dawid-sw if you have thoughts either way. LGTM regardless.

dawid-sw: Interesting scenario. After tracing through what happens, it seems that there's no direct risk - the `in_receivership` guard prevents non-receivership accounts from clearing the lock, and the liquidation-critical state is untouched.

That said, I think banning it is the safer choice. It gives us a simpler invariant - locked banks are only touched by the receivership flow - and eliminates an entire class of edge cases to reason about. The only tradeoff is that liquidators can't withdraw from their own accounts to fund repayment directly, but they can achieve the same outcome via Jupiter or a pre-funded wallet. We can always relax it later if there's demand.

IliaZyrin: It looks good to me, but I think there is one more exotic edge case - simultaneous deleverage and liquidation in the same transaction done by two different entities (risk_admin and some regular liquidator - for two different marginfi accounts sharing the same bank). Thought that it is also possible - to submit two simultaneous independent liquidations but there is a guard for it so my test failed (which is good!), now tweaking it for deleverage case.

IliaZyrin: Nevermind, mixing deleverage & liquidation is impossible, we gracefully fail with ForbiddenIx. So all good, approving!

PR #531, opened 2026-03-11
Minimal IDL generation for mocks programs

jgur-psyops:
WARN: You probably need to `anchor clean` before this will build on your local machine, or you will be blocked by build artifacts. 

## Summary

This is a silly update that adds a python script which removes everything from the IDL files we don't need. Mock programs now declare that IDL instead of the "real" one, which enables them to compile very fast and without spamming the console with stack errors.

We now have two IDL folders: one with "complete" IDLs for the TS suite, another with "minimal" IDLs for the declare program macro. See the readme in scripts for more details. All Rust code, including the test fixtures, now use the Minimal versions of any integration struct. All TS tests continue to use the actual IDL as it appears on mainnet (or a fixed version of that IDL, in Kamino's case)

The incremental compile time of `anchor build -p marginfi --no-idl -- --no-default-features` is not actually improved since the mocks programs rarely (almost never) rebuild, but the full build time is down slightly, so that's something.

There are no program changes here, except that Kamino instructions have renamed some of the types to sidestep the duplicate type names in the Kamino IDL when generating the minimal one.


IliaZyrin (commented):
LGTM!

PR #534, opened 2026-03-15
Cli program revamp

IMEF-FEMI:
## Summary

  This PR is a broad cleanup and expansion of marginfi-cli.

  It refactors the CLI into clearer command and processor modules, improves help output and config examples, and reduces redundant inputs by deriving accounts where
  possible instead of requiring them in configs.

  It also adds a GitHub Actions release workflow for packaging and publishing the mfi binary, and updates the README.md and added an AGENTS doc to better reflect the actual command surface and usage.


jgur-psyops (approved):
Looks great! Let's merge for 1.8 and we can make any updates as needed, no specific changes needed here.

clients/rust/marginfi-cli/assets/assetGroups.json
```diff
@@ -0,0 +1,84 @@
+{
+  "blue-chip": {
```
IliaZyrin, line 2: Shouldn't we better fetch it from some API rather than hardcoding like this?

clients/rust/marginfi-cli/src/commands/account.rs
```diff
@@ -0,0 +1,502 @@
...
+    after_long_help = "Common subcommands:\n  mfi account list\n  mfi account get <ACCOUNT_PUBKEY>\n  mfi account deposit <BANK_PUBKEY> 10\n  mfi account withdraw <BANK_PUBKEY> 5\n  mfi account borrow <BANK_PUBKEY> 3\n  mfi account repay <BANK_PUBKEY> 3\n  mfi account place-order --bank-1 <BANK_PUBKEY> --bank-2 <BANK_PUBKEY> --trigger-type stop-loss --stop-loss 0.9 --max-slippage-bps 50"
+)]
+pub enum AccountCommand {
+    /// List all marginfi accounts for the current authority
+    ///
+    /// Example: `mfi account list`
+    #[clap(
+        after_help = "Example:\n  mfi account list",
+        after_long_help = "Example:\n  mfi account list"
+    )]
```
IliaZyrin, line 76: Nitpick: let's just leave one of the three, given that they're the same? E.g. only `after_help`?

clients/rust/marginfi-cli/src/commands/account.rs
```diff
@@ -0,0 +1,502 @@
...
+    LiquidateReceivership {
+        #[clap(long)]
+        liquidatee_marginfi_account: Pubkey,
+        /// If set, auto-add init_liq_record if missing
+        #[clap(long, default_value_t = false)]
+        init_liq_record_if_missing: bool,
+        /// Optional path to JSON file with extra instructions placed between start/end
+        #[clap(long)]
+        extra_ixs_file: Option<PathBuf>,
```
IliaZyrin, line 268: Just a thought: can we maybe do it "interactive" instead of passing the file with extra ixes?
E.g. hit a command for start, then hit the withdraw/repay commands which will be accumulated, and then the end command would actually submit the whole batch?

clients/rust/marginfi-cli/src/commands/drift.rs
```diff
@@ -0,0 +1,550 @@
...
+        config_example: bool,
+        bank_pk: Option<Pubkey>,
+        #[clap(long, help = "Native amount for seed deposit (minimum 10)")]
+        amount: Option<u64>,
+    },
+    /// Deposit into Drift via marginfi
+    ///
+    /// Example: `mfi drift deposit --config ./configs/drift/deposit/config.json.example`
+    #[clap(
+        after_help = "Example:\n  mfi drift deposit --config ./configs/drift/deposit/config.json.example",
```
IliaZyrin, line 57: Just a general comment on this: shouldn't we use `p0` instead of `mfi`?

clients/rust/marginfi-cli/src/commands/drift.rs
```diff
@@ -0,0 +1,550 @@
...
+        anyhow::bail!(
+            "Drift spot market account {} data too small ({} bytes)",
+            spot_market,
+            spot_market_data.len()
+        );
+    }
+    let spot_market_state: &MinimalSpotMarket =
+        bytemuck::from_bytes(&spot_market_data[8..8 + spot_market_size]);
+
+    let (drift_state, _) = Pubkey::find_program_address(&[b"drift_state"], &DRIFT_PROGRAM_ID);
```
IliaZyrin, line 199: Should we factor these PDA derivation functions out to the type crate and import from there? To keep a single source of truth since these may change (as Kamino showed recently).

IliaZyrin (commented):
Just realized I never posted my review comments here.

PR #535, opened 2026-03-17
unify integration banks depsit/withdraw instructions

IMEF-FEMI:
Summary                                                                                                                                      
         
  This PR consolidates all wrapped-integration user deposit and withdraw flows behind a single interface: integration_deposit and integration_withdraw. Instead of maintaining separate entrypoints for Kamino, Drift, Solend, and JupLend, marginfi now shares one common validation, accounting, rate-limit, health-check, and event pipeline, while keeping only the protocol-specific CPI logic isolated in per-integration handlers.                                                                                                                      
                                                                                                                                               
Motivation                                                                                                                                     
  
  The previous pattern duplicated the most security-sensitive parts of the flow across multiple integrations. Authorization, paused-state handling, deleverage accounting, rate-limiter enforcement, health validation, cache updates, and event emission were all copied per-protocol. Every change had to be replicated across nearly-identical handlers, raising the odds of inconsistent fixes and integration-specific gaps.     
                                                                              
Changes

  - Shared pipeline — Common pre-deposit validation, post-deposit finalization, and post-withdraw finalization (health check, rate limiting, deleverage tracking, event emission) now live in integration_common.rs and integration/mod.rs
  - Protocol handlers — Each integration (kamino_handler, drift_handler, solend_handler, juplend_handler) contains only its CPI construction, account validation, and balance computation                                                                                                    
  - Protocol accounts via remaining_accounts — The unified IntegrationDeposit/IntegrationWithdraw Anchor structs hold common accounts; protocol-specific accounts are passed and split by withdraw_protocol_account_count                                                             
  - TS test utilities — All instruction builders now target integrationDeposit/integrationWithdraw with protocol accounts packed as remaining accounts, validated at build time against INTEGRATION_PROTOCOL_ACCOUNT_COUNTS                                                                  
                                                                                                                                               
  Advantages                                                                                                                                     

  - Protocol-specific code is narrower and easier to audit
  - Future integrations only need a new handler — no new entrypoint or duplicated pipeline                                                       
  - Clients get one deposit path and one withdraw path with a clear separation between marginfi behavior and protocol account packing


programs/marginfi/src/instructions/integration/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+        ],
+        bump = bank.load()?.liquidity_vault_authority_bump
+    )]
+    pub liquidity_vault_authority: SystemAccount<'info>,
+
+    #[account(mut)]
+    pub liquidity_vault: InterfaceAccount<'info, TokenAccount>,
+
+    pub mint: Box<InterfaceAccount<'info, Mint>>,
+
```
jgur-psyops, line 530: Is it possible to pass integration accounts 1/2/3 here as regular accounts and keep using has_one with them? I think this would be easier on the caller side even if it's little odd you would have to pass pubkey default for some of the integrations.

programs/marginfi/src/instructions/integration/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    };
+    let signer_seeds: &[&[&[u8]]] =
+        bank_signer!(BankVaultType::Liquidity, bank_key, authority_bump);
+    let cpi_ctx = CpiContext::new_with_signer(program, cpi_accounts, signer_seeds);
+    transfer_checked(cpi_ctx, amount, accounts.mint.decimals)?;
+    Ok(())
+}
+
+#[derive(Accounts)]
+pub struct IntegrationDeposit<'info> {
```
jgur-psyops, line 447: At a high level I do like the unification but wonder if cutting down the number of instructions is actually beneficial: maybe one-instruction-per-venue is still preferable, but we keep this unified instruction and have the top-level one either call this by reference or CPI our own program (the latter is much easier because shared ix context is a pita, e.g. see the silly hacks in add_bank which is far more trivial). 

The main issue with CPI-own-program is the max call depth is 4 and e.g. Kamino already goes to depth 2 so I believe this maxes us out and makes it unusable by CPI, however for consumers that care about CPI depth, they have the option to use this base ix, so that's perfectly fine imo.

I lean to one-deposit/withdraw-ix-per-venue because (1) easier to index if each has a different ix, (2) easier to structure the accounts to utilize has_one and seeds to make it easier on the caller side, (3) lets us restrict individual integrations later, instead of all-or-nothing.

Either way we should probably add an "op_mode" arg here so you can tell which kind of deposit/withdraw it is by reading instruction args. 

cc @IliaZyrin @dawid-sw for thoughts

IMEF-FEMI: like you mentioned above 👆, i think we could maybe keep a unified low level function and have a the top-level instruction call it with an op_mode enum. that would mean we'd have no need for the unifying integration_account -> 1, 2, 3 you suggested [here](https://github.com/0dotxyz/marginfi-v2/pull/535#discussion_r2949094554)

and we would just keep the current non breaking structure (multiple deposit/withdrawal instructions)

dawid-sw: I agree, keeping per-venue entrypoints but unifying the internals is the way to go. We get the dedup without breaking callers and indexers

IliaZyrin: Also agree with the above!

programs/marginfi/src/instructions/drift/deposit.rs
```diff
@@ -184,19 +81,15 @@ pub struct DriftDeposit<'info> {
...
-    /// Owned by authority, the source account for the token deposit
     #[account(mut)]
     pub signer_token_account: InterfaceAccount<'info, TokenAccount>,
 
-    /// The Drift state account
-    /// CHECK: validated by the Drift program
+    /// CHECK: validated by Drift program
     pub drift_state: UncheckedAccount<'info>,
 
-    /// The Drift user account owned by liquidity_vault_authority
```
IliaZyrin, line 208: I would keep these comments, they are especially useful for the "integration" fields.

programs/marginfi/src/instructions/integration/mod.rs
IliaZyrin: Actually, after more thinking, I'm not not that sure such unification (even if only internal) is net beneficial. We clearly make it easier to change the common functionality of all deposits/withdrawals, e.g. checking for new limits/account states.

But on the other hand we get the logic of a previously-self-contained instruction distributed over several files which is harder to grasp/debug. The new integration implementation thus becomes not only a task of understanding the new venue internals + our common checks, **but also** fitting this into this new unified framework.

The unification itself is not perfect either: e.g. as mentioned [below](https://github.com/0dotxyz/marginfi-v2/pull/535/changes#r2949094554), we'll face trade offs like missing `has_one` checks VS using placeholder keys for the unused fields. In general, the more integrations we'll have, the more corner cases may appear which would require customizing the unified logic making it less elegant and harder to ponder about.

Last but not least the amount of lines only increased by this change, same goes for the files count.

jgur-psyops:
Per the conversation in March, we still want to do some unification of withdraw functionality (without losing our nice account inferences and has_one checks) that will dedupe logic between the various withdraw/deposit handlers, but this branch goes too far beyond that and is also very stale, so will close for now.

PR #537, opened 2026-03-22
docs: add onboarding FAQ for common marginfi anchor test issue and EMODE typo fix

sanjsol:
Adds a FAQ/Common issue entry for the "package_filter[@]: unbound variable" error in local testing. Clarifies root cause and practical setup steps to get onboarding runs unblocked.

Also fixed some typo errors on the fourth example on the EMODE section.


guides/DEVELOPERS_INTEGRATORS/GETTING_STARTED_DEV.md
```diff
@@ -65,6 +65,8 @@ number after the prefix determines their run order through the magic of filename
 anchor build --no-idl
 ./scripts/test-program.sh all
 ```
+Note: we recommend running `anchor clean` before using `anchor build --no-idl` if you ran
```
jgur-psyops, line 68: Worthwhile hint, but our goal is to never have to `anchor clean`, since it takes like several minutes to rebuild, so if you spot any of these you-have-to-anchor-clean-or-it's-always broken things try to flag it for later. We are forever hammering these out between different system builds. I think I am down to only a handful of anchor cleans a week, woohoo!

PR #541, opened 2026-03-25
Fix double counting bug

IMEF-FEMI:
# PR Summary
Fix double-counting of interest/exchange rate in group rate limiter for Drift and JupLend withdrawals

## Why
Both integrations embed an exchange rate into the oracle price. The withdrawal amount was also exchange-rate-adjusted, causing calc_value in the `record_withdrawal_outflow` function to apply the factor twice  inflating the USD value used for group rate limit checks.

### DIFF
--- a/type-crate/Cargo.toml
+++ b/type-crate/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "marginfi-type-crate"
-version = "0.1.9"
+version = "0.1.8"
 edition = "2021"
 publish = true
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3116,7 +3116,7 @@ dependencies = [
 
 [[package]]
 name = "marginfi-type-crate"
-version = "0.1.9"
+version = "0.1.8"
 dependencies = [
  "anchor-lang",
  "bs58",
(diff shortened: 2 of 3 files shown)


type-crate/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "marginfi-type-crate"
-version = "0.1.8"
+version = "0.1.9"
```
IliaZyrin: I assume this is not needed given that you target 0.1.8-main branch.

programs/marginfi/src/utils/general.rs
```diff
@@ -505,9 +505,9 @@ pub fn record_withdrawal_outflow(
             check!(price > I80F48::ZERO, MarginfiError::InvalidRateLimitPrice);
 
             let value = calc_value(
-                I80F48::from_num(token_amount),
+                I80F48::from_num(balance_amount),
```
jgur-psyops, line 508: here group records the balance_amount vs native, which means in the borrow flow the bank records `amount` but the group records `amount_pre_fee`. That's inconsistent with the deposit side, which records the pre-fee amount for both group/bank.

This doesn't really matter since none of the banks in prod have a transfer fee anyways, but it probably wasn't intentional.

This also reminds we should probably sunset T22 fee support since literally nobody has ever used it and it's a waste of dev time!

type-crate/Cargo.toml
IMEF-FEMI: Thats correct... fixing now

PR #542, opened 2026-03-26
minor: remove bank metadata is-initialized requirement

IMEF-FEMI:
## Summary                                                                                                                                                         
                                         
Relaxed bank account validation  in InitBankMetadata and WriteBankMetadata instructions, enabling metadata accounts to  be pre-created before the bank is initialized

### DIFF
--- a/tests/04_configureBank.spec.ts
+++ b/tests/04_configureBank.spec.ts
@@ -12,6 +12,7 @@ import { Marginfi } from "../target/types/marginfi";
 import {
   bankKeypairUsdc,
   bankrunProgram,
+  ecosystem,
   groupAdmin,
   marginfiGroup,
   oracles,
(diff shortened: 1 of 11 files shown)


tests/04_configureBank.spec.ts
```diff
@@ -341,36 +342,92 @@ describe("Lending pool configure bank", () => {
...
       program.programId,
-      bankKeypairUsdc.publicKey
+      bankKeypairUsdc.publicKey,
     );
     const meta = await program.account.bankMetadata.fetch(metaKey);
     assertKeysEqual(meta.bank, bankKeypairUsdc.publicKey);
     assert.equal(meta.bump, metaBump);
   });
 
+  it("(attacker) create metadata for a bank that does not exist yet - should fail", async () => {
```
jgur-psyops: Added this test to show the current approach allows an attacker to create their own group and squat on the metadata for another group. 

I think to pull off this feature without enabling this attack, the writeBankMetadata ix must take the bank's seeds as arguments, then it can determine the bank is a valid PDA for this group even if it doesn't yet exist. 

You'd end up with something like
```
    #[account(
        seeds = [
            marginfi_group.key().as_ref(),
            bank_mint.key().as_ref(),
            &bank_seed.to_le_bytes(),
        ],
        bump,
    )]
    pub bank: UncheckedAccount<'info>,
```

PR #543, opened 2026-03-27
Kamino Refresh Reserves Batch

jgur-psyops:
Thank you to @abelmarnk for the contribution in #528 and #527

Changes since #528:

* Minimal IDL now includes batch refresh reserves. 
* Batch refresh now support in liquidation
* Kamino deposit/withdraw can now optionally refresh (see #527 for performance implications and benchmarks when refreshing internally vs outside the ix)

### DIFF
--- a/clients/rust/marginfi-cli/src/processor/integrations.rs
+++ b/clients/rust/marginfi-cli/src/processor/integrations.rs
@@ -621,11 +621,7 @@ pub fn kamino_withdraw(
             instruction_sysvar_account: sysvar::instructions::ID,
         }
         .to_account_metas(Some(true)),
-        data: marginfi::instruction::KaminoWithdraw {
-            amount,
-            flags,
-        }
-        .data(),
+        data: marginfi::instruction::KaminoWithdraw { amount, flags }.data(),
     };
     ix.accounts.extend(observation_metas);
 
--- a/test-utils/src/marginfi_account.rs
+++ b/test-utils/src/marginfi_account.rs
@@ -1579,11 +1579,7 @@ impl MarginfiAccountFixture {
                 instruction_sysvar_account: sysvar::instructions::ID,
             }
             .to_account_metas(Some(true)),
-            data: marginfi::instruction::KaminoWithdraw {
-                amount,
-                flags,
-            }
-            .data(),
+            data: marginfi::instruction::KaminoWithdraw { amount, flags }.data(),
         };
 
         self.append_integration_withdraw_health_accounts(&mut ix)


programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -76,9 +79,14 @@ use marginfi_type_crate::{
 pub fn kamino_withdraw<'info>(
     ctx: Context<'_, '_, 'info, 'info, KaminoWithdraw<'info>>,
     amount: u64, // Collateral token amount
-    withdraw_all: Option<bool>,
+    flags: Option<u8>,
```
jgur-psyops, line 82: I wrote this both ways, as 
```
pub fn kamino_withdraw<'info>(
    ctx: Context<'_, '_, 'info, 'info, KaminoWithdraw<'info>>,
    amount: u64, // Collateral token amount
    flags: Option<u8>,
``` 
and as
```
pub fn kamino_withdraw<'info>(
    ctx: Context<'_, '_, 'info, 'info, KaminoWithdraw<'info>>,
    amount: u64, // Collateral token amount
    withdraw_all: Option<bool>,
    refresh_reserves: Option<bool>,
```

both suck, but at least the former is backwards-compatible, same saves 2 bytes of ix space (maybe, or it just pads to 8 both ways). Although deposit is breaking regardless, so I guess it's a wash. Vote for which one you think is less-bad.

IliaZyrin: Since we're changing it anyways, can we consider removing `Option<>` at all? Never liked it.
In any case I prefer the combined "flags" version.

programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -76,9 +79,14 @@ use marginfi_type_crate::{
...
 ) -> MarginfiResult {
-    let withdraw_all = withdraw_all.unwrap_or(false);
+    const WITHDRAW_ALL_FLAG: u8 = 1 << 0;
```
IliaZyrin, line 84: If we decide to proceed with the flags way, it makes sense to expose these consts from the type crate.

IliaZyrin (approved):
LGTM

programs/marginfi/src/instructions/kamino/withdraw.rs
dawid-sw: I'm voting for flags as well, but we need to make it good devex on SDK side

PR #550, opened 2026-04-02
Allow deleverage and handle_bankruptcy during protocol pause

dawid-sw:
## Summary

  - Withdraw, repay, and integration withdraws (Drift, Kamino, JupLend, Solend) now bypass the
    protocol pause check when the account is in deleverage (`ACCOUNT_IN_DELEVERAGE` flag).
    Permissionless liquidations remain blocked during pause.
  - `handle_bankruptcy` bypasses the protocol pause check when called by admin or risk_admin.
    Permissionless callers are still blocked during pause.

  ## Test plan

  - [x] `deleverage_succeeds_during_pause`
  - [x] `normal_withdraw_still_blocked_during_pause`
  - [x] `normal_repay_still_blocked_during_pause`
  - [x] `liquidation_still_blocked_during_pause`
  - [x] `handle_bankruptcy_by_admin_succeeds_during_pause`
  - [x] `handle_bankruptcy_by_non_admin_fails_during_pause`
  - [x] `handle_bankruptcy_without_pause_permissionless_still_works`
  - [x] `handle_bankruptcy_non_admin_succeeds_after_unpause`


programs/marginfi/tests/admin_actions/actions_during_pause.rs
IliaZyrin, line 1: I would rename this to `actions_during_pause` since it's not only deleverage.

dawid-sw: good point

programs/marginfi/src/instructions/marginfi_group/handle_bankruptcy.rs
```diff
@@ -49,23 +49,28 @@ pub fn lending_pool_handle_bankruptcy<'info>(
...
+        let is_admin_or_risk_admin = signer == group.risk_admin || signer == group.admin;
+        let permissionless_bad_debt_settlement =
+            bank.get_flag(PERMISSIONLESS_BAD_DEBT_SETTLEMENT_FLAG);
+
+        if permissionless_bad_debt_settlement {
+            // if permissionless, users can bankrupt reduce-only or operational banks
+            validate_bank_state(&bank, InstructionKind::FailsInPausedState)?;
+        } else {
+            // admin can bankrupt banks in any state
+            validate_bank_state(&bank, InstructionKind::Unrestricted)?;
```
jgur-psyops, line 65: It occurs to me that at some point we may want to switch Drift banks to Paused instead of ReduceOnly, I added support for that here, I don't think we need to do it anywhere else but something to bear in mind.

dawid-sw: is this validate state needed? if we like to do it in any state, prob we just ignore this check, its dead code now

PR #552, opened 2026-04-02
Admin cross bank socialize

IMEF-FEMI:
Added a new admin_cross_bank_socialize instruction for risk admins to move a specified loss out of a bank and socialize it across that  bank’s depositors with option to use banks insurance.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/admin_cross_bank_socialize.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/admin_cross_bank_socialize.rs
@@ -129,30 +129,35 @@ pub fn lending_pool_admin_cross_bank_socialize<'info>(
         I80F48::ZERO
     };
 
-    let socialized_loss = max(effective_loss - covered_by_insurance, I80F48::ZERO);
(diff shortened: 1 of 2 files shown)


programs/marginfi/src/instructions/marginfi_group/admin_cross_bank_socialize.rs
```diff
@@ -0,0 +1,246 @@
...
+/// bank's vault. When `use_insurance` is true, draws from insurance fund first, then socializes
+/// the remainder. Used to spread bad debt from a compromised integration across multiple banks.
+/// No pause checks.
+pub fn lending_pool_admin_cross_bank_socialize<'info>(
+    mut ctx: Context<'_, '_, 'info, 'info, AdminCrossBankSocialize<'info>>,
+    amount: u64,
+    use_insurance: bool,
+) -> MarginfiResult {
+    check!(amount > 0, MarginfiError::InvalidTransfer);
+
```
jgur-psyops, line 35: in hindsight we both forgot one of the most important parts: msg! the amount for signers! I will try to remember next time!

PR #553, opened 2026-04-02
Super admin withdraw ix

jgur-psyops:
Very dangerous ix that allows the group admin to arbitrarily withdraw/deposit directly from the LP and shares. We intend to use this to evenly distribute Drift-hack-related impacts.

### DIFF
--- a/programs/marginfi/tests/admin_actions/super_deposit_withdraw.rs
+++ b/programs/marginfi/tests/admin_actions/super_deposit_withdraw.rs
@@ -1,10 +1,12 @@
 use anchor_lang::prelude::Clock;
+use anchor_spl::token::spl_token;
 use fixed::types::I80F48;
 use fixed_macro::types::I80F48;
 use fixtures::time;
 use fixtures::{assert_eq_noise, prelude::*, ui_to_native};
 use marginfi::state::bank::BankImpl;
 use solana_program_test::*;
+use solana_sdk::program_pack::Pack;
 
 #[tokio::test]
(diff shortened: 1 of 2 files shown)


programs/marginfi/src/instructions/marginfi_group/super_admin_withdraw.rs
```diff
@@ -0,0 +1,165 @@
...
+    let total_asset_shares: I80F48 = bank.total_asset_shares.into();
+    check!(
+        total_asset_shares > ZERO_AMOUNT_THRESHOLD,
+        MarginfiError::NoAssetFound
+    );
+
+    let assets_before = bank.get_asset_amount(total_asset_shares)?;
+    let withdrawal_amount = I80F48::from_num(amount);
+    check!(
+        assets_before >= withdrawal_amount,
```
IMEF-FEMI, line 81: This mainly checks the accounting value (shares × share_value), which includes tokens lent out to borrowers...  If utilization is high, the vault may hold fewer tokens than the inputed `amount`. we should probably consider checking the actual vault balance via `accessor::amount(&liquidity_vault)` and returning an error if  `amount > accessor::amount(&liquidity_vault)`

jgur-psyops: Yeah definitely a valid point. Luckily I think it will just fail at the transfer CPI in this case so not security concern I think just awkward for the simulation.

Does that affect any prod banks for the amounts we're withdrawing? We're only doing <10% and I don't think any banks have 90% utilization so I think we're good.

PR #555, opened 2026-04-04
Add same-asset-emode

abelmarnk:
### Summary

Adds a same-asset emode feature that automatically grants a boosted collateral weight to positions whose collateral mint exactly matches the single liability mint(across all liabilities) active in the account.

The boost magnitude is derived from a group-level leverage parameter using the formula:

$$
w_{\text{asset}} = w_{\text{liab}} \cdot \left(1 - \frac{1}{\text{leverage}}\right)
$$

The core health engine functions now require the account's group in order to read the same-asset leverage configuration. Accordingly, all instruction handlers that perform health checks pass the loaded group down into the engine.

The emode reconciliation pass now produces a `ReconciledEmodeConfig` which holds only the data actively needed during the health calculation rather than carrying the full `EmodeConfig` through.

---

### Breaking changes

* The `marginfi_group_configure` instruction now uses two new optional parameters:

  * `same_asset_emode_init_leverage`
  * `same_asset_emode_maint_leverage`

* Some instructions now require the group to be passed in, in order to assess the same asset config:

  * `lending_account_end_flashloan`
  * `lending_account_liquidate_start`
  * `lending_account_liquidate_end`
  * `lending_account_pulse_health`

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/configure.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/configure.rs
@@ -3,10 +3,13 @@ use crate::state::marginfi_group::MarginfiGroupImpl;
 use crate::{MarginfiError, MarginfiResult};
 use anchor_lang::prelude::*;
 use fixed::types::I80F48;
-use marginfi_type_crate::types::{basis_to_u32, MarginfiGroup, WrappedI80F48};
+use marginfi_type_crate::types::{
+    basis_to_u32, same_asset_leverage_to_u32, u32_to_same_asset_leverage, MarginfiGroup,
+    WrappedI80F48,
+};
 
 /// Validate and apply an optional emode leverage value. If `Some`, validates it is in [1, 100] and
-/// writes the basis-point encoding. If `None`, requires the current on-chain value is non-zero.
+/// writes the basis-point encoding. If `None`, leaves the value as is.
 fn validate_and_apply_emode_leverage(
     new_value: Option<WrappedI80F48>,
     current: &mut u32,
@@ -22,6 +25,28 @@ fn validate_and_apply_emode_leverage(
             return Err(MarginfiError::BadEmodeConfig.into());
         }
         *current = basis_to_u32(leverage);
+    }
+    Ok(())
+}
+
+/// Validate and apply an optional same-asset emode leverage value.
+/// If `Some`, validates it is in [1, 1000] and writes the encoded `u32` value.
+/// If `None`, leaves the stored value as is.
+fn validate_and_apply_same_asset_emode_leverage(
+    new_value: Option<WrappedI80F48>,
+    current: &mut u32,
+) -> MarginfiResult {
+    if let Some(wrapped) = new_value {
+        let leverage: I80F48 = wrapped.into();
+        if leverage < I80F48::ONE {
+            msg!("same-asset emode leverage {} must be >= 1", leverage);
+            return Err(MarginfiError::BadEmodeConfig.into());
+        }
+        if leverage > I80F48::from_num(1000) {
+            msg!("same-asset emode leverage {} must be <= 1000", leverage);
+            return Err(MarginfiError::BadEmodeConfig.into());
+        }
+        *current = same_asset_leverage_to_u32(leverage);
     }
     Ok(())
 }
@@ -44,6 +69,8 @@ pub fn configure(
     new_risk_admin: Option<Pubkey>,
     emode_max_init_leverage: Option<WrappedI80F48>,
     emode_max_maint_leverage: Option<WrappedI80F48>,
+    same_asset_emode_init_leverage: Option<WrappedI80F48>,
+    same_asset_emode_maint_leverage: Option<WrappedI80F48>,
 ) -> MarginfiResult {
     let marginfi_group = &mut ctx.accounts.marginfi_group.load_mut()?;
     if let Some(new_admin) = new_admin {
@@ -86,6 +113,32 @@ pub fn configure(
             "emode init leverage ({}) must be < maint leverage ({})",
             marginfi_group.emode_max_init_leverage,
             marginfi_group.emode_max_maint_leverage
+        );
+        return Err(MarginfiError::BadEmodeConfig.into());
+    }
+
+    validate_and_apply_same_asset_emode_leverage(
+        same_asset_emode_init_leverage,
+        &mut marginfi_group.same_asset_emode_init_leverage,
+    )?;
+    validate_and_apply_same_asset_emode_leverage(
+        same_asset_emode_maint_leverage,
+        &mut marginfi_group.same_asset_emode_maint_leverage,
+    )?;
+
+    let same_asset_init_leverage =
+        u32_to_same_asset_leverage(marginfi_group.same_asset_emode_init_leverage);
+    let same_asset_maint_leverage =
+        u32_to_same_asset_leverage(marginfi_group.same_asset_emode_maint_leverage);
+
+    let same_asset_disabled =
+        same_asset_init_leverage <= I80F48::ONE && same_asset_maint_leverage <= I80F48::ONE;
+
+    if !same_asset_disabled && same_asset_init_leverage >= same_asset_maint_leverage {
+        msg!(
+            "same-asset emode init leverage ({}) must be < maint leverage ({})",
+            same_asset_init_leverage,
+            same_asset_maint_leverage
         );
         return Err(MarginfiError::BadEmodeConfig.into());
     }
--- a/programs/marginfi/src/instructions/marginfi_group/configure.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/configure.rs
@@ -3,10 +3,7 @@ use crate::state::marginfi_group::MarginfiGroupImpl;
 use crate::{MarginfiError, MarginfiResult};
 use anchor_lang::prelude::*;
 use fixed::types::I80F48;
-use marginfi_type_crate::types::{
-    basis_to_u32, same_asset_leverage_to_u32, u32_to_same_asset_leverage, MarginfiGroup,
-    WrappedI80F48,
-};
+use marginfi_type_crate::types::{basis_to_u32, u32_to_basis, MarginfiGroup, WrappedI80F48};
 
 /// Validate and apply an optional emode leverage value. If `Some`, validates it is in [1, 100] and
 /// writes the basis-point encoding. If `None`, leaves the value as is.
@@ -30,7 +27,7 @@ fn validate_and_apply_emode_leverage(
 }
 
 /// Validate and apply an optional same-asset emode leverage value.
-/// If `Some`, validates it is in [1, 1000] and writes the encoded `u32` value.
+/// If `Some`, validates it is in [1, 100] and writes the encoded `u32` value.
 /// If `None`, leaves the stored value as is.
 fn validate_and_apply_same_asset_emode_leverage(
     new_value: Option<WrappedI80F48>,
@@ -42,11 +39,11 @@ fn validate_and_apply_same_asset_emode_leverage(
             msg!("same-asset emode leverage {} must be >= 1", leverage);
             return Err(MarginfiError::BadEmodeConfig.into());
         }
-        if leverage > I80F48::from_num(1000) {
-            msg!("same-asset emode leverage {} must be <= 1000", leverage);
+        if leverage > I80F48::from_num(100) {
+            msg!("same-asset emode leverage {} must be <= 100", leverage);
             return Err(MarginfiError::BadEmodeConfig.into());
         }
-        *current = same_asset_leverage_to_u32(leverage);
+        *current = basis_to_u32(leverage);
     }
     Ok(())
 }
@@ -126,10 +123,8 @@ pub fn configure(
         &mut marginfi_group.same_asset_emode_maint_leverage,
     )?;
 
-    let same_asset_init_leverage =
-        u32_to_same_asset_leverage(marginfi_group.same_asset_emode_init_leverage);
-    let same_asset_maint_leverage =
-        u32_to_same_asset_leverage(marginfi_group.same_asset_emode_maint_leverage);
+    let same_asset_init_leverage = u32_to_basis(marginfi_group.same_asset_emode_init_leverage);
+    let same_asset_maint_leverage = u32_to_basis(marginfi_group.same_asset_emode_maint_leverage);
 
     let same_asset_disabled =
         same_asset_init_leverage <= I80F48::ONE && same_asset_maint_leverage <= I80F48::ONE;
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -15,13 +15,13 @@ use marginfi_type_crate::{
         MAX_INTEGRATION_POSITIONS, ORDER_ACTIVE_TAGS, ZERO_AMOUNT_THRESHOLD,
     },
     types::{
-        compute_same_asset_emode_weight, reconcile_emode_configs, u32_to_same_asset_leverage,
+        compute_same_asset_emode_weight, reconcile_emode_configs, u32_to_basis,
         Balance, BalanceSide, Bank, BankOperationalState, EmodeConfig, HealthCache, HealthPriceMode,
         LendingAccount, LiquidationPriceCache, MarginfiAccount, MarginfiGroup, OraclePriceType,
         OraclePriceWithConfidence, OracleSetup, PriceBias, ReconciledEmodeConfig,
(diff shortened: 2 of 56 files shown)


abelmarnk:
The majority of the changes are from testing with P0 and different venues for both rust and typescript, the program implementation changes are minimal.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -642,6 +692,137 @@ impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
...
+
+    let mut account_index = 0usize;
+    let mut shared_mint: Option<Pubkey> = None;
+    let mut weakest_liab_weight: Option<I80F48> = None;
+    for balance in lending_account.balances.iter() {
+        if !balance.is_active() {
+            continue;
+        }
+
+        let heap_checkpoint = heap_pos();
```
abelmarnk, line 777: Will remove this, when making the other review changes.

type-crate/src/types/emode.rs
```diff
@@ -179,74 +242,67 @@ impl EmodeEntry {
...
+        merged_entries.insert(
+            entry.collateral_bank_emode_tag,
+            (projected_asset_weight(entry, requirement_type), 1),
+        );
+    }
+
+    for cfg in iter {
+        num_configs += 1;
+        for entry in cfg.entries.iter().filter(|entry| !entry.is_empty()) {
             let tag = entry.collateral_bank_emode_tag;
```
IMEF-FEMI, line 274: the `flags` field seems to not have been merged here.  I think we should have it merged too while we maintain the "least favourable" rule used prior

Also, `ReconciledEmodeEntry` now only carries collateral_bank_emode_tag and asset_weight. This is not an issue today since no e-mode flag is consumed by the risk engine yet, but it  would weaken future compatibility. I think we should fix this

type-crate/src/types/emode.rs
```diff
@@ -308,3 +364,265 @@ pub fn reconcile_emode_configs_classic(configs: Vec<EmodeConfig>) -> EmodeConfig
...
+    let factor = I80F48::ONE - inverse;
+    liability_weight.checked_mul(factor).unwrap_or(I80F48::ZERO)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use fixed::types::I80F48;
+
+    macro_rules! assert_eq_with_tolerance {
```
IMEF-FEMI, line 390: There is already an assert_eq_with_tolerance function [here](https://github.com/abelmarnk/marginfi-v2/blob/a4ccbea972bd19437db2b52720a773c426c291e3/programs/marginfi/src/utils/general.rs#L449)

Lets use that instead

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -642,6 +692,137 @@ impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
...
+                let liab_weight = bank
+                    .config
+                    .get_weight(requirement_type, BalanceSide::Liabilities);
+                if !update_reconciled_same_asset_config(
+                    &mut shared_mint,
+                    &mut weakest_liab_weight,
+                    bank.mint,
+                    liab_weight,
+                ) {
+                    return Ok(());
```
IMEF-FEMI, line 807: This early return skips the heap_restore(heap_checkpoint) [here](https://github.com/abelmarnk/marginfi-v2/blob/a4ccbea972bd19437db2b52720a773c426c291e3/programs/marginfi/src/state/marginfi_account.rs). If your planned [update](https://github.com/0dotxyz/marginfi-v2/pull/555#discussion_r3050700962) removes the checkpoint entirely, ignore this... otherwise please restore before returning. The path would fire on every health check where liabilities spans more than one mint, so the memory leak isn't rare.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -26,6 +29,28 @@ fn validate_and_apply_emode_leverage(
...
+    new_value: Option<WrappedI80F48>,
+    current: &mut u32,
+) -> MarginfiResult {
+    if let Some(wrapped) = new_value {
+        let leverage: I80F48 = wrapped.into();
+        if leverage < I80F48::ONE {
+            msg!("same-asset emode leverage {} must be >= 1", leverage);
+            return Err(MarginfiError::BadEmodeConfig.into());
+        }
+        if leverage > I80F48::from_num(1000) {
```
IMEF-FEMI, line 45: I would suggest lowering the max leverage from 1000 to 100, matching the current emode cap. At L=1000 the safety margin collapses to 0.1% which, for example, is smaller than a realistic oracle confidence spread (even if we use the same oracle for both banks -- as the bias would be different for asset and liab).

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -26,6 +29,28 @@ fn validate_and_apply_emode_leverage(
...
+            msg!("same-asset emode leverage {} must be <= 1000", leverage);
+            return Err(MarginfiError::BadEmodeConfig.into());
+        }
+        *current = same_asset_leverage_to_u32(leverage);
```
IMEF-FEMI, line 49: once we cap leverage at 100, we can take advantage of the `basis_to_u32` / `u32_to_basis` helper functions and clean up the `same_asset_leverage_to_u32` / `u32_to_same_asset_leverage` helpers

IMEF-FEMI (commented):
I think we should also add a couple more test cases to cover a few edge scenarios:
                                                                                                           
* survivability test on a same-mint position for small bad-debt haircut events on P0 / integration banks (10-50 bp). These are normal operational incidents, not attack scenarios, and can happen when liquidators fail to clear a position before it goes slightly underwater.
* oracle confidence spread test. Even with a shared oracle, opposite-side confidence bias creates an inherent spread; that makes L=1000 fragile, while L=100 still leaves reasonable headroom.

type-crate/src/types/emode.rs
abelmarnk: the deps would be circular, it's from the program crate. I can inline it if that would be better.

type-crate/src/types/emode.rs
abelmarnk: yeah, i left out the flags since it wasn't being used in the health calculations, i can add it back.
the "least favorable" rule is still being maintained, i removed some parts of the loop that weren't necessary, i.e since we are taking the AND of all the entries, then we only ever need to track the first one and we can ignore the stuff that comes and doesn't match.
more generally, i removed stuff that wasn't necessary since i was adding more code to health calc which is probably the hottest path, to offset the difference.

programs/marginfi/src/state/marginfi_account.rs
abelmarnk: yeah, i planned on removing that, there is no heap allocation on that path so it's a no-op.

abelmarnk:
> I think we should also add a couple more test cases to cover a few edge scenarios:
> 
> * survivability test on a same-mint position for small bad-debt haircut events on P0 / integration banks (10-50 bp). These are normal operational incidents, not attack scenarios, and can happen when liquidators fail to clear a position before it goes slightly underwater.
> * oracle confidence spread test. Even with a shared oracle, opposite-side confidence bias creates an inherent spread; that makes L=1000 fragile, while L=100 still leaves reasonable headroom.

Appreciate the feedback, will make the changes!

abelmarnk:
also are the tests redundant(i.e the extra liquidation & receivership tests)?, i wanted to remove some since it makes the change unnecessarily large.

programs/marginfi/src/instructions/marginfi_group/configure.rs
abelmarnk: yep that makes sense, thanks for flagging this, 😅 i previously considered upping it even higher.
resolved [in](https://github.com/0dotxyz/marginfi-v2/pull/555/changes/8b59a37251ffd05e1e2c6ca7dbd68911c7be45c4#diff-56254f61124ab5bf7158d05e4cdcdd823b23183f616316419a95900a0dff4055R42).

programs/marginfi/src/instructions/marginfi_group/configure.rs
abelmarnk: resolved [here](https://github.com/0dotxyz/marginfi-v2/pull/555/changes/8b59a37251ffd05e1e2c6ca7dbd68911c7be45c4#diff-56254f61124ab5bf7158d05e4cdcdd823b23183f616316419a95900a0dff4055R46).

programs/marginfi/src/state/marginfi_account.rs
abelmarnk: resolved [here](https://github.com/0dotxyz/marginfi-v2/pull/555/changes/8b59a37251ffd05e1e2c6ca7dbd68911c7be45c4#diff-72cc8cf3980f03d6e102e5beef1db9a576e08f5c27c9656a742a5c2ea8bc585bL777).

type-crate/src/types/emode.rs
abelmarnk: resolved [here](https://github.com/0dotxyz/marginfi-v2/pull/555/changes/8b59a37251ffd05e1e2c6ca7dbd68911c7be45c4#diff-ac781c9c29635e847653e07b7a41934ab49e55fc7ce5c83c59ebe27823dfda02L390).

type-crate/src/types/emode.rs
abelmarnk: added the flags [here](https://github.com/0dotxyz/marginfi-v2/pull/555/changes/8b59a37251ffd05e1e2c6ca7dbd68911c7be45c4#diff-ac781c9c29635e847653e07b7a41934ab49e55fc7ce5c83c59ebe27823dfda02R155).

abelmarnk:
For clarity these are the assumptions i made for the review fixes:-
- I made the assumption that what you mean't by survivability is that in those bad debt scenarios we'd want for the ~50bps drop not to affect other borrowers and take their accounts below the margin in equity terms.
- For the oracle spread check i assumed you meant that it should weigh assets and liabs differently though they are the same mint.

I can make further changes if they are not correct.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -26,6 +26,28 @@ fn validate_and_apply_emode_leverage(
     Ok(())
 }
 
+/// Validate and apply an optional same-asset emode leverage value.
+/// If `Some`, validates it is in [1, 100] and writes the encoded `u32` value.
+/// If `None`, leaves the stored value as is.
+fn validate_and_apply_same_asset_emode_leverage(
```
IliaZyrin, line 32: I would not introduce a new function here. Not exact error message is not that important, because the ix will anyway be called only by the group admin.

On a more general note, I don't think we really need separate limits for same-asset emode VS just using the existing max ones. Cannot imagine a good use case where we would want to only tweak one of those.

cc @jgur-psyops @IMEF-FEMI @dawid-sw

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -492,6 +512,31 @@ pub fn calc_amount(value: I80F48, price: I80F48, mint_decimals: u8) -> MarginfiR
...
+// -----------------------------------------------------------------------------
+// Same-Asset Automatic Emode
+// -----------------------------------------------------------------------------
+
+/// Configuration for same-asset automatic emode, extracted from the group.
+/// When all liabilities in an account share the same mint, this leverage is
+/// automatically applied to matching collateral, independent of the regular
+/// emode tag system.
+#[derive(Copy, Clone, Debug, Default)]
+pub struct SameAssetEmodeConfig {
```
IliaZyrin, line 524: This struct seems redundant. Can we just pass the group where necessary?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -565,6 +610,133 @@ impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
...
+
+/// Updates the running same-asset reconciliation state for one liability mint/weight pair.
+///
+/// This helper assumes the caller is iterating over all active liabilities for a single
+/// reconciliation pass and is reusing the same `shared_mint` and `weakest_liab_weight`
+/// accumulators across calls. Ideally it should have been defined as an inner function
+/// where it is used, but is left outside since it is also used in the tests.
+/// In that calling pattern it:
+/// - records the first liability mint as the candidate shared mint
+/// - keeps the least favorable liability weight seen so far for that shared mint
```
IliaZyrin, line 634: This is a bit confusing since **the least liability** is actually **the most favourable** (you want to owe less, not more, right?=)). Same goes for the `weakest_liab_weight`, i.e. better rename it to `lowest_liab_weight`.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -565,6 +610,133 @@ impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
...
+            *weakest_liab_weight = Some(liab_weight);
+            true
+        }
+    }
+}
+
+/// Populates the nested same-asset auto-emode configuration on the reconciled runtime config.
+///
+/// Same-asset auto-emode only activates when all active liabilities share one mint. When enabled,
+/// the selected asset weight is derived from the least favorable liability-side weight across those
```
IliaZyrin, line 668:
```suggestion
/// the selected asset weight is derived from the lowest liability-side weight across those
```

type-crate/src/types/emode.rs
```diff
@@ -140,36 +142,100 @@ impl EmodeEntry {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ReconciledEmodeRequirementType {
```
IliaZyrin, line 146: This seems unnecessary.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -565,6 +610,133 @@ impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
...
+/// liabilities for the active requirement type
+fn populate_reconciled_same_asset_config<'info>(
```
IliaZyrin, line 670: I think we should merge this one with `EmodeConfigIterator` to not iterate twice over the same active balances. E.g. something like this:

```
struct EmodeConfigIterator<'a, 'info> {
    lending_account: &'a LendingAccount,
    remaining_ais: &'info [AccountInfo<'info>],
    balance_index: usize,
    account_index: usize,
    banks_only: bool,
    shared_mint: Option<Pubkey>,
    lowest_liab_weight: Option<I80F48>,
    has_different_liab_mints: bool,
}

impl<'a, 'info> EmodeConfigIterator<'a, 'info> {
    fn new(
        lending_account: &'a LendingAccount,
        remaining_ais: &'info [AccountInfo<'info>],
        banks_only: bool,
    ) -> Self {
        Self {
            lending_account,
            remaining_ais,
            balance_index: 0,
            account_index: 0,
            banks_only,
            shared_mint: None,
            lowest_liab_weight: None,
            has_different_liab_mints: false,
        }
    }
}

impl<'a, 'info> Iterator for EmodeConfigIterator<'a, 'info> {
    type Item = EmodeConfig;

    fn next(&mut self) -> Option<Self::Item> {
        // Find next active balance with liabilities
        while self.balance_index < self.lending_account.balances.len() {
            let balance = &self.lending_account.balances[self.balance_index];

            if !balance.is_active() {
                self.balance_index += 1;
                continue;
            }

            // Try to load bank to get account count and emode config
            let bank_ai = self.remaining_ais.get(self.account_index)?;
            let bank_al = AccountLoader::<Bank>::try_from(bank_ai).ok()?;
            let bank = bank_al.load().ok()?;

            if balance.bank_pk != *bank_ai.key {
                return None;
            }

            let num_accounts = if self.banks_only {
                1
            } else {
                get_remaining_accounts_per_bank(&bank).ok()?
            };

            // Advance indices
            self.account_index += num_accounts;
            self.balance_index += 1;

            if !balance.is_empty(BalanceSide::Liabilities) {
                let liab_weight = bank
                    .config
                    .get_weight(requirement_type, BalanceSide::Liabilities);
                if !self.has_different_liab_mints && !self.update_reconciled_same_asset_config(
                    bank.mint,
                    liab_weight,
                ) {
                    self.has_different_liab_mints = true;
                }

                // Only yield emode config if this balance has liabilities
                return Some(bank.emode.emode_config);
            }
        }
        None
    }
}
```

And then we could populate the reconciled config's fields in some new function of Iterator OR in the updated `reconcile_emode_configs()` as you see fit:
```
if let (Some(mint), Some(liab_weight)) = (shared_mint, weakest_liab_weight) {
        reconciled_emode_config.same_asset.mint = mint;
        reconciled_emode_config.same_asset.asset_weight =
            compute_same_asset_emode_weight(leverage, liab_weight);
    }
```

PR #557, opened 2026-04-08
Remove super admin instructions

IMEF-FEMI:
# Summary

                                                                                                                           
This PR removes the emergency admin instructions and reverts protocol changes that were added in the wake of the Drift incident


dawid-sw:
q if we like to deploy it now, or in 0.1.9 upgrade

programs/marginfi/src/instructions/marginfi_group/super_admin_deposit.rs
```diff
@@ -1,124 +0,0 @@
...
-    types::{Bank, MarginfiGroup},
-};
-
-/// Group admin only.
-/// Transfers tokens from `admin_token_account` into the bank liquidity vault and raises
-/// `asset_share_value` so that existing depositor shares are increased proportionally.
-///
-/// Note: Token-2022 transfer-fee extensions are not handled here. This instruction assumes
-/// the vault receives exactly `amount`.
-pub fn super_admin_deposit<'info>(
```
jgur-psyops, line 23: Let's add `if mainnet && not bpf-test! {panic}` to these and keep them in source in case we need to re-enable, similar to clone_bank these might be helpful to edit staging.

type-crate/src/types/panic_state_cache.rs
```diff
@@ -77,8 +77,8 @@ pub struct PanicState {
 
 impl PanicState {
     pub const FLAG_PAUSED: u8 = 1 << 0;
-    pub const PAUSE_DURATION_SECONDS: i64 = 7 * 24 * 60 * 60; // 7 days
-    pub const MAX_CONSECUTIVE_PAUSES: u8 = 10;
+    pub const PAUSE_DURATION_SECONDS: i64 = 24 * 60 * 60; // 24 hours
```
jgur-psyops: I think the way the system works you have so many max consecutive pauses PER DAY, so if pauses * pause time > 24h you can pause forever. Is that right? If so we should probably adjust it such that it's not unlimited.

programs/marginfi/tests/admin_actions/actions_during_pause.rs
```diff
@@ -1,774 +0,0 @@
...
-};
-use solana_program_test::*;
-use solana_sdk::pubkey::Pubkey;
-use solana_sdk::signature::Keypair;
-use solana_sdk::signer::Signer;
-use solana_sdk::transaction::Transaction;
-
-/// Deleverage with withdraw + repay succeeds while protocol is paused.
-#[tokio::test]
-async fn deleverage_succeeds_during_pause() -> anyhow::Result<()> {
```
jgur-psyops, line 19: this file and the super_ tests I think we can keep, these are logic changes that probably should persist.

programs/marginfi/src/instructions/marginfi_group/handle_bankruptcy.rs
```diff
@@ -49,28 +49,23 @@ pub fn lending_pool_handle_bankruptcy<'info>(
...
-        } else {
-            // admin can bankrupt banks in any state
-            validate_bank_state(&bank, InstructionKind::Unrestricted)?;
-            check!(is_admin_or_risk_admin, MarginfiError::Unauthorized);
-        }
-
-        utils::maybe_take_bank_mint(&mut ctx.remaining_accounts, &bank, token_program.key)?
-    };
+    let bank = bank_loader.load()?;
+    validate_bank_state(&bank, InstructionKind::FailsInPausedState)?;
```
jgur-psyops: The logic is slightly different here where the admin can't bankrupt a `Paused` bank. In fairness, that has never yet come up.

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
```diff
@@ -248,7 +248,6 @@ pub struct LendingAccountWithdraw<'info> {
     #[account(
         constraint = (
             !group.load()?.is_protocol_paused()
-            || marginfi_account.load()?.get_flag(ACCOUNT_IN_DELEVERAGE)
```
jgur-psyops, line 251: I think we should leave this in withdraw/repay wdt @IliaZyrin ? It's limited to the risk-admin only implicitly since only they can deleverage, so seems like a reasonable allowance during pause.

programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
```diff
@@ -68,6 +69,8 @@ pub struct LendingAccountPurgeDelevBalance<'info> {
     #[account(
         mut,
         has_one = group @ MarginfiError::InvalidGroup,
+        constraint = is_marginfi_asset_tag(bank.load()?.config.asset_tag)
```
jgur-psyops: Here I think the check was in error, we should be able to purge any kind of bank, with the caveat that we know this instruction is inherently destructive and bricks the bank.

jgur-psyops:
> q if we like to deploy it now, or in 0.1.9 upgrade

In 1.9 I think,  don't think there's a huge hurry because withdraw hardcodes a cold wallet, so even worst-case it's a troll.

programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
IliaZyrin: Yep

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
IliaZyrin: Exactly, let's keep it, otherwise we are very limited in what we can do to "fix" the things during such pauses.

type-crate/src/types/panic_state_cache.rs
IMEF-FEMI: reducing this to 6 hours to at least give more room for response.. in the case of an incidence

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
IliaZyrin: Still not reverted it seems

programs/marginfi/src/instructions/marginfi_group/handle_bankruptcy.rs
```diff
@@ -225,10 +225,9 @@ pub fn lending_pool_handle_bankruptcy<'info>(
 #[derive(Accounts)]
 pub struct LendingPoolHandleBankruptcy<'info> {
     #[account(
-        constraint = {
-            let g = group.load()?;
-            !g.is_protocol_paused() || signer.key() == g.admin || signer.key() == g.risk_admin
```
IliaZyrin, line 230: We probably also want to keep this, since deleveraging often leads to bankruptcy.

programs/marginfi/src/instructions/marginfi_account/withdraw.rs
IMEF-FEMI: This has actually been resolved here: https://github.com/0dotxyz/marginfi-v2/blob/5ad8fcb3271d9b1e94b31521b2fd85756b4efbbd/programs/marginfi/src/instructions/marginfi_account/withdraw.rs#L251

IliaZyrin: Ah, yes, I meant the integration withdraws (e.g. [this one](https://github.com/0dotxyz/marginfi-v2/pull/557/changes#diff-99bb30dfa38177622ee870b8c9eaebdf431f37703e4f09a90ea515e91723a6bbL311)).

programs/marginfi/tests/admin_actions/actions_during_pause.rs
```diff
@@ -304,102 +304,6 @@ async fn liquidation_still_blocked_during_pause() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Admin can handle_bankruptcy while protocol is paused.
-#[tokio::test]
-async fn handle_bankruptcy_by_admin_succeeds_during_pause() -> anyhow::Result<()> {
```
dawid-sw, line 309: this test should be reverted as we decided to allow bankruptcy by admin

dawid-sw:
We should also sync the docs with these new admin-permitted actions during pause and document the new function, which is turned off on mainnet, with an explanation about the rationale. I'm not 100% sure if this is the proper approach. I'm rather against keeping this function and having clutter in the codebase, but I understand the rationale behind it. 

so at least we should document it, bc it can look as little centralization risk.

programs/marginfi/src/instructions/marginfi_group/super_admin_deposit.rs
```diff
@@ -14,16 +14,24 @@ use marginfi_type_crate::{
...
 ///
-/// Note: Token-2022 transfer-fee extensions are not handled here. This instruction assumes
-/// the vault receives exactly `amount`.
+/// Token-2022 transfer-fee extensions are not handled here; the vault is assumed to receive
+/// exactly `amount`.
 pub fn super_admin_deposit<'info>(
     mut ctx: Context<'_, '_, 'info, 'info, SuperAdminDeposit<'info>>,
     amount: u64,
 ) -> MarginfiResult {
+    if crate::ID == MAINNET_PROGRAM_ID {
```
dawid-sw: just a nitpick, but we can do same checks as in clone_bank just be consistent. or even extract that to some helper function

```
    if crate::ID != STAGING_ID && crate::ID != LOCALNET_ID {
        panic!("Staging or localnet only!");
    }

    // Sanity check
    if crate::ID == MAINNET_PROGRAM_ID || *ctx.program_id == MAINNET_PROGRAM_ID {
        panic!("clone bank cannot run on mainnet deployment");
    }
```

programs/marginfi/src/instructions/marginfi_group/super_admin_deposit.rs
```diff
@@ -14,16 +14,24 @@ use marginfi_type_crate::{
...
-/// Group admin only.
-/// Transfers tokens from `admin_token_account` into the bank liquidity vault and raises
-/// `asset_share_value` so that existing depositor shares are increased proportionally.
+const MAINNET_PROGRAM_ID: Pubkey = pubkey!("MFv2hWf31Z9kbCa1snEPYctwafyhdvnV7FZnsebVacA");
+
+/// Group admin only. Staging/localnet only — panics on mainnet. See
+/// `guides/ADMIN/PERMISSIONS_AND_ROLES.md` ("Protocol Panic-Pause") for rationale.
+///
+/// Transfers `amount` from `admin_token_account` into the bank liquidity vault and raises
+/// `asset_share_value` so existing depositor shares are increased proportionally.
```
dawid-sw, line 22: just small note, the number of shares each depositor holds is fixed; what changes is the per-share value. After this IX runs, every existing depositor still owns the same share count, but each share is worth more

dawid-sw (approved):
some last comments, but overall LGTM

PR #558, opened 2026-04-08
Add on-chain indexer flags to MarginfiAccount

IMEF-FEMI:
# Summary                                                                                                                                                                                                  
                                                                                                                                                                                                             
  Added an `IndexerFlags` struct to `MarginfiAccount` so off-chain consumers like liquidators can filter accounts via memcmp without fully deserializing them. Each flag is a full byte for memcmp compatibility.
                                                                                                                                                                                                             
  ## Flags                  

  **Balance-derived** (synced automatically on every balance-mutating instruction):                                                                                                                          
  - `is_lending_only` — no liabilities (0 when account is empty)
  - `is_empty` — no balances above dust threshold                                                                                                                                                            
  - `is_single_borrower` — exactly one liability position                                                                                                                                                    
  - `has_staked` / `has_kamino` / `has_drift` / `has_juplend` — has a position with the corresponding asset tag                                                                                              
                                                                                                                                                                                                             
  **Permanent** (set once, never cleared):                                                                                                                                                                   
  - `has_ever_been_liquidated` — set during liquidation/receivership 
  - `has_ever_been_deleveraged` — set during deleverage             
  - `has_been_bankrupted` — set during bankruptcy                                                                                                                     
                                                                                                                                                                                                             
  **Pulse-derived** (refreshed best-effort during `pulse_health`):
  - `was_liquidatable` — maintenance health negative at last pulse                                                                                                                                           
  - `was_underwater` — equity health negative at last pulse                                                                                                                                                  
  - `was_active_30d` / `was_active_60d` — activity recency
  - `has_trivial_balance` — equity value < $1 at last pulse                                                                                                                                                  
  - `has_isolated` — has an isolated-tier liability. Set in real time by `lending_account_borrow` when borrowing from a `RiskTier::Isolated` bank, cleared in real time by health pulse.                           
                                                                                                                                                                                                             
  ## New instructions                                                                                                                                                                                        
                            
  - `sync_indexer_flags` — Permissionless batch sync via remaining accounts, for backfilling existing accounts.                                                                                              
  - `admin_close_account` — **Permissionless.** Closes accounts that are empty and have been inactive for ≥60 days (per the last `pulse_health` snapshot) and carry no blocking flags (disabled / flashloan /
   receivership). Rent is routed to the group's global fee wallet, enforced via an Anchor constraint against `group.fee_state_cache.global_fee_wallet` (`InvalidGlobalFeeWallet` on mismatch).


programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    #[account(
+        mut,
+        has_one = group @ MarginfiError::InvalidGroup,
+        close = rent_destination
+    )]
+    pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
+
+    #[account(
+        constraint = group.load()?.admin == admin.key() @ MarginfiError::Unauthorized
```
jgur-psyops: probably not feasible for us to sign for all of these with the MS. We could make it permless and send funds to the global fee wallet, or send funds to the keeper, or some mix of those.

type-crate/src/types/user_account.rs
```diff
@@ -103,6 +107,108 @@ impl MarginfiAccount {
...
+    /// 1 if account was active within the last 30 days (at last pulse)
+    pub was_active_30d: u8,
+    /// 1 if account was active within the last 90 days (at last pulse)
+    pub was_active_90d: u8,
+    /// 1 if account was active within the last year (at last pulse)
+    pub was_active_1y: u8,
+    /// 1 if asset_value_equity < $1 (at last pulse)
+    pub has_trivial_balance: u8,
+    /// 1 if account is empty and inactive for >30 days (at last pulse). Cleared on user activity.
+    pub pending_closure: u8,
```
jgur-psyops: isn't this basically the same as was_active_30d? Also, after we start closing accounts, was_active_90d and was_active_1y will be unreachable.

programs/marginfi/tests/user_actions/indexer_flags.rs
```diff
@@ -0,0 +1,317 @@
+use fixtures::marginfi_account::MarginfiAccountFixture;
+use fixtures::prelude::*;
+use pretty_assertions::assert_eq;
+use solana_program_test::*;
+use solana_sdk::{clock::Clock, signature::Keypair};
+
```
jgur-psyops, line 6: since it's somewhat cumbersome to test the Kamino/Drift/Juplend flags here in Rust you can add those assertions to existing tests in the TS test suites, there should already be enough tests there for you to piggy-back on to show flag sets on deposit, persists on withdraw, closes on last withdraw-all.

type-crate/src/types/user_account.rs
```diff
@@ -103,6 +107,108 @@ impl MarginfiAccount {
...
+                continue;
+            }
+            has_any_balance = true;
+
+            if !balance.is_empty(BalanceSide::Liabilities) {
+                liability_count = liability_count.saturating_add(1);
+            }
+
+            let tag = balance.bank_asset_tag;
+            if tag != ASSET_TAG_DEFAULT && tag != ASSET_TAG_SOL {
```
jgur-psyops: Doesn't this treat Kamino/staked/drift/etc as isolated? I think it's probably misnamed then, it's more like "has non-default" which is already exclusive with some of the other flags and probably not needed.

dawid-sw: Agree, seems like just its enough to filter by tags

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,61 @@
...
+};
+
+/// Permissionless instruction to close accounts that are empty and have been inactive for >60
+/// days (per the last pulse_health snapshot). The account must also have no blocking flags
+/// (disabled, flashloan, receivership). Rent is returned to the group's global fee wallet.
+pub fn admin_close_account(ctx: Context<AdminCloseAccount>) -> MarginfiResult {
+    let marginfi_account = ctx.accounts.marginfi_account.load()?;
+
+    check!(
+        marginfi_account.indexer_flags.is_empty == 1
```
dawid-sw, line 18: I'm not convinced about activity flags. It can reflect the state of account wrongly. 

a) Activity flags are only refreshed during pulse_health, but last_update (the source of truth) is refreshed on every balance-mutating instruction.
This creates a window where the flags are wrong:

1. T=0: User deposits. last_update=0
2. T=61d: Someone calls pulse_health. was_active_60d=0
3. T=62d: User deposits again. last_update=62d — but was_active_60d stays 0 (stale)
4. T=62d: User withdraws everything. is_empty=1, was_active_60d still 0
5. Anyone calls admin_close_account — passes both checks, account closed, rent goes to fee wallet instead of user

b) Pulsing empty inactive accounts is a net cost. The whole point of these flags is to identify accounts worth closing. But to mark an account as inactive, someone must call pulse_health, which does not make much sense on empty account.

c) Not sure if there is much usability on frontend. Knowing how old account is doesn't seems like a useful function for liquidators. Our own crank tool will close old accounts anyway, as fee is landing on global fee wallet, so there is incentive for 3rd party to do so. 

For closing function it should be enough to compare against last updated time. 

Or if you guys think its useful maybe we can just bump that flags in sync_flags function.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -159,6 +165,11 @@ impl MarginfiAccountImpl for MarginfiAccount {
 
         !is_disabled && only_has_empty_balances && !is_in_flashloan && !is_in_receivership
```
dawid-sw, line 166: we should also disallow closing frozen accounts. if we froze because of some compliance issue, then such account should not be re-opened.

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,61 @@
...
+            && marginfi_account.indexer_flags.was_active_60d == 0,
```
dawid-sw, line 19: We should do a separation of concepts here, and use flags just for clients pulling the data, which is very small risk comparing to closing account. I'd do invariant checks for closing accounts like `can_be_closed` is doing and add there check of account in-activity. 

for example 30d and 60d active flags are going to be `0` after deployment untill someone call pulse, so by default all empty accounts will be treated as inactive and closable

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -192,6 +192,11 @@ pub fn lending_account_borrow<'info>(
     let mut health_cache = HealthCache::zeroed();
     health_cache.timestamp = clock.unix_timestamp;
     marginfi_account.lending_account.sort_balances();
+    marginfi_account.sync_indexer_flags();
+
+    if ctx.accounts.bank.load()?.config.risk_tier == RiskTier::Isolated {
+        marginfi_account.indexer_flags.has_isolated = 1;
```
dawid-sw, line 198: as some other comments already pointed, seems like this flag has a role of showing bank risk tier but also type of asset user is having

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -121,6 +121,7 @@ pub fn start_receivership<'info>(
     write_liquidation_price_cache_from(marginfi_account, remaining_ais, &liq_price_cache)?;
     marginfi_account.health_cache = health_cache;
     marginfi_account.set_flag(ACCOUNT_IN_RECEIVERSHIP, false);
+    marginfi_account.indexer_flags.has_ever_been_liquidated = 1;
```
dawid-sw, line 124: it might be useful do differentiate between standard liquidation and receivenship liquidation and just have separate flags there

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
IMEF-FEMI: Just made some updates wrt to the concerns.

1. user actions like deposit/withdraw/borrow would now trigger flag updates including the was_active_* flags 
2. an account does not require pulse_health call to be eligible for closure (just helpful for setting flags)

In my understanding, the flags would be to let an indexer skip the hurdle of fetching filtering multiple accounts on the client side. saving them some time...

with that being said. we only really need to pulse each account once to backfill (if we can afford to)
after that, user actions do the rest

programs/marginfi/src/instructions/marginfi_account/borrow.rs
IMEF-FEMI: it currently only shows if a user has an isolated liability

i dont think there would be an advantage to having it represent isolated "asset" also

if there is, i'd rather we have two different flags for them: [has_isolated_liability, has_isolated_asset]

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
IMEF-FEMI: same here also... would there be a benefit to separating them

dawid-sw: ah I meant to have a differentiation between liquidation and deleverage so we can quickly get all accounts that were deleveraged but not sure if this is needed

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,69 @@
...
+/// The account must also have no blocking flags (disabled, flashloan, receivership).
+/// Rent is returned to the group's global fee wallet.
+pub fn admin_close_account(ctx: Context<AdminCloseAccount>) -> MarginfiResult {
+    let marginfi_account = ctx.accounts.marginfi_account.load()?;
+    let clock = Clock::get()?;
+    let elapsed = clock
+        .unix_timestamp
+        .saturating_sub(marginfi_account.last_update as i64);
+    let is_inactive =
+        marginfi_account.indexer_flags.was_active_60d == 0 || elapsed > 60 * SECONDS_PER_DAY;
```
dawid-sw: there is potential issue:
1) program gets updated, so by default was_active_60d flag is `0`
2) so this check is always true no matter what is the `last_update` parameter
3) then if account is empty someone can run sync_flags_ix and mark empty flag as `1`

so in result anyone can close empty account no matter if it's inactive or not. I think this was_active_60d should be skipped.

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,69 @@
...
+    state::marginfi_account::MarginfiAccountImpl,
+    MarginfiError, MarginfiResult,
+};
+
+/// Permissionless instruction to close accounts that are empty and have been inactive for >60
+/// days. Inactivity is accepted from either the `was_active_60d` indexer flag (for memcmp
+/// discovery by indexers) or `clock - last_update > 60d` (so a pulse is not required).
+/// The account must also have no blocking flags (disabled, flashloan, receivership).
+/// Rent is returned to the group's global fee wallet.
+pub fn admin_close_account(ctx: Context<AdminCloseAccount>) -> MarginfiResult {
```
dawid-sw, line 18: we need to change this function name, currently its permission-less not only admin

type-crate/src/types/user_account.rs
```diff
@@ -103,6 +107,118 @@ impl MarginfiAccount {
...
+        self.is_empty = (!has_any_balance) as u8;
+        self.is_lending_only = (has_any_balance && liability_count == 0) as u8;
+        self.is_single_borrower = (liability_count == 1) as u8;
+        self.has_staked = staked as u8;
+        self.has_kamino = kamino as u8;
+        self.has_drift = drift as u8;
+        self.has_juplend = juplend as u8;
+
+        if liability_count != 1 {
+            self.has_isolated = 0;
```
dawid-sw, line 215: Pre-upgrade accounts that already hold an isolated borrow read has_isolated = 0 because that byte was old _padding0, and nothing on-chain ever sets the flag back to 1 from balances alone (sync_balance_derived only clears it; the borrow ix is the sole setter), so indexers filtering on has_isolated == 1 will silently miss those legacy positions until the user repays and re-borrows. 

 We don't have access to bank here. two possible solutions:
 1) ignore fact that after migration accounts can show incorrect liability flag
 2) do this check on pulse health where we have access to bank
 3) do another version of sync flags which should be run after migration

dawid-sw (commented):
last comments

type-crate/src/types/user_account.rs
jgur-psyops: That raises another concern actually: if a Bank goes from Collateral -> Isolated, then this flag is incorrect for all previous holders where it was marked false. I think that's fine since it's a non-authoritative flag.

We could have the backfill instruction take an `isolated_bank_maybe: Option<Bank>` where the bank is any Isolated bank that the user holds. The backfill ix validates (1) the user has a balance in that bank, (2) it's isolated, then sets the flag. When the bank is not provided, the flag remains unchanged, as we cannot know it is false.

It's still a little cumbersome in the case Dawid describes but not too bad in the Collateral -> Isolated case since we just send a backfill to all holders with that bank flagged.

We could also pass all Banks as remaining accounts to the backfill/sync ix but that's super cumbersome.

programs/marginfi/src/instructions/marginfi_account/admin_close.rs
```diff
@@ -0,0 +1,69 @@
...
+        MarginfiError::IllegalAction,
+        "Account is not eligible for close (not empty or active within 60d)"
+    );
+
+    check!(
+        marginfi_account.can_be_closed(),
+        MarginfiError::IllegalAction,
+        "Account cannot be closed"
+    );
+
```
jgur-psyops, line 40: Add the usual state checks too to be on the safe side (no flashloan, liquidation, in order execution).

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
jgur-psyops: Agreed, I think it would have been helpful during the Drift incident to quickly grab all the deleveraged accounts.

programs/marginfi/src/instructions/marginfi_account/pulse_health.rs
```diff
@@ -102,7 +112,24 @@ pub fn lending_account_pulse_health<'info>(
...
+    // Reborrow through a single DerefMut so the borrow checker can split indexer_flags
+    // (mut) from lending_account.balances (shared).
+    let account = &mut *marginfi_account;
+    account
+        .indexer_flags
+        .sync_balance_derived(&account.lending_account.balances);
+    account.indexer_flags.sync_activity_flags(elapsed);
+    account.indexer_flags.was_liquidatable = is_liquidatable as u8;
+    account.indexer_flags.was_underwater = (equity_assets < equity_liabs) as u8;
+    account.indexer_flags.has_trivial_balance =
```
jgur-psyops: I'm not sure if we can set flags like this on pulse_health: a pulse will always succeed unless it panics, if it hits an oracle error or similar it will still return a result even with an invalid price. See for example `(user 1) health pulse with bad oracle - cache notes the missing price`. We could set these flags only if the pulse does not have an error flag raised internally.

Should have a test demonstrate that a pulse where all the oracles are bad (and therefore all prices set to 0) does not raise this flag incorrectly.

PR #559, opened 2026-04-19
Fix i128→i64 overflow panic in StakedWithPythPush

arussel:
The adjusted_price calculation in the StakedWithPythPush oracle path uses .try_into().unwrap() for i128→i64 conversion. This panics when cached sol_pool/lst_supply values are transiently out of sync, which happens off-chain with gRPC streaming (different slots for stake pool vs oracle accounts).

Replace .unwrap() with .map_err(|_| MathError)? so both on-chain and client-side callers get a clean error instead of a panic.


dawid-sw (changes requested):
math_error!() macro can be used to have better debugging log

dawid-sw (changes requested):
math_error!() macro can be used to have better debugging log

PR #560, opened 2026-04-19
Add close_liquidation_record instruction

arussel:
Adds a permissionless instruction to close LiquidationRecord PDAs and return
  ~0.005 SOL rent to the original payer (record_payer field).

  Currently these PDAs are created during v2 liquidations but never closed,
  permanently locking rent. This is particularly relevant for active liquidators
  that accumulate hundreds of records over time.

  ## Design

  - Permissionless — anyone can call, rent always returns to `record_payer`
  - Fails if account is in receivership or deleverage (active liquidation)
  - Fails if `liquidation_receiver != default` (record still in use)
  - Resets the account's `liquidation_record` field to default
  - Implements `Hashable` so the instruction is correctly blocked between
    start/end liquidation

  The `record_payer` field already existed with an anticipatory comment in
  LiquidationRecord — this implements that intent.

  ## Tests

  - Happy path: init record, close, verify rent returned and field reset
  - Permissionless: third-party caller succeeds, rent goes to original payer
  - Fails during active liquidation (ForbiddenIx)
  - Fails with wrong record_payer (Unauthorized)
  - Fails with record belonging to different account (InvalidLiquidationRecord)


jgur-psyops:
Note: also see comments on #509 (this version will likely supercede that PR as it is quite stale.)

arussel:
@jgur-psyops Hey — I wasn't aware of the preceding PR when I started this one. I've added the 60-day inactivity cooldown as discussed (derived from max(entries[].timestamp)), and records that were never liquidated can be closed immediately.  I left out the admin closing path — is there a concrete use case for it? Would marginfi ever bother closing these?

jgur-psyops:
> @jgur-psyops Hey — I wasn't aware of the preceding PR when I started this one. I've added the 60-day inactivity cooldown as discussed (derived from max(entries[].timestamp)), and records that were never liquidated can be closed immediately. I left out the admin closing path — is there a concrete use case for it? Would marginfi ever bother closing these?

There are 100k+ marginfi accounts that have been empty and inactive for a long time (years). We have only had Liquidation Records for a short time (less than a year) and there not many of them today, but if we find ourselves with 100k+ some years from now that are clearly not being used, then we would probably close them!

In general we prefer to simply leave accounts open, until it becomes annoying for indexing, like when a fetchMultiple call to get all the accounts of some type gets you mostly-defunct entries.

arussel:
@jgur-psyops Thanks for the merge! Any rough ETA on prod deploy? Asking because I've got some rent stuck from this exact case and want to plan the recovery.

jgur-psyops:
> @jgur-psyops Thanks for the merge! Any rough ETA on prod deploy? Asking because I've got some rent stuck from this exact case and want to plan the recovery.

Thank you for the contribution! We are expecting roughly end of April, but might end up deferring to the second week of May due to Accelerate conference.

PR #563, opened 2026-04-22
Price multiplier split in cache

jgur-psyops:
The price cache now tracks the Kamino/Drift/Juplend/Staked multiplier separately from the actual underlying price, enabling consumers that already have the multiplier from some other source to consume the price (or vice-versa). 

For example, our own front end already tracks the multiplier and applies it to the token amount rather than the price. When fetching the price, it wants the oracle price before multipliers. 

This should not affect any functionality except the cache.

Also fixes the old bug where Staked Collateral position confidence was not being multiplied by the multiplier like all other integrations are.

Note: In some flows, to keep math consistent, we have to compute the multiplier twice because mul-div and div-mul are not exactly associative, i.e. `A * B / C` and `A * (B / C)` have different precision. If we tolerate converting all math into I80F48 and using the `A * (B / C)` for all computations we can simplify this considerably.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/borrow.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/borrow.rs
@@ -14,8 +14,8 @@ use crate::{
         rate_limiter::GroupRateLimiterImpl,
     },
     utils::{
-        self, fetch_unbiased_price_for_bank, is_marginfi_asset_tag, record_withdrawal_outflow,
-        validate_asset_tags, validate_bank_state, InstructionKind,
+        self, fetch_unbiased_price_for_bank_cache, is_marginfi_asset_tag,
+        record_withdrawal_outflow, validate_asset_tags, validate_bank_state, InstructionKind,
     },
 };
 use anchor_lang::prelude::*;
@@ -204,9 +204,13 @@ pub fn lending_account_borrow<'info>(
 
     let bank_pk = ctx.accounts.bank.key();
     let mut bank = ctx.accounts.bank.load_mut()?;
-    let price = fetch_unbiased_price_for_bank(&bank_pk, &bank, &clock, ctx.remaining_accounts).ok();
+    let price_for_cache =
+        fetch_unbiased_price_for_bank_cache(&bank_pk, &bank, &clock, ctx.remaining_accounts).ok();
 
-    let rate_limit_price = price.as_ref().map(|p| p.price).unwrap_or(I80F48::ZERO);
+    let rate_limit_price = price_for_cache
+        .as_ref()
+        .map(|p| p.oracle_price.price)
+        .unwrap_or(I80F48::ZERO);
     record_withdrawal_outflow(
         group_rate_limit_enabled,
         amount_pre_fee,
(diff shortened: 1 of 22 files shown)


programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -204,9 +204,13 @@ pub fn lending_account_borrow<'info>(
 
     let bank_pk = ctx.accounts.bank.key();
     let mut bank = ctx.accounts.bank.load_mut()?;
-    let price = fetch_unbiased_price_for_bank(&bank_pk, &bank, &clock, ctx.remaining_accounts).ok();
+    let price_for_cache =
+        fetch_unbiased_price_for_bank_cache(&bank_pk, &bank, &clock, ctx.remaining_accounts).ok();
 
-    let rate_limit_price = price.as_ref().map(|p| p.price).unwrap_or(I80F48::ZERO);
+    let rate_limit_price = price_for_cache
```
dawid-sw: normal price (without multiplier is used here), I guess for now its fine as these are normal assets (not integrations) and staked cannot be borrowed. but just to be double sure, I'd use final price here, so outflow is recording real-value.

programs/marginfi/src/state/price.rs
```diff
@@ -31,6 +31,17 @@ use switchboard_on_demand::{
...
+/// Price per unit before any multipliers are applied, where `price_multiplier` shows what
+/// multipliers will be applied to generate the true deposited-token price.
+/// * Example: a Staked Collateral bank with an exchange rate of 2 for stake/sol and a price of $50
+/// for SOL will show $50 here, and multiplier will be 2.
+/// * For any bank that does not have a multiplier, `price_multiplier = 1` and `oracle_price`
+///   matches the adjusted value.
+#[derive(Copy, Clone, Debug)]
+pub struct OraclePriceWithMultiplier {
+    pub oracle_price: OraclePriceWithConfidence,
+    pub price_multiplier: I80F48,
```
dawid-sw, line 43: we need to remember to pulse all banks after program update

programs/marginfi/src/state/price.rs
```diff
@@ -321,17 +424,20 @@ impl OraclePriceFeedAdapter {
...
+                    Some(price_feed.get_price_and_confidence_of_type(price_type, u32::MAX)?)
+                } else {
+                    None
+                };
 
-                let (total_liq, total_col) = reserve.scaled_supplies()?;
-                if total_col > I80F48::ZERO {
-                    let multiplier = total_liq / total_col;
-
+                if multiplier > I80F48::ZERO {
```
dawid-sw: The gate changed from if total_col > 0 to if multiplier > 0 here - they're equivalent in normal operation, but in the corner case where total_col > 0 && total_liq == 0 the old code zeroed the price and the new one leaves it at the raw oracle. I don't think Kamino/Solend reserves can actually reach that state in practice, so this isn't really a bug, just pointing it out as some bevaviour change.

dawid-sw (approved):
lgtm, some small comments

PR #564, opened 2026-04-23
Add close_liquidation_record, update close account

jgur-psyops:
Thank you to https://github.com/arussel for starting the feature #560 

## New Instructions

* Add close_liquidation_record instruction

Allows closing a liquidation record PDA and returning rent to the original payer (record_payer). This enables liquidators to reclaim the ~0.005 SOL rent they paid when initializing records for accounts they liquidated. To be closed, Records must have been inactive for 60d+ or have never been used. 

Permissionless — anyone can call, but rent always goes to record_payer (the wallet that originally paid). This also allows cleanup bots to reduce on-chain state bloat from stale records. 

## New Fields

Marginfi Account now has `active_orders`, which tracks how many Orders the account has open.

## Other Notes

Now requires Records and Orders bound to an Account to be closed before Accounts can be closed, preventing them from being orphaned.

### Known Issues

Liquidators can "grief" empty accounts by opening a Record for them and liquidating them for zero dollars, making them unable to close for 60 days. This doesn't really cost the victim anything, most users never close accounts anyways, and the liquidator locks up $1.5 in rent to pull it off, so there is no serious risk.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
@@ -17,7 +17,7 @@ use marginfi_type_crate::{
     constants::MARGINFI_ACCOUNT_SEED,
     types::{
         LendingAccount, MarginfiAccount, MarginfiGroup, ACCOUNT_DISABLED, ACCOUNT_IN_FLASHLOAN,
-        ACCOUNT_IN_RECEIVERSHIP,
+        ACCOUNT_IN_ORDER_EXECUTION, ACCOUNT_IN_RECEIVERSHIP,
     },
 };
 
@@ -41,6 +41,17 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
     check!(
         !old_account.get_flag(ACCOUNT_IN_RECEIVERSHIP),
         MarginfiError::ForbiddenIx
+    );
+
+    check!(
+        !old_account.get_flag(ACCOUNT_IN_ORDER_EXECUTION),
+        MarginfiError::ForbiddenIx
+    );
+
+    check!(
+        old_account.active_orders == 0,
+        MarginfiError::IllegalAction,
+        "Close all active orders before transfer"
     );
 
     // Prevent multiple migrations from the same account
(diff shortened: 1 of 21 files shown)


type-crate/src/types/user_account.rs
```diff
@@ -65,8 +65,13 @@ pub struct MarginfiAccount {
     pub third_party_index: u16,
     /// This account's bump, if a PDA-based account (0.1.5 or later). Otherwise, does nothing.
     pub bump: u8,
+    /// Count of how many Orders this account has active. One is added when an Order is opened, and
+    /// subtracted when an Order is executed or cancelled.
+    /// * Accounts cannot open more than u8::MAX orders. Sorry power users: hopefully 256 stop
+    ///   losses is enough for you.
+    pub active_orders: u8,
```
dawid-sw, line 72: we need to make sure that no orders exists before doing this migration. should be fine as orders feature is not used in mainnet, just pointing that. 

if account has order and we are setting active_orders as `0`, then closing such order is not possible

tests/16_orders.spec.ts
```diff
@@ -23,8 +23,12 @@ import {
   closeOrderIx,
   keeperCloseOrderIx,
   setKeeperCloseFlagsIx,
+  accountInit,
```
dawid-sw, line 26: unused variables in this import, also Keypair above

tests/19_closeAccount.spec.ts
```diff
@@ -0,0 +1,179 @@
+import { Program } from "@coral-xyz/anchor";
+import {
+  ComputeBudgetProgram,
+  Keypair,
+  PublicKey,
```
dawid-sw, line 5: unused variable

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
```diff
@@ -43,6 +43,17 @@ pub fn transfer_to_new_account(ctx: Context<TransferToNewAccount>) -> MarginfiRe
...
+    check!(
+        !old_account.get_flag(ACCOUNT_IN_ORDER_EXECUTION),
+        MarginfiError::ForbiddenIx
+    );
+
+    check!(
+        old_account.active_orders == 0,
+        MarginfiError::IllegalAction,
+        "Close all active orders before transfer"
+    );
```
dawid-sw, line 55: addressing one point from security report, lets discuss if health_check should also be transfered, imo not necesarly

dawid-sw (approved):
lgtm, some insights to discuss

type-crate/src/types/user_account.rs
jgur-psyops: I was debating allowing it to saturating_sub in 1.9 to avoid someone running into this, but as we haven't launched Orders officially yet, I think we'll just say in the patch notes that any Orders placed prior to 1.9 (which is none, currently) must be closed before 1.9 launches or they will be stuck/orphaned forever.

programs/marginfi/src/instructions/marginfi_account/transfer_account.rs
jgur-psyops: Agree, I think it doesn't really matter either way. One weird thing I didn't address is that the new account inherits your `bank_asset_tag` on Balances, but this is benign as without an active Order they don't do anything.

PR #567, opened 2026-04-27
Per bank Circuit breaker guard against oracle attack

IMEF-FEMI:
This PR adds a per-bank oracle circuit breaker to reduce the blast radius of bad or manipulated price moves. The goal is to stop a bank from continuing normal risk-taking when the live oracle price jumps too far from its cached reference, using tiered temporary halts similar to market circuit breakers.

The breaker compares each fresh oracle observation against an EMA-based reference price:

`new_reference = alpha * current_price + (1 - alpha) * old_reference`

To avoid fast re-anchoring on attacker-controlled moves, the reference update is clipped per pulse:

`clipped_shift = clamp(new_reference - old_reference, -max_shift, +max_shift)`

`reference_after_update = old_reference + clipped_shift`

where:

- `alpha = cb_ema_alpha_bps / 10_000`
- `max_shift = old_reference * CB_MAX_EMA_SHIFT_BPS_PER_PULSE / 10_000`

The breaker then measures the observed move in basis points after discounting oracle confidence:

`raw_delta = abs(current_price - reference_price)`

`effective_delta = max(raw_delta - confidence, 0)`

`deviation_bps = effective_delta * 10_000 / reference_price`

If `deviation_bps` crosses a configured tier threshold for enough consecutive counted observations, the bank trips into a temporary halt. Repeated re-breaches inside the escalation window ratchet the bank into longer halts, and repeated severe tier-3 trips can promote the bank into a full paused state.

## Changes

- Added configurable circuit-breaker settings to bank config: enable flag, 3 deviation tiers, 3 halt durations, sustain-observation count, escalation window multiplier, and EMA alpha for the reference price.
- Added bank-side circuit-breaker runtime/state machine logic that:
  - tracks an EMA reference price,
  - ignores duplicate/stale observations,
  - subtracts oracle confidence before evaluating the move,
  - requires repeated breach observations before tripping,
  - escalates halts across tiers on repeated re-breaches,
  - can auto-promote repeated severe tier-3 events into a full bank pause.
- Enforced halt behavior in instruction flows:
  - block borrow/withdraw and adapter flows that increase risk,
  - allow risk-reducing actions like deposit/repay to continue,
  - freeze interest accrual while a bank is circuit-breaker halted,
  - restrict direct liquidation to admin/risk-admin when a halted bank is involved.
- Added admin/risk-admin recovery path via `lending_pool_clear_circuit_breaker`, with optional reference-price reseed.
- Added new events and errors so trips, clears, observed breaches, and auto-pauses are visible on-chain.
- Wired the new fields through shared types, TypeScript helpers, test utils, and CLI config/update surfaces.
- Added Rust and TS coverage for config validation, halt triggering, blocked/allowed actions during halt, escalation behavior, and admin clear flow.


programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -176,6 +152,60 @@ pub fn lending_account_liquidate<'info>(
 
     liquidatee_marginfi_account.lending_account.sort_balances();
 
+    {
+        let group = marginfi_group_loader.load()?;
+        let cb_admin_liquidation = any_balance_bank_is_cb_halted(
```
IliaZyrin, line 158: Shouldn't we only check for asset_bank and liab_bank here, not for all the balances?

programs/marginfi/src/state/bank.rs
```diff
@@ -432,7 +460,57 @@ impl BankImpl for Bank {
...
+            if flag && !was_enabled {
+                let last: I80F48 = self.cache.last_oracle_price.into();
+                check!(
+                    last > I80F48::ZERO,
+                    MarginfiError::CircuitBreakerRequiresWarmCache
+                );
+                let now = Clock::get()?.unix_timestamp;
+                let age = now.saturating_sub(self.cache.last_oracle_price_timestamp);
+                check!(
+                    (0..=CB_ENABLE_MAX_PRICE_AGE_SECONDS).contains(&age),
```
IliaZyrin, line 657: Why not just:

```suggestion
                    age <= CB_ENABLE_MAX_PRICE_AGE_SECONDS,
```

programs/marginfi/src/state/bank.rs
```diff
@@ -432,7 +460,57 @@ impl BankImpl for Bank {
             self.update_flag(flag, TOKENLESS_REPAYMENTS_ALLOWED);
         }
 
+        if let Some(flag) = config.circuit_breaker_enabled {
+            msg!("setting circuit breaker enabled: {:?}", flag);
+            let was_enabled = self.get_flag(CIRCUIT_BREAKER_ENABLED);
+            // Seed the EMA from the cached oracle price (must be fresh) so the first post-enable
```
IliaZyrin: This is confusing. The attacker might have been feeding the wrong price for some time already (by the time admin decides to enable cb), so it may be wrong in the cache as well. Unless we want to just enable cb on all the banks once & forever?

programs/marginfi/src/state/bank.rs
```diff
@@ -617,13 +701,162 @@ impl BankImpl for Bank {
...
+    ///
+    /// Pulses are deduped by Solana slot, by `CB_MIN_PULSE_SLOT_GAP`, and by oracle publish-time;
+    /// confidence is subtracted from raw delta before tier comparison; the EMA shift is clipped
+    /// to `CB_MAX_EMA_SHIFT_BPS_PER_PULSE` of the current reference. After
+    /// `CB_MAX_TIER3_BEFORE_PAUSE` consecutive tier-3 trips the bank is forced to `Paused`.
+    fn update_circuit_breaker(
+        &mut self,
+        now: i64,
+        slot: u64,
+        obs: OraclePriceWithConfidence,
```
IliaZyrin: Nitpick but I would rename this one to just `oracle`.

programs/marginfi/src/state/bank.rs
```diff
@@ -617,13 +701,162 @@ impl BankImpl for Bank {
...
+        if !self.get_flag(CIRCUIT_BREAKER_ENABLED) {
+            return Ok(());
+        }
+
+        // Halted: freeze detection and EMA until the tier timer expires.
+        if self.cb_tier > 0 && now < self.cb_halt_ended_at {
+            return Ok(());
+        }
+
+        // Escalation window expired without a re-breach: fall back to operational.
```
IliaZyrin: Nitpick: I think it's better to move this comment to the line 752.

programs/marginfi/src/state/bank.rs
```diff
@@ -798,6 +1031,65 @@ impl BankImpl for Bank {
...
+        msg!("CB halt tier {} duration {}s", new_tier, dur_sec);
+        #[cfg(not(test))]
+        emit!(crate::events::CircuitBreakerTrippedEvent {
+            tier: new_tier,
+            deviation_bps,
+            halt_started_at: now,
+            halt_ended_at: self.cb_halt_ended_at,
+        });
+        #[cfg(test)]
+        let _ = deviation_bps;
```
IliaZyrin, line 1434: No longer needed I believe.

IliaZyrin (commented):
Amazing implementation!

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IMEF-FEMI: The check intentionally covers every bank backing an active balance on the account, not just asset_bank/liab_bank. Liquidation eligibility is decided by the account's global health, which is computed across all balances... so a CB-halted bank oracle on any of the liquidatee's banks can distort the health

programs/marginfi/src/state/bank.rs
IMEF-FEMI: you're right...

What the check does is actually narrower. it meant stops an attacker from sandwiching the enable tx to control the seed. Without seeding here,  cb_reference_price stays 0 and the first post-enable pulse seeds from whatever the live price is at that moment, which is attacker-timeable. Seeding from the cache anchors to a price some prior, unrelated instruction already observed. which a subsequent instruction can reference 

i've updated the comment to capture this

programs/marginfi/src/state/bank.rs
IMEF-FEMI: the idea was to check age >= 0 && age <= MAX initially but clippy's manual_range_contains actually enforces the (0..=MAX).contains(&age) form for bounded checks... so our lint script would fail

IMEF-FEMI:
> Amazing implementation!

Thanks 🙏

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IliaZyrin: Well, in the worst case it would just block a valid liquidation then, right?
I.e. the same effect as having this check in place, but with less individual checks.

programs/marginfi/src/state/bank.rs
IliaZyrin: Hmm, but what prevents attacker from issuing that "prior, unrelated instruction"? Isn't it the same as sandwiching in terms of the final result?

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IMEF-FEMI: What I meant is:

 * Liquidation eligibility here is based on the account’s global health, not just the selected asset_bank / liab_bank. So an unrelated bank that is already CB-halted can still distort the liquidatee’s total health contribution, because that bank is still one of the active balances used by the risk engine. 

* Checking only the asset/liab pair can therefore hurt the user: their account may genuinely still be solvent and the price divergence just transient market manipulation. 

* That's why it only makes sense to gate this to an admin someone who can judge whether the price spike is real or a facade (or someone that can tell off chain when market has returned back to normalcy) rather than letting any liquidator act on a health number we know is unreliable.

programs/marginfi/src/state/bank.rs
IMEF-FEMI: the check here ```(0..=CB_ENABLE_MAX_PRICE_AGE_SECONDS).contains(&age)``` is basically the same as 
```age >= 0 && age <= CB_ENABLE_MAX_PRICE_AGE_SECONDS ``` as age is an i64, and there is still a chance it could be negative. but if we use  `age <= CB_ENABLE_MAX_PRICE_AGE_SECONDS` it does not account for the negative check and clippy does not allow us do ```"age >= 0 && age <= CB_ENABLE_MAX_PRICE_AGE_SECONDS"```

@IliaZyrin

File: programs/marginfi/src/state/bank.rs (hunk 3 above)
Comment (IMEF-FEMI) [reply programs/marginfi/src/state/bank.rs]: this does not fully prevent an attacker from controlling the seed if they are able to get a manipulated price into the cache first through a prior permissionless pulse. What it does prevent is leaving cb_reference_price = 0 and then letting the first post-enable observation choose the seed. Instead, it requires the seed to come from a recent cached price, which the admin may also have refreshed explicitly by bundling a pulse ix in the same transaction, rather than from an arbitrary later pulse.

File: programs/marginfi/src/state/bank.rs (hunk 2 above)
Comment (IliaZyrin) [reply programs/marginfi/src/state/bank.rs:657]: Yep, I understood, thank you!

File: programs/marginfi/src/instructions/marginfi_account/liquidate.rs (hunk 1 above)
Comment (IliaZyrin) [reply programs/marginfi/src/instructions/marginfi_account/liquidate.rs:158]: Yeah, makes sense!

**Review by IliaZyrin — APPROVED** [review]

File: programs/marginfi/src/instructions/marginfi_account/borrow.rs [hunk 7]
```diff
@@ -86,7 +86,7 @@ pub fn lending_account_borrow<'info>(
         let mut bank = bank_loader.load_mut()?;
 
         validate_asset_tags(&bank, &marginfi_account)?;
-        validate_bank_state(&bank, InstructionKind::FailsIfPausedOrReduceState)?;
+        validate_bank_state(&bank, InstructionKind::FailsIfPausedOrReduceState, false)?;
```
jgur-psyops, line 89: Here I think the approach is fundamentally flawed to some extent. We are only checking the bank being borrowed, so an attacker lending A and borrowing B can manipulate the price of A, then exfiltrate funds from B - but no breaker is tripped. The same applies to risk-carrying withdraws (although for withdraws on accounts with no borrows, the breaker should never activate). 

I believe we must actually check/record the breaker state for EVERY involved bank. This may require some re-wiring of the risk engine to either update the breaker state as it goes through the banks or pass the prices back up the callstack (I would prefer the latter, I think, as it can simplify the existing cache logic too), because while incurring a `fetch_unbiased_price_for_bank_with_cache` was fine for one bank, it's probably too CU intensive for 16. This will be a similar pattern to how the receivership liquidation price cache works at the moment.

The other side effect is that the banks likely will all have be marked writeable/mutable. That's breaking but probably fine: as long as deposit + no-risk withdraws are unaffected it will have minimal impact on consumers. Again we already did this for the liquidation price cache.

cc @IliaZyrin @dawid-sw thoughts on what the best approach might be here?

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops, line 240: I believe since breaker validation happens here, but is checked above, the tx that triggers the halt will actually still pass, and only the next one will fail?

programs/marginfi/src/state/bank.rs
```diff
@@ -798,6 +1042,65 @@ impl BankImpl for Bank {
...
+
+        if new_tier == 3 {
+            self.cb_tier3_consecutive_trips = self.cb_tier3_consecutive_trips.saturating_add(1);
+            // Force Paused so all state-changing operations (incl. deposit/repay) stop until
```
jgur-psyops: Perhaps we should add a "CircuitBroken" state and consider what should be allowed. It seems like deposit/repay should continue to operate, along with riskless-withdraw (no associated borrow).

We must also consider that "Paused" suspends liquidations: if the oracle price were actually being manipulated that would be the correct course of action, but in a genuine market panic that would be quite bad. Perhaps what we might do is allow the risk admin to continue liquidating under a "CircuitBroken" state, but block everyone else. I'm not sure really.

jgur-psyops: Of course if "CircuitBroken" is its own state, then a bank shifting from "ReduceOnly" should not re-enable deposits/repays, so it's actually somewhat complex.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
dawid-sw: I agree. ideally we should check all banks/oracles. the option with passing prices back to callstack sounds good to me as well. yeah it will be breaking change for borrow but it is what it is

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -177,6 +157,60 @@ pub fn lending_account_liquidate<'info>(
...
+            &liquidatee_marginfi_account,
+            liquidatee_remaining_accounts,
+        )?;
+
+        // During a CB halt direct liquidation is admin-only; otherwise the standard liquidator
+        // ownership check applies.
+        let signer = ctx.accounts.authority.key();
+        if cb_admin_liquidation {
+            require!(
+                signer == group.risk_admin || signer == group.admin,
```
dawid-sw, line 168: liquidations like that becomes tricky from ops point of view. as every liquidation if any bank is cb_halted needs to go through multisig. 

just a note, I don't have better idea to do that. not sure if creating additional role make sense. At least we need to prepare some scripts and process to facilitate that.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
dawid-sw: also was thinking about that

dawid-sw: so if price drifts away really strongly, the tx is still going to pass I guess

programs/marginfi/src/instructions/marginfi_account/borrow.rs
IliaZyrin: Good catch! And agreed with @dawid-sw

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IliaZyrin: I think It's okay to give risk_admin to a hot wallet temporarily, as we already did in the past. Once the bank is unhalted, we just reset it back to MS.

programs/marginfi/src/state/bank.rs
IliaZyrin: I'm not sure adding a new state just for allowing deposit and repay is worth it. Since interest accrual is anyway stopped, there is little sense in those operations. Regarding riskless withdrawals though, it's a good point. But this CircuitBroken state shouldn't be long normally.

IliaZyrin: + We still have deleveraging capability in Paused state, so we can sort some exotic/prolonged scenarios manually.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
IMEF-FEMI: Agreed the original approach was flawed. But I don't think passing all the risk accounts as mutable is a great tradeoff as it's a breaking change.
  
Instead, I've updated the non-halt-safe paths to use a read-only price gate: on every borrow / risk-withdraw / order execution, it checks each involved bank's live oracle price against the breaker reference and reverts the tx if the deviation crosses the breach threshold. That way the reference doesn't need updating there, and the risk accounts can stay read-only. The reference can still be updated using the `lending_pool_pulse_bank_price_cache` instruction

I've also made the breaker trip on the first breach rather than a sustained one, so a single-tx manipulation can't slip through before a halt.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
IMEF-FEMI: fixed this

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IMEF-FEMI: I agree with @IliaZyrin That would be cheaper than having a separate admin for this

jgur-psyops: it does require us to act quickly if e.g. BONK was to suddenly dip 90% and hit the breaker we would have to quickly scramble to empower the risk admin. I guess that's preferable to letting an attacker freely liquidate using the "bad" price so it is what it is.

type-crate/src/types/bank_config.rs
```diff
@@ -108,7 +108,17 @@ pub struct BankConfig {
     /// Stored oracle price for `OracleSetup::Fixed`, otherwise does nothing
     pub fixed_price: WrappedI80F48,
 
-    pub _padding1: [u8; 16],
+    /// Deviation thresholds in basis points for tiers 1/2/3, strictly monotonic.
```
jgur-psyops, line 118: I think some of the limits like `MAX_DEVIATION_BPS` should be in this file and documented in the comment here.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
jgur-psyops: The main issue with the revised approach is that it obligates us to pulse regularly or banks can enter a soft-lock state. Consider for example:

* Price of A was $10 last week, and that was our last pulse (no activity this week)
* T1 breaker at 20%
* Price is now $12

User goes to borrow B against A -> hits breaker on A. T1 breaker is NOT engaged because bank A doesn't mutate on a borrow for B. This continues indefinitely until someone pulses A to actually engage the breaker and waits until the halt ends or resets it. So borrow -> borrow -> borrow, will just fail forever unless the price reverts.

So here I guess the trade-off is either (1) make banks mutable and update the breaker in risk instructions directly, (2) keep the banks non-mutable and know that we must pulse banks failing due to breaker issues or they will be "stuck" in that state forever unless the price reverts.

But if we do (1) it occurs to me we must have this weird borrow ix that succeeds but does not actually trigger a borrow, since if it errors the whole tx will revert and no breaker will be set. That makes me lean towards (2) since users aren't going to sign a (1) tx anyways. So tl;dr I think this approach is fine but we must now be prepared to pulse banks that would trigger the breaker.

type-crate/src/types/bank_config.rs
```diff
@@ -108,7 +108,17 @@ pub struct BankConfig {
...
+    pub cb_deviation_bps_tiers: [u16; 3],
+    /// Halt durations in seconds for tiers 1/2/3, strictly monotonic.
+    pub cb_tier_durations_seconds: [u16; 3],
+    /// Consecutive counted breach observations required before a halt trips.
+    pub cb_sustain_observations: u8,
+    /// Escalation window multiplier: a re-breach within `prev_tier_duration * mult` seconds
+    /// after a halt ends ratchets to the next tier.
+    pub cb_escalation_window_mult: u8,
+    /// EMA smoothing factor for the reference price, in basis points (e.g. 1000 = α=0.1).
+    pub cb_ema_alpha_bps: u16,
```
jgur-psyops, line 127: Overall I think this approach is great to trigger breakers during genuine price movement. Nice work @IMEF-FEMI !

It's still not particularly robust against an attacker who has control over an oracle. Right now, the attacker can walk the price up, staying under the T1 threshold and raising the EMA. So to move from 10 dollars they would go 10.1, 10.2, 10.3.... a few seconds apart, and as long as they make small enough moves they can stay under the breaker but move the price very fast in a few seconds. They are limited to one move per slot but can basically make one move <T1 in size each slot, which would still be parabolic over a few minutes they could easy go from 10 -> 100 in a minute if T1 ~500bps or similar without tripping a breaker.

Ideas:
(1) The breaker price can only update once every e.g. minute. Now the attacker can only walk 1 unit of ~T1 every minute. If T1 is very small and short that limits them considerably. Simplest to implement but not super effective, still unbounded over longer time windows.
(2) The breaker tracks cumulative price growth and also limits that. For example, no more than 10% in 24h. Here the attacker might wait for the cumulative growth window to reset, but can basically grow the price no more than 20% cumulatively in 24h.
(3) we have a "slow price" anchor that's like 24h+ old and we also break if that gets exceeded. This is hard because it must reset on some sliding interval, otherwise the attacker will just capture it at reset time. 
(4) We punt and just keep breaker as-is to prevent economic issues only

cc @dawid-sw @IliaZyrin for other thoughts/ideas

IliaZyrin: I vote for (2) if the implementation is not much harder than (1). If it is, then (1) is also ok.

dawid-sw: I also prefer 2, then 1

IMEF-FEMI: Thanks @jgur-psyops 
 
 I also agree, the slow "staircase" attack is a real gap in the current design.
  
Options (1)–(3) all help, but they also compare the price against a reference derived from the same oracle the attacker controls... so they slow the attack down but can't really stop it.

Two ideas we could also consider:

* Consecutive-move detection: The staircase attack would only work as a long run of one direction updates. genuine market volatility retraces. We could count consecutive same direction moves and tighten the allowed deviation as the run grows (e.g. 500 → 250 → 100 → eventually 0). A tiny tick the other way shouldn't reset the counter, or the attacker would sprinkle in some free noise. we require a real retracement (say ≥50% of the accumulated run). This forces the attacker to either slow down significantly or break the price into obvious up-and-down swings that are easy to spot and cost them half their gains each time.

 * Retracement-based budget: Like cumulative-growth tracking in (2),  but instead of refilling the budget on a timer, it refills only when the price actually retraces, Upward moves spend budget, downward moves refund it.. so net displacement is hard-capped unless the attacker first gives back part of their move. There's no reset window to wait out. To still allow normal long-term price growth, the budget could also refill slowly over time e.g. ~1%/hour — on top of the retracement refills. That slow refill rate then becomes the attacker's top sustained speed, turning a 1-minute staircase into something that takes days.

IliaZyrin: This feels a bit too complex to me, I would just stick to 24hr 20% cap. Yes, it still allows the oracle-controlling attacker to manipulate the price, but 24hr is enough for us to notice such a big and persistent rise and verify whether it's genuine or not.

jgur-psyops: consecutive move probably too complex and might not work: you would have to measure the proportion of the moves, so like +1% -0.1% wouldn't reset it, and that's fairly hard to do. but also, sometimes prices genuinely do staircase upwards, doing nothing but upwards moves, and that's fine as long as they stay within the breaker.

You could do the retracement budget but it might not be possible vs N hour window without some kind of regular crank. it's basically a dynamic window which we also wanted for flow control and could not ultimately achieve within our limitations. the N time window similar to what we settled on for flow control is probably fine. here I think if N is fairly small (say 4h) then the dynamic window performs similarly anyways.

IMEF-FEMI: Agreed. I've implemented a blend of (2) alongside the existing per-observation breaker: we now track a long-window anchor price and trip the breaker if cumulative movement exceeds the configured cap before the window rolls.

programs/marginfi/src/state/bank.rs
```diff
@@ -615,15 +749,246 @@ impl BankImpl for Bank {
...
+            self.cb_window_started_at = now;
+        }
+
+        let up_delta = (current - window_ref_price - confidence).max(I80F48::ZERO);
+        let down_delta = (window_ref_price - current - confidence).max(I80F48::ZERO);
+        let window_up_bps = cb_deviation_bps(up_delta, window_ref_price);
+        let window_down_bps = cb_deviation_bps(down_delta, window_ref_price);
+        let window_deviation_bps = window_up_bps.max(window_down_bps);
+
+        if window_up_bps >= CB_WINDOW_MAX_UP_BPS || window_down_bps >= CB_WINDOW_MAX_DOWN_BPS {
```
jgur-psyops: We check the long-window breach here before we roll it. 

Let's say we have a 4 hour window and the cap is 20%:
* t=0, price = 100
* whatever samples happen for 4h, price anchor stays 100
* t =4h+1s, price = 121
* (121-100)/100 = 21%, so t2 breaker trips. The anchor never rolls up to 121 because we break right here, the anchor is still 100.
* t=5h, price = 121, the breaker trips again and escalates to t3.

I would say this is maybe fine and expected. 

To make this state a little easier to recover from without breaching into t3, we could creep the anchor up to the breach level (120 here) when that breach at 4h+1 happens if `now >= started_at + CB_WINDOW_SECONDS`. That would let an attacker creep up to 20% each window, or basically 40% at the edge of a window, but they would still hit the t2 breaker as they do, so it should be fairly noticeable. Notably they can move up 39.99% right now, without hitting T2, so I suppose it's not much of a tradeoff.

Thoughts @IMEF-FEMI @IliaZyrin @dawid-sw ?

programs/marginfi/src/state/bank.rs
```diff
@@ -33,14 +33,67 @@ use drift_mocks::constants::scale_drift_deposit_limit;
...
+/// from an attacker-controlled stale price.
+pub const CB_ENABLE_MAX_PRICE_AGE_SECONDS: i64 = 30;
+
+/// Per-pulse cap on EMA reference movement, in basis points of the current reference. Bounds
+/// how fast a sustained attacker can reanchor the EMA even at maximum α.
+pub const CB_MAX_EMA_SHIFT_BPS_PER_PULSE: u64 = 500;
+
+/// Long-window cap that catches slow oracle walking under the per-observation threshold.
+pub const CB_WINDOW_SECONDS: i64 = 4 * 60 * 60;
+pub const CB_WINDOW_MAX_UP_BPS: u64 = 2_000;
```
jgur-psyops, line 67: Can we make these configurable as well? I suppose if you have run out space at the end of BankConfig there is some padding here and there you can capture.

programs/marginfi/src/state/bank.rs
dawid-sw: +1, I like this.

I think we should work on good monitoring as well. Otherwise we risk silently accepting a manipulated price as the new baseline. At minimum, we’d need alerts on breaker trip/clear events, a quick/scripted way to verify against CEX/secondary oracle prices, pulse-health alerts, and a easy way to reseed or keep the bank halted.

programs/marginfi/src/state/bank.rs
dawid-sw: we can always resize in case of missing padding, like batch resize for around 400 banks that we have on mainnet rn is doable

programs/marginfi/src/state/bank.rs
IliaZyrin: Agreed with updating the anchor even in case of a breach AND monitoring for such events so that we can spot if it's not genuine.

programs/marginfi/src/state/bank.rs
```diff
@@ -112,6 +179,25 @@ pub trait BankImpl {
         fee_vault_authority_bump: u8,
         bank_seed: u64,
     ) -> Self;
+    #[allow(clippy::too_many_arguments)]
+    fn init(
```
jgur-psyops, line 183: One of those amusing things to be aware of: the bank::new() approach we were using earlier was blowing out the stack with those extra config bytes:
```
marginfi-v2 git:(circuit_breaker_guard_against_oracle_attack) ✗ anchor build -p marginfi -- --no-default-features
   Compiling marginfi v0.1.9 (/home/fish/mrgn/marginfi-v2/marginfi-v2/programs/marginfi)
Error: Function _ZN8marginfi12instructions14marginfi_group23add_pool_permissionless36lending_pool_add_bank_permissionless17hc2df0b6ad424dfecE Stack offset of 4120 exceeded max offset of 4096 by 24 bytes, please minimize large stack variables. Estimated function frame size: 4224 bytes. Exceeding the maximum stack offset may cause undefined behavior during execution.
```

That sort of thing just silently warns on localnet but can cause data corruption on mainnet.

Here we avoid putting a default() on the stack by just passing a mut ref to the bank, which fixes it.

IliaZyrin (approved):

PR #568, opened 2026-04-27
add extra init/write metadata instructions without seeds constraint

IMEF-FEMI:
Add extra metadata instructions without the seeds constraints 

                                                                                                                            
  - The prior strict-seed init_bank_metadata / write_bank_metadata instruction update require bank_mint + bank_seed to re-derive the bank PDA. That was added so metadata could be created before the bank exists but it forces every caller to know the exact seed/mint.   
 
  - For some common case (the bank already exists and we just want to init/write to metadata account), adding strict "seeds" constraints isn't really necessary caller should just be able to pass a live Bank account (Anchor's discriminator check on AccountLoader<Bank> already  proves it's a real bank. )
  
  - Adds init_bank_metadata_no_seed / write_bank_metadata_no_seed to restore that simpler path so callers operating on      
  existing banks don't need to track or look up seeds. Shared ticker/description copy logic is extracted into an            
  apply_metadata_write helper.


programs/marginfi/src/instructions/marginfi_group/init_bank_metadata_no_seed.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    metadata.bank = ctx.accounts.bank.key();
+    metadata.bump = ctx.bumps.metadata;
+
+    Ok(())
+}
+
+#[derive(Accounts)]
+pub struct InitBankMetadataNoSeed<'info> {
+    pub bank: AccountLoader<'info, Bank>,
```
jgur-psyops: It is somewhat annoying to have a whole extra ix just for this one constraint difference.

I think we could just use the existing `init_bank_metadata` function and remove the bank seed check: if the caller inits some meta for a non-existent PDA, that doesn't really matter anyways. Only downside is it violates the invariant that all meta have a corresponding valid bank.

We can also have the existing ix take a `pub bank: UncheckedAccount<'info>,` but then verify in the handler that EITHER (1) it's blank (owned by System program) but the seeds are valid, or (2) it's a bank (from discriminator and/or AccountLoader).

Or we can just keep this second identical-except-different-bank-constraint ix.

We should also go back and backfill `seed` to Banks at some point, I can see how the existing `init_bank_metadata` is annoying.

cc @dawid-sw @IliaZyrin for input as well.

programs/marginfi/src/instructions/marginfi_group/write_bank_metadata_no_seed.rs
```diff
@@ -0,0 +1,35 @@
...
+
+#[derive(Accounts)]
+pub struct WriteBankMetadataNoSeed<'info> {
+    #[account(
+        has_one = metadata_admin,
+    )]
+    pub group: AccountLoader<'info, MarginfiGroup>,
+
+    #[account(
+        has_one = group,
```
jgur-psyops: likewise we could have `apply_metadata_write` take a `pub bank: UncheckedAccount<'info>,`, then verify in the handler that either (1) seeds match, or (2) it's an existing bank, and group matches. Here if it's an existing bank then the `bank_seed` passed to the function is ignored and can be whatever. Depends on what we decide re: above.

programs/marginfi/src/instructions/marginfi_group/init_bank_metadata_no_seed.rs
IliaZyrin: I agree that we should not introduce new ixes here. And I think we can be fine with just removing the bank seed check. IMHO the downside is not that big.

dawid-sw: +1 - not a fan of new ix just for this

jgur-psyops: we now have an instruction to backfill seeds (ie not created with an old keypair-based add_bank ix), for the banks where seeds exist, and new banks will have seeds stored on them moving forward, so this is now simplified somewhat.

programs/marginfi/src/instructions/marginfi_group/write_bank_metadata.rs
```diff
@@ -52,16 +91,32 @@ pub fn write_bank_metadata(
...
+    #[account(mut)]
+    pub metadata_admin: Signer<'info>,
+
+    #[account(mut, has_one = bank)]
+    pub metadata: AccountLoader<'info, BankMetadata>,
+}
+
+#[derive(Accounts)]
+#[instruction(bank_seed: u64)]
+pub struct WriteBankMetadataPreInit<'info> {
```
jgur-psyops, line 112: Added an instruction to write meta before a bank exists, assuming you know the bank's canonical seed. 

@IMEF-FEMI I assume with this you would have all the tools you need to write meta while the tx to add a bank is pending on the MS?

@IliaZyrin @dawid-sw can you double-check that I am correct this cannot be trolled by an attacker (I added a small test).

IliaZyrin: LGTM

IliaZyrin (approved):

programs/marginfi/src/instructions/marginfi_group/write_bank_metadata.rs
IMEF-FEMI: looks okay @jgur-psyops  LGTM

PR #569, opened 2026-04-27
Vote accounts on-chain

jgur-psyops:
Adds validator vote accounts to on-chain bank data for Staked Collateral banks

* `lending_pool_add_bank_permissionless` now requires the validator vote account and automatically updates the bank with it.
* `lending_pool_backfill_staked_bank_validator_vote_account` permissionlessly backfills existing banks, and will be removed in 1.10


dawid-sw (approved):
LGTM

PR #570, opened 2026-04-28
Misc legacy tests

jgur-psyops:
Various legacy tests that didn't make it into their respective updates. 

* Remove the juplend reference program from our repo
* SVSP MEV tests using latest svsp .so (not quite working as expected, but demonstrate that the new instructions exist, at least)
* juplend supplementary tests
* switchboard-based tests around tx limits (added m03 scenarios)
* updates package manager in TS tests from yarn 1 and the weird transient Node 23 version to Node 24 and yarn 4


dawid-sw (approved):
lgtm

PR #573, opened 2026-04-29
Remove curve migration and emissions cleanup IX

dawid-sw:
1. Removed curve migration ix - all curves are migrated
2. Removed emissions clean up functions


jgur-psyops:
LGTM! Renamed the legacy interest curve fields to placeholder just to force all consumers to upgrade wherever they are using those.

PR #578, opened 2026-05-12
Audit fixes

dawid-sw:
- M12 — Dust clamping: Clamps tolerated sub-threshold residuals to zero on RepayOnly, DepositOnly, WithdrawOnly, and BorrowOnly, preventing shares from leaking onto the prohibited side.
- M9 - disallow acquire interest during pause (https://github.com/0dotxyz/marginfi-v2/pull/578/commits/68b6b7bee31b556cdb7f8fc90cb8de5f3015b9bf)
- L6/L7 — close_balance accounting: Decrements bank totals and position counters, and routes asset residuals to collected_insurance_fees_outstanding.
- I1 — Event signer fix: LendingPoolBankCreateEvent.signer now reports fee_payer instead of group.admin on permissionless bank creation.
- I7 — JupLend CPI defense in depth: Pins liquidity_program and cross-checks rewards_rate_model against integration_acc_1 in JuplendDeposit and JuplendInitPosition.
- I8 — Integration deposit_limit semantics documented: Adds doc comments on Kamino, Solend, and JupLend ConfigCompact.deposit_limit, clarifying that it is denominated in collateral/fToken units.
- I11 — Protocol pause on account init: MarginfiAccountInitialize and MarginfiAccountInitializePda now reject execution when the protocol is paused.
- L20 — Frozen freeze on order maintenance: CloseOrder and SetKeeperCloseFlags now use the deposit-style account_not_frozen_for_authority and is_signer_authorized constraint pair.
- L22 - prevent i64 overflow in rate limiter


p0-cli/src/processor/account.rs
```diff
@@ -731,6 +731,11 @@ pub fn marginfi_account_close_order(
     let authority = config.authority();
     let marginfi_account_pk = profile.get_marginfi_account()?;
 
+    let marginfi_account = config
```
IliaZyrin, line 734: The result (`marginfi_account`) is not used.

p0-cli/src/processor/account.rs
```diff
@@ -1023,6 +1029,11 @@ pub fn marginfi_account_set_keeper_close_flags(
 ) -> Result<()> {
     let marginfi_account_pk = profile.get_marginfi_account()?;
 
+    let marginfi_account = config
```
IliaZyrin, line 1033: The result (`marginfi_account`) is not used.

IliaZyrin (approved):

PR #584, opened 2026-05-18
fix bank rate limit overflow bug

IMEF-FEMI:
### Problem
  `RateLimitWindow` accepts `u64` amounts but compared and accumulated them via
  unchecked `as i64` casts. An amount above `i64::MAX` wraps to a negative `i64`,
  so `amount as i64 > remaining` reads false and the limiter **fails open** — and
  `cur_window_outflow` is corrupted downward. Reachable on a high-decimal bank
  (e.g. an 18-decimal token: a 10-token outflow exceeds `i64::MAX` native units).

  ### Fix
  - All `u64 → i64` conversions now go through `amount_as_i64` (`i64::try_from`),
    which returns `None` instead of wrapping.
  - **Outflows** above `i64::MAX` are rejected (an un-representable amount is
    treated as exceeding every window).
  - **Inflows** above `i64::MAX` are clamped to the max representable credit —
    deliberately asymmetric, so a legitimate large deposit isn't trapped behind
    the outflow cap.
  - `remaining_capacity_from_state` weighting rewritten in `i128`, removing the
    `unsigned_abs`/sign-reapply path that could panic on `i64::MIN`.
  - Configured caps and admin-supplied inflows are bounds-checked at the
    instruction boundary (`configure_rate_limits`, `update_group_rate_limiter`)
    via `is_valid_rate_limit_amount`; oversized/corrupted state fails closed.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/configure_rate_limits.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/configure_rate_limits.rs
@@ -1,6 +1,7 @@
 use crate::{
+    check,
     state::marginfi_group::MarginfiGroupImpl,
-    state::rate_limiter::{BankRateLimiterImpl, GroupRateLimiterImpl},
+    state::rate_limiter::{is_valid_rate_limit_amount, BankRateLimiterImpl, GroupRateLimiterImpl},
     MarginfiError, MarginfiResult,
 };
 use anchor_lang::prelude::*;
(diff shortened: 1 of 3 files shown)


dawid-sw:
good job @IMEF-FEMI, I added fix [here](https://github.com/0dotxyz/marginfi-v2/pull/578) - but prefer yours as you added some nice checks and tests, so we can merge these two 🔥

IMEF-FEMI:
Thanks @dawid-sw  🚀🚀🚀🚀 Glad i could help

PR #585, opened 2026-05-18
Same asset emode internal

jgur-psyops:
Thanks to @abelmarnk for starting the feature (see #555)

### Summary

Adds same-asset eMode: a new risk mode enabling banks that share the same mint to automatically gain emode benefits, with configurable leverage limits


programs/marginfi/src/state/emode.rs
```diff
@@ -19,6 +21,14 @@ pub const DEFAULT_INIT_MAX_EMODE_LEVERAGE: I80F48 = I80F48!(15);
 /// A value of 20 means positions can theoretically leverage up to 20x through recursive borrowing.
 pub const DEFAULT_MAINT_MAX_EMODE_LEVERAGE: I80F48 = I80F48!(20);
 
+/// Default maximum allowed same-asset leverage for group initialization (initial).
+/// This is set to 2x the regular initial emode default leverage.
+pub const DEFAULT_INIT_MAX_SAME_ASSET_EMODE_LEVERAGE: I80F48 = I80F48!(30);
```
dawid-sw: Same-asset e-mode is enabled by default for new groups at 30x / 40x, and I think the safety here relies on price similarity of two mints. The concept only seems safe if both banks price the mint identically, but each bank has its own oracle config, and nothing enforces that the two feeds match. The tests reuse the same default bank config for both banks, so this case is not really covered. At 40x maintenance, the margin buffer is only around 2.5%. A stale or divergent oracle feed, wide confidence interval, or borrow/supply interest spread could eat through that buffer before the account becomes cleanly liquidatable. Since the leverage is group-wide, the same buffer applies regardless of the asset’s liquidity or oracle quality. So this is real risk that we should discuss. 

ofc this defaults doesn't matter much as our main group is already deployed and to enable this feature we should modify group config, but still defaults should be lower or even 1x, so forks do not accidentally shoot themselves in the foot with this.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -90,6 +92,30 @@ pub fn configure(
...
+    validate_and_apply_emode_leverage(
+        same_asset_emode_maint_leverage,
+        &mut marginfi_group.same_asset_emode_maint_leverage,
+    )?;
+
+    let same_asset_init_leverage = u32_to_basis(marginfi_group.same_asset_emode_init_leverage);
+    let same_asset_maint_leverage = u32_to_basis(marginfi_group.same_asset_emode_maint_leverage);
+
+    let same_asset_disabled =
+        same_asset_init_leverage <= I80F48::ONE && same_asset_maint_leverage <= I80F48::ONE;
```
dawid-sw: worth to document that to disable same asset leverage, we need to insert `1`, as configuring number `< 1` like `0` will error-out in `validate_and_apply_emode_leverage`. when I was checking it 1st it was contrintuitive for me, usually to disable we are setting `0`.

programs/marginfi/src/instructions/marginfi_group/configure.rs
```diff
@@ -90,6 +92,30 @@ pub fn configure(
...
+
+    if !same_asset_disabled && same_asset_init_leverage >= same_asset_maint_leverage {
```
dawid-sw: The ordering check only rejects init ≥ maint, so an asymmetric init =  (disabled) / maint > 1 (enabled) config slips through and silently produces a same-asset boost that's off at open but on at maintenance/liquidation - fix by also requiring (init > 1) == (maint > 1)

programs/marginfi/src/state/emode.rs
dawid-sw: I'm voting for having some flag on banks like `BANK_FLAG_SAME_ASSET_EMODE_ELIGIBLE`, so we can control which banks can use this mode and mitigate some weird scenarios.

dawid-sw (changes requested):
some comments and point for discussion

dawid-sw:
Looked into this as a possible edge case, but I don’t think it’s exploitable on the permissionless liquidation path.

A receiver can clear a secondary liability mint mid-receivership and flip the same-asset boost on, which improves maintenance health. But end_liquidation still has boost-blind caps:

HealthyAccount prevents liquidating into positive health
`seized <= repaid * 1.05 is checked in equity terms with the boost off`

So the boosted health doesn’t turn into extra value extraction.

The only uncapped path is permissioned `end_deleverage`, but that already relies on the trusted risk-admin role, so this feature doesn’t really change that risk.

dawid-sw:
+ also worth some backward compatibility issues with this group account added in several IXes

programs/marginfi/src/instructions/marginfi_group/configure.rs
IliaZyrin: Agreed with Dawid, don't see why we would want to allow this behavior.

For the fix, we can also just inverse the logic for simplicity and use it like:
```
let same_asset_enabled =
        same_asset_init_leverage > I80F48::ONE && same_asset_maint_leverage > I80F48::ONE;
if same_asset_enabled && same_asset_init_leverage >= same_asset_maint_leverage {
```

programs/marginfi/src/state/emode.rs
IliaZyrin: Actually, the buffer for price divergence is even worse:
```
Example for 30x init / 40x maint e-mode:

init CW/LW  = 1 - 1/30 = 29/30 = 0.9666667
maint CW/LW = 1 - 1/40 = 39/40 = 0.975

If user maxes out init leverage with $100 equity:

equity  = $100
collateral value = $3,000
liability value  = $2,900
liquidatability threshold = 2900 / 0.975 = ~$2974

So it's less than 1% of allowed price divergence before the user gets liquidated:
1 - 2974 / 3000 = 0.87%

The opposite confidence biases will easily be bigger than that.
```

I think we should just stick to the regular 15/20.

programs/marginfi/src/lib.rs
```diff
@@ -45,6 +45,8 @@ pub mod marginfi {
         new_risk_admin: Option<Pubkey>,
         emode_max_init_leverage: Option<WrappedI80F48>,
         emode_max_maint_leverage: Option<WrappedI80F48>,
+        same_asset_emode_init_leverage: Option<WrappedI80F48>,
```
IliaZyrin, line 52: I [still](https://github.com/0dotxyz/marginfi-v2/pull/555#discussion_r3221975048) don't think we need a separate pair of values for same-asset emode.

jgur-psyops: This discussion link just takes me to the PR: do you mean that `EmodeConfigIterator` should be combined into one thing to avoid the double-iteration?

programs/marginfi/src/state/emode.rs
jgur-psyops: Banks must now be opted into the feature, and there is some rudimentary checking to see if they at least have the same price oracles (keys[0]). The default is now 1x.

programs/marginfi/src/lib.rs
IliaZyrin: Oh, it seems link is indeed useless, thought GitHub would navigate to the specific comment. The point is that I think we can get away without these new fields: `same_asset_emode_init_leverage` and `same_asset_emode_maint_leverage` and just use existing `emode_max_init_leverage` and `emode_max_maint_leverage` respectively.

programs/marginfi/tests/user_actions/same_asset_emode/regular.rs
```diff
@@ -0,0 +1,919 @@
...
+    .await?;
+    user.try_bank_borrow(
+        user_sol_borrow_b.key,
+        &sol_bank_c,
+        matching_liability_bank_c_ui,
+    )
+    .await?;
+
+    // Phase 1: regular weight is the largest input for the matching SOL collateral.
+    // Init SOL deposit contribution = 5 SOL * $10 * max(0.81, 0.79, 0.75) = $40.5.
```
IliaZyrin, line 369: Where does `0.75` come from?

programs/marginfi/tests/user_actions/liquidate.rs
```diff
@@ -14,6 +14,27 @@ use solana_program_test::*;
 use solana_sdk::clock::Clock;
 use test_case::test_case;
 
+async fn disable_same_asset_emode_for_liquidation_tests(
```
IliaZyrin: This shouldn't be necessary I think.

tests/utils/user-instructions.ts
```diff
@@ -410,11 +410,15 @@ export const startLiquidationIx = (
...
+  const [liquidationRecord] = deriveLiquidationRecord(
+    program.programId,
+    args.marginfiAccount,
+  );
   return program.methods
     .startLiquidation()
     .accounts({
       marginfiAccount: args.marginfiAccount,
-      // liquidationRecord: // implied from account
+      liquidationRecord,
```
IliaZyrin, line 421: Was it failing for you without it?

tests/k20_sameAssetEmode.spec.ts
```diff
@@ -0,0 +1,1106 @@
...
+    );
+
+  const tightenSameAssetLeverage = async (options?: {
+    newRiskAdmin?: PublicKey;
+  }) =>
+    configureSameAssetLeverage(
+      SAME_ASSET_TIGHTENED_INIT_LEVERAGE,
+      SAME_ASSET_TIGHTENED_MAINT_LEVERAGE,
+      options
+    );
```
IliaZyrin, line 286: These can be factored out to some utils, since they're shared between d16 and jlr08.

tests/zb01_bankruptcy.spec.ts
```diff
@@ -76,6 +82,21 @@ describe("Bank bankruptcy tests", () => {
     banks = result.banks;
     throwawayGroup = result.throwawayGroup;
 
+    // These suites cover classic eMode math; disable same-asset leverage explicitly so the
+    // tests are not affected
+    const disableSameAssetTx = new Transaction().add(
```
IliaZyrin: It seems unnecessary in this test suite, not seeing it dealing with emode.

type-crate/src/types/same_asset_emode_registry.rs
```diff
@@ -0,0 +1,113 @@
...
+assert_struct_size!(SameAssetEmodeGroup, 96);
+assert_struct_align!(SameAssetEmodeGroup, 1);
+#[repr(C)]
+#[cfg_attr(feature = "anchor", zero_copy)]
+#[cfg_attr(not(feature = "anchor"), derive(PartialEq, Pod, Zeroable, Copy, Clone))]
+#[derive(Debug)]
+pub struct SameAssetEmodeGroup {
+    /// Representative bank for this `(mint, oracle_key)` grouping. Typically, this is whichever
+    /// bank was added to the group first. The bank's mint = mint and oracle_keys[0] = oracle_key.
+    pub bank: Pubkey,
```
IliaZyrin: What is the purpose of having this representative? We anyway have a full list in the registry.

IliaZyrin (approved):
LGTM, just a few minor things

tests/utils/user-instructions.ts
IMEF-FEMI: Yes it is actually. If Left to be autofilled by anchor, the batched init + start liquidation tx fails at StartLiquidation with AccountOwnedByWrongProgram (3007): Anchor resolves liquidation_record from the marginfi_account has_one, but that field isn’t set yet at that time, so it picks the wrong account. Explicit PDA derivation is required.

If we separate the txs though, it works

IliaZyrin (approved):
LGTM

PR #588, opened 2026-05-20
Upgrate to Solana 3

IliaZyrin:
- Migrates toolchain/deps to Rust 1.90.0, Anchor 1.0.2, Agave/Solana 3.1.11, and updated SPL/Solana v3 crates.
- Moves TS tests to LiteSVM, refactors shared TS test setup/utilities, especially Kamino reserve setup. Tightens and centralizes TS numeric conversions/comparisons for BN, I80F48, Kamino U68F60, and Linux-safe precision handling.
- Stabilizes Drift/Kamino/Solend/JupLend/staked/emissions tests under the new runner and dependency stack.
- Updates fuzz harness/dependencies for the new Solana/Anchor stack.
- Refreshes CI/build scripts, caches, docs, and test instructions for the new toolchain and LiteSVM workflow.


.github/workflows/test.yaml
```diff
@@ -234,7 +248,7 @@ jobs:
         set +e
 
         run_tests() {
-          anchor run all-tests 2>&1 | tee test_output.log
```
jgur-psyops, line 237: this should probably stay all-tests iirc it skips some weird connection step that can throw occasionally (at least locally)

IliaZyrin: Yeah, sure, I'm just trying to find the mismatch at the moment. Then will discard all unnecessary changes!

PR #591, opened 2026-05-28
Add uninitialized bank state

dawid-sw:
## Summary

`juplend_init_position` previously used `operational_state == Paused` as its one-shot init gate.

Admin emergency-pause also sets the state to `Paused`, so after the admin paused a live JupLend bank, any signer could re-run `juplend_init_position` with a 10-unit seed deposit and flip the bank back to `Operational`, defeating the pause.

## Fix

Introduce `BankOperationalState::Uninitialized` as the bank's pre-activation state:

- `JuplendConfigCompact::to_bank_config` now sets `operational_state = Uninitialized` instead of `Paused`.
- `juplend_init_position` now requires `operational_state == Uninitialized`; on success, it transitions the bank to `Operational`.
- `Bank::configure` rejects setting state to `Uninitialized`, joining the existing `KilledByBankruptcy` rejection, so admin cannot re-arm init via `lending_pool_configure_bank`.
- `validate_bank_state` rejects `Uninitialized` for every instruction kind with new error `BankUninitialized` (`6134`).

Lifecycle is now explicit:

```text
Uninitialized → Operational ⇄ Paused ⇄ ReduceOnly → KilledByBankruptcy

### DIFF
--- a/programs/marginfi/src/instructions/juplend/init_position.rs
+++ b/programs/marginfi/src/instructions/juplend/init_position.rs
@@ -18,12 +18,13 @@ use marginfi_type_crate::types::{Bank, BankOperationalState};
 /// This instruction:
 /// 1. Transfers a small amount of underlying from the fee payer into the bank liquidity vault
 /// 2. Performs a `deposit` CPI into JupLend (seed deposit)
-/// 3. Flips the bank operational state from `Paused` -> `Operational`
+/// 3. Transitions the bank from `Uninitialized` to `Operational`
 ///
 /// Notes:
 /// - The fToken vault is created during `add_pool`, not here.
 /// - The seed deposit is intentionally not credited to any marginfi account.
-/// - The bank is created in `Paused` state; this instruction is what activates it.
+/// - `Uninitialized` is unreachable from `lending_pool_configure_bank`, so an admin pause cannot
+///   re-arm this instruction.
 pub fn juplend_init_position(ctx: Context<JuplendInitPosition>, amount: u64) -> MarginfiResult {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: programs/marginfi/src/instructions/juplend/init_position.rs [hunk 1]
```diff
@@ -18,12 +18,13 @@ use marginfi_type_crate::types::{Bank, BankOperationalState};
...
 /// 1. Transfers a small amount of underlying from the fee payer into the bank liquidity vault
 /// 2. Performs a `deposit` CPI into JupLend (seed deposit)
-/// 3. Flips the bank operational state from `Paused` -> `Operational`
+/// 3. Transitions the bank from `Uninitialized` to `Operational`
 ///
 /// Notes:
 /// - The fToken vault is created during `add_pool`, not here.
 /// - The seed deposit is intentionally not credited to any marginfi account.
-/// - The bank is created in `Paused` state; this instruction is what activates it.
+/// - `Uninitialized` is unreachable from `lending_pool_configure_bank`, so an admin pause cannot
```
jgur-psyops, line 26: I suppose this means you can soft-brick a juplend bank by
* add_bank
* configure_bank to something other than `Uninitialized`
* now juplend_init_position can never run.

I suppose if we are being very careful and pedantic we want to block configure_bank when  asset tag is juplend and trying to configure op state away from Uninitialized.

I'll apply a tiny patch.

dawid-sw: good point, I agree and patch looks ok

PR #595, opened 2026-06-02
Svsp fix

IliaZyrin:


programs/marginfi/src/state/price.rs
```diff
@@ -90,6 +88,68 @@ fn check_primary_oracle_key(
...
+    check!(
+        bank.integration_acc_1 != Pubkey::default(),
+        MarginfiError::StakePoolValidationFailed
+    );
+
+    Ok(derive_staked_onramp_from_vote(bank.integration_acc_1))
+}
+
+// To be removed once SVSP update is rolled out (likely in 1.10)
+fn staked_oracle_uses_onramp(bank_config: &BankConfig) -> bool {
```
jgur-psyops, line 115: Not sure we'll be able to coordinate this precisely, a dedicated attacker could slide in before we enable the flag. I think we would have to hit the flag right as SVSP program upgrades which is probably not possible. Perhaps we go in two stages, assuming they will give us a day and time, we can set some flag that puts it in a "panic on price" state which essentially bricks it, then this flag for when the upgrade is done. For example if they tell us upgrade will be around 2pm ET on Monday then we do
```
Monday 1:50pm - all staked banks set to panic mode
Monday 2ish - svsp updates
Monday 2:05 - all staked banks set to uses_onramp and panic mode released
```

I could be wrong about the implications here, it may also be safe to:
```
Monday 1:50pm - all staked banks set to uses_onramp
Monday 2ish - svsp updates
```
Assuming that the uses_onramp state is self-bricking before the svsp upgrade completes.

IliaZyrin: I thought we agreed to simply pause the banks during the rollout? And then only unpause with the flag set.

PR #596, opened 2026-06-03
Record Kamino/Solend withdraw rate-limit outflow in underlying liquidity units

IMEF-FEMI:
* Summary
   Small fix to record Kamino / Solend withdraw rate-limit outflow in underlying native liquidity units (via collateral_to_liquidity), instead of raw collateral-share amounts.

 * Why
  The bank rate limiter tracks net outflow, netting deposit inflows against withdraw outflows in one shared window, so both  sides must use the same unit. (They previously didn't)


dawid-sw (approved):
LGTM

programs/marginfi/src/instructions/kamino/withdraw.rs
```diff
@@ -157,9 +151,15 @@ pub fn kamino_withdraw<'info>(
             amount
         };
 
+        let native_outflow = ctx
```
jgur-psyops: does it matter that we are calling this before the reserves refresh around line 196?   I think the price could change as the reserve accrues and the exchange rate updates, iirc this is why the receivership price cache update executes after the refresh.

jgur-psyops: moved it after accrue

IliaZyrin (approved):

PR #597, opened 2026-06-03
Upgrade to solana 3

IliaZyrin:
- Migrates toolchain/deps to Rust 1.90.0, Anchor 1.0.2, Agave/Solana 3.1.11, and updated SPL/Solana v3 crates.
- Moves TS tests to LiteSVM, refactors shared TS test setup/utilities, especially Kamino reserve setup. Tightens and centralizes TS numeric conversions/comparisons for BN, I80F48, Kamino U68F60, and Linux-safe precision handling.
- Stabilizes Drift/Kamino/Solend/JupLend/staked/emissions tests under the new runner and dependency stack.
- Updates fuzz harness/dependencies for the new Solana/Anchor stack.
- Refreshes CI/build scripts, caches, docs, and test instructions for the new toolchain and LiteSVM workflow.


p0-cli/src/config.rs
```diff
@@ -71,7 +69,7 @@ pub enum CliSigner {
 }
 
 pub fn clone_keypair(keypair: &Keypair) -> Keypair {
-    Keypair::from_bytes(&keypair.to_bytes()).unwrap()
+    keypair.insecure_clone()
```
dawid-sw, line 72: worth to think about some refactor later that do not require cloning

Cargo.toml
```diff
@@ -42,14 +54,12 @@ mocks = { path = "programs/mocks", package = "mocks" }
 id-crate = { path = "id-crate" }
 marginfi-type-crate = { path = "type-crate" }
 
-pyth-solana-receiver-sdk = "0.6.0"
-# pythnet-sdk = "2.3.1"
-switchboard-on-demand = "0.3.5"
+pyth-solana-receiver-sdk = { git = "https://github.com/IliaZyrin/pyth-crosschain", package = "pyth-solana-receiver-sdk", rev = "913ce2196a104635582d745c8fbd9a69e240070c" }
```
dawid-sw: the better option would be creating fork in p0 organization, so we can have more control over it and mitigate supply chain type of attacks.

dawid-sw (commented):
good work! just some small comment. 

next big step for us is probably to move out totally from typescript and have integration tests in liteSVM Rust tests

p0-cli/src/config.rs
dawid-sw: ofc not now, later

.github/workflows/test.yaml
```diff
@@ -14,10 +14,9 @@ defaults:
     working-directory: .
 
 env:
-  RUST_TOOLCHAIN: 1.79.0
-  SOLANA_CLI_VERSION: 2.1.20
-  ANCHOR_CLI_VERSION: 0.31.1
-  ANCHOR_SHA: e6d7dafe12da661a36ad1b4f3b5970e8986e5321
+  RUST_TOOLCHAIN: 1.90.0
+  SOLANA_CLI_VERSION: 3.1.11
```
dawid-sw: can we try to have it 3.1.13? Trident fuzz suite requires this specific version, shouldnt make much diff here

IliaZyrin: Yes, definitely! I was just juggling different versions to cope with dependency hell and stopped as soon as it got stable, so no particular reason to stick to 3.1.11.

Cargo.toml
IliaZyrin: Agreed, I actually don't like depending on these custom versions at all. So I hope eventually my PRs to those 3rd-party repos will be merged, but at least we stick to a very specific commit here.

dawid-sw (approved):
LGTM, good work

PR #598, opened 2026-06-04
Svsp fix

IliaZyrin:
Adds logic for a smooth transition to a [new staked pricing model accounting for on-ramp](https://github.com/solana-program/single-pool/pull/601). To be rolled out on mainnet soon.

### The transition will happen in two phases:
1. Slightly before the SVSP upgrade the main group admin will set a newly introduced flag on the group level - `STAKED_ORACLE_DISABLED` - which will temporarily block all protocol interactions with the P0 stake banks. This is needed to safely switch the pricing logic, so that nobody can trigger/benefit from a sudden price drop due to the improper math.
2. Once the SVSP upgrade is complete, the admin will set another flag - `STAKED_ORACLE_PRICE_USES_ONRAMP` - which persists the actual math and at the same times restores stake banks' functionality on the platform.

This approach - using group-level flags - also guarantees that any new potentially added stake banks will follow the same rules as the already present ones.

### New instructions:
- `disable_staked_oracles` (admin only)
- `enable_staked_oracle_onramp` (admin only)

### BREAKING CHANGES for a few existing instructions:
`lending_account_end_flashloan`, `start_liquidation`, `end_liquidation` and `lending_account_pulse_health` now all take a new account - `group`.

New Rust and TS tests are implemented for both the logic switch and for the new math itself.


programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
```diff
@@ -219,6 +229,8 @@ pub struct StartLiquidation<'info> {
     )]
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
 
+    pub group: AccountLoader<'info, MarginfiGroup>,
```
dawid-sw, line 232: downside is that we are now adding new account to start_liquidation which is breaking change, and then after svps turned on this group acct will nto be required, so we are going to do round-trip with changing the liquidation ix acc schemas

dawid-sw: we can keep this info in bank flag, but then it can be easily dysynchornized so we need to make sure to migrate all banks and have proper defaults on bank creation. so also not ideal pattern

programs/marginfi/src/lib.rs
```diff
@@ -117,6 +117,21 @@ pub mod marginfi {
         marginfi_group::lending_pool_backfill_staked_bank_validator_vote_account(ctx)
     }
 
+    /// (admin only) Disable stake pricing, i.e. effectively forbidding all operations involving stake banks.
+    /// To be used during the rollout of the SVSP upgrade.
+    /// To be removed once SVSP update is rolled out (likely in 1.10)
+    pub fn disable_staked_oracles(ctx: Context<DisableStakedOracles>) -> MarginfiResult {
```
dawid-sw, line 123: potentially can be just 1 ix with param but as this is temp, doesnt matter that much

type-crate/src/types/group.rs
```diff
@@ -91,9 +96,28 @@ pub struct MarginfiGroup {
...
+    OnRampEnabled,
+}
+
 impl MarginfiGroup {
     pub const LEN: usize = std::mem::size_of::<MarginfiGroup>();
     pub const DISCRIMINATOR: [u8; 8] = discriminators::GROUP;
+
+    // To be removed once SVSP update is rolled out (likely in 1.10)
+    pub fn on_ramp_transition(&self) -> OnRampTransition {
+        if self.group_flags & STAKED_ORACLE_PRICE_USES_ONRAMP != 0 {
```
dawid-sw, line 113: can you confirm this logic? so when we enable this feature we cannot disable it? as STAKED_ORACLE_DISABLED is checked in second branch after checking STAKED_ORACLE_PRICE_USES_ONRAMP flag. so flipping STAKED_ORACLE_DISABLED does nothing if this feature will be enabled at least once?

dawid-sw (changes requested):
some stuff to discuss

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
IliaZyrin: Yeah, initially it was a per-bank flag implementation, but then we realized it would require too many signatures and won't protect new (permissionlessly created) staked banks. Using flag in staked settings would help with the new banks, but would still require propagation to all existing, which is still a lot of signing.

So the group-wide flag was chosen as both simple to execute and safe in terms of auto-sync for all new banks.

Not ideal though, I agree. Although for TS-clients it should not be a breaking change since the account is derived by Anchor. For Rust, what we can do is to NOT reverse this change even after the SVSP upgrade and keep demanding groups in all those affected ixes. Redundant, but less hectic in terms of breaking changes. I would still prefer to reverse, but just as an option.

programs/marginfi/src/lib.rs
IliaZyrin: Agreed, I just opted for this for more clarity -> easier execution.

type-crate/src/types/group.rs
IliaZyrin: Yes, that's by design. I thought of it as a prevention mechanism against accidental re-enabling of DISABLED state by running some wrong script/setting the wrong flag in another way.

Do you think we should allow getting back to DISABLED?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -371,22 +371,7 @@ fn calc_weighted_asset_value_cached_standalone(
...
-                    RequirementType::Initial => I80F48::from(emode_entry.asset_weight_init),
-                    RequirementType::Maintenance => I80F48::from(emode_entry.asset_weight_maint),
-                    RequirementType::Equity => I80F48::ONE,
-                };
-                max(bank_weight, emode_weight)
-            } else {
-                bank.config
-                    .get_weight(requirement_type, BalanceSide::Assets)
-            };
+            let mut asset_weight = bank.get_asset_weight(requirement_type, emode_config);
```
IliaZyrin, line 374: NOTE: Emode-related changes are all in [a single commit](https://github.com/0dotxyz/marginfi-v2/pull/598/changes/878fab5bd8898dec866630e2f58f90d9c2dee055) (planned to create a separate PR for it but just pushed here to not create more conflicts). Those are for Eva mostly.

programs/marginfi/src/instructions/marginfi_account/liquidate_start.rs
dawid-sw: okay, makes sense

type-crate/src/types/group.rs
dawid-sw: I just observed this inconsistency and like to make sure its intended behaviour. I think its fine for now, but we will fix that after svps migration as part of cleanup.

IMEF-FEMI (approved):
Looks good from here

LGTM 🚀

jgur-psyops (approved):
LGTM! Great work!

PR #599, opened 2026-06-09
Resolve todos

IMEF-FEMI:
This PR clears the inline TODO/FIXME backlog across the program, tests, type-crate, and infra, and deletes the stale `todo.md` launch roadmap.

Also included:

**CI: fix the silent multi-hour "Anchor LiteSVM Tests" hang.** A V8 JIT bug made bignumber.js spin forever on k08's ~100-digit operand; the math is now bigint + double. Test randomness is seeded for reproducible runs, and the job is capped at 40 minutes.

**CLI: fix Kamino oracle derivation (merged from `cli-kamino-oracle-fix`).** Bank-creation previously mis-derived the reserve oracle setup; the CLI now loads the reserve via the complete Kamino IDL (off-chain only, `cfg(not(target_os = "solana"))`) and applies the reserve-oracle override to the correct slot of the oracle setup. (credits: @IliaZyrin)


p0-cli/src/commands/kamino.rs
```diff
@@ -103,61 +102,97 @@ struct KaminoHarvestDerivedAccounts {
...
+            Some(reserve_oracle),
+        )
+    } else if switchboard_price_oracle.is_some() || switchboard_twap_oracle.is_some() {
+        (
+            pyth_oracle,
+            Some(reserve_oracle),
+            switchboard_twap_oracle,
+            scope_prices,
+        )
+    } else {
```
IliaZyrin: You can remove the first `if` if it's treated as default anyway.

p0-cli/src/commands/kamino.rs
```diff
@@ -103,61 +102,97 @@ struct KaminoHarvestDerivedAccounts {
...
-        reserve_oracle_override
-            .or((bank.config.oracle_keys[0] != Pubkey::default())
-                .then_some(bank.config.oracle_keys[0]));
-    let (pyth_oracle, scope_prices) = match bank.config.oracle_setup {
-        OracleSetup::KaminoPythPush => (reserve_oracle, None),
-        OracleSetup::KaminoSwitchboardPull => (None, reserve_oracle),
-        _ => (None, None),
-    };
+    let (pyth_oracle, switchboard_price_oracle, switchboard_twap_oracle, scope_prices) =
+        apply_reserve_oracle_override(get_oracle_setup(&reserve_state), reserve_oracle_override);
```
IliaZyrin: What is the use case for such override? I'm not sure it will work with Kamino like this.

programs/kamino-mocks/src/state.rs
```diff
@@ -188,11 +188,27 @@ impl MinimalReserve {
             - pending_referrer_fees
     }
 
+    /// Returns true if this reserve has not been refreshed in `current_slot`.
+    ///
+    /// we call this on-chain (in `ensure_kamino_reserve_fresh`) to require that a Kamino
```
IliaZyrin:
```suggestion
    /// We call this on-chain (in `ensure_kamino_reserve_fresh`) to require that a Kamino
```

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -24,15 +24,26 @@ use anchor_spl::token_interface::{TokenAccount, TokenInterface};
...
+/// Converts a per-bank liquidation fee in basis points to an I80F48 fraction. A 0 value falls back
+/// to `DEFAULT_LIQUIDATION_FEE_BPS`
+fn liquidation_fee_fraction(bps: u16) -> I80F48 {
+    let bps = if bps == 0 {
+        DEFAULT_LIQUIDATION_FEE_BPS
+    } else {
+        bps
+    };
+    I80F48::from_num(bps) / I80F48::from_num(10_000)
+}
```
IliaZyrin: Nitpick but you could simply return the default in I80F48 here, to avoid redundant division:

```suggestion
fn liquidation_fee_fraction(bps: u16) -> I80F48 {
   if bps == 0 {
        DEFAULT_LIQUIDATION_FEE // const of I80F48 type
    } else {
        I80F48::from_num(bps) / I80F48::from_num(10_000)
    }
}
```

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -222,17 +222,26 @@ pub struct EndLiquidation<'info> {
     pub global_fee_wallet: UncheckedAccount<'info>,
 
     pub system_program: Program<'info, System>,
+
+    /// Optional separate payer for the flat liquidation fee. When provided it must sign and pays the
+    /// fee; when omitted, the `liquidation_receiver` pays (the default).
+    #[account(mut)]
+    pub fee_payer: Option<Signer<'info>>,
```
IliaZyrin, line 239: I'm not sure this feature is particularly useful.
What do you think guys?
@jgur-psyops @dawid-sw

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -985,7 +985,13 @@ pub fn check_account_bankrupt<'info>(
...
     health_cache: &mut Option<&mut HealthCache>,
 ) -> MarginfiResult {
-    // TODO remove this check here and raise it to the top-level instruction
+    // Reject bankruptcy checks while a flashloan is open. During a flashloan the account is allowed
+    // to be temporarily insolvent (collateral isn't posted yet), so it can look bankrupt while it
+    // actually isn't. Without this, an attacker could open a flashloan and have their own solvent
+    // account "bankrupted" — which seizes the position and socializes its losses to depositors.
+    //
+    // This lives in the shared check, not in each instruction (as an old TODO suggested), so callers
+    // can react differently: `handle_bankruptcy` fails on this error, `pulse_health` just records it.
```
IliaZyrin: I think we can still move it to the ix level, to `handle_bankruptcy` in particular. `pulse_health` is not that important to keep it here, plus I don't think anyone actually uses it during flashloans.

programs/marginfi/src/state/price.rs
```diff
@@ -1409,10 +1404,47 @@ impl PriceAdapter for SwitchboardPullPriceFeed {
...
+///   `i128:128` entry, so it inherits `i64`'s 8-byte alignment), and account data sits at an
+///   8-aligned offset, so upstream's zero-copy `PullFeedAccountData::parse` succeeds. We use it to
+///   avoid copying ~3.2 KB onto the 4 KB BPF stack on every price read.
+/// * Off-chain (host / `client`): native `i128` is 16-byte aligned, so the data at offset 8 is
+///   misaligned and `parse`'s `bytemuck::try_from_bytes` fails with `AccountDeserializeError`. We
+///   copy via `parse_swb_ignore_alignment` (`try_pod_read_unaligned`), which has no alignment
+///   requirement.
+///
+/// Verified alignment-sensitive upstream through switchboard-sdk `main` and the pinned rev f1a570a
+/// (2026-06-01) — `parse` still returns a `Ref` via `try_from_bytes`, so the off-chain copy stays.
```
IliaZyrin, line 1505: Should be reported to switchboard, they seem to be willing to improve things like that (better to prepare a PR ourselves though, just not sure how hard it is).

programs/marginfi/src/lib.rs
```diff
@@ -799,11 +801,16 @@ pub mod marginfi {
         )
     }
 
-    // TODO deprecate and incorporate this functionality into forced-withdraw in 1.7+
-    /// (risk admin only) Purge a user's lending balance without withdrawing anything. Only usable
-    /// after all the debt has been settled on a bank in deleveraging mode, e.g. when
-    /// `TOKENLESS_REPAYMENTS_ALLOWED` and `TOKENLESS_REPAYMENTS_COMPLETE`. used to purge remaining
-    /// lending assets in a now-worthless bank before it is fully sunset.
+    /// (risk admin only) DEPRECATED: this functionality has been folded into the forced-deleverage
```
IliaZyrin: The planned "force-withdraw" feature was a bit different, though. It would withdraw the funds to the user's ATA, not to risk_admin's one. Correct me if I'm wrong @jgur-psyops 

But anyway, good point about `withdraw_all` during deleverage, seems we don't need this purge ix anymore.

programs/marginfi/tests/admin_actions/setup_bank.rs
```diff
@@ -1600,17 +1600,14 @@ async fn configure_bank_interest_only_success() -> anyhow::Result<()> {
 
     let bank_after: Bank = test_f.load_and_deserialize(&bank.key).await;
 
-    // TODO deprecate in 1.7
     assert_eq!(
```
IliaZyrin, line 1707: I think these TODOs should stay for now, until we actually deprecate all legacy-curves-related functionality. Can be actualized, though!

type-crate/src/types/emode.rs
```diff
@@ -182,11 +182,29 @@ impl EmodeEntry {
 /// - Result
 /// * tag | flags | init | maint
 /// * 101    0       60     75
+///
+/// # Compute cost
```
IliaZyrin, line 243: Great that we finally did this benchmarking. Looking at the results, I don’t think we need to keep the classic version around anymore.

I already removed it [on my branch](https://github.com/0dotxyz/marginfi-v2/pull/598/changes/878fab5bd8898dec866630e2f58f90d9c2dee055) and used the non-classic version from Eva, who is a typical client.

programs/mocks/src/lib.rs
```diff
@@ -24,6 +24,12 @@ pub mod mocks {
         instructions::do_nothing::do_nothing(ctx)
     }
 
+    /// Benchmark the emode reconcile functions' compute cost (logs CU via sol_log_compute_units).
+    /// Not used in normal flows — invoked only by the opt-in `anchor run bench-emode`.
+    pub fn bench_reconcile_emode(ctx: Context<BenchReconcileEmode>, num_configs: u8) -> Result<()> {
```
IliaZyrin: I suggest we remove all benchmark functionality given that we already have the results anyway. If we ever need it again for some other purposes, we can always search in the git history.

tests/d14_driftRecLiquidation.spec.ts
```diff
@@ -81,7 +106,7 @@ describe("d14: Drift rec liquidation", () => {
       1,
       USER_ACCOUNT_D15,
       THROWAWAY_GROUP_SEED_D15,
-      STARTING_SEED,
+      STARTING_SEED
```
IliaZyrin: Format this file please.

tests/d14_driftRecLiquidation.spec.ts
```diff
@@ -463,17 +487,402 @@ describe("d14: Drift rec liquidation", () => {
...
+// Same flow as d14 above, but the liquidatee holds collateral in BOTH a Drift
+// and a Kamino bank, and a single receivership transaction
+// (start -> driftWithdraw + kaminoWithdraw -> repay -> end) unwinds both legs.
+const USER_ACCOUNT_D14B = "d14b_account";
+const THROWAWAY_GROUP_SEED_D14B = Buffer.from(
+  "MARGINFI_GROUP_SEED_123400000016"
+);
+const STARTING_SEED_B = 160;
+
+describe("d14b: Drift + Kamino mixed rec liquidation", () => {
```
IliaZyrin: This is probably redundant as we already test all multi scenarios like this one in `m03`.

tests/d15_fixedPrice.spec.ts
```diff
@@ -113,7 +142,7 @@ describe("d15: Fixed Drift price bank", () => {
       bankrunProgram.programId,
       driftGroup.publicKey,
       ecosystem.usdcMint.publicKey,
-      FIXED_SEED,
```
IliaZyrin, line 129: Please format the file.

tests/k07_kaminoWithdraw.spec.ts
```diff
@@ -120,8 +127,13 @@ describe("k07: Kamino Withdraw Tests", () => {
         (b) => b.bankPk.equals(bank) && b.active === 1
       );
 
-      // TODO assert balances
       assert.equal(kaminoBankBalance.active, 1);
+      // The Kamino bank position shrinks by the partial withdrawal.
+      assert.isTrue(
+        wrappedI80F48toBigNumber(kaminoBankBalance.assetShares).lt(
+          wrappedI80F48toBigNumber(kaminoBalanceBefore.assetShares)
```
IliaZyrin: If possible, would be better to have more precise assertions here.

tests/k09_kaminoWithdrawInterest.spec.ts
```diff
@@ -264,37 +280,46 @@ describe("k09: Withdraw from Kamino reserve with accrued interest", () => {
     assert.approximately(diffUser, amt * exchangeRate.toNumber(), amt * 0.0001);
     assert.equal(diffUser, -diffReserve);
 
-    // TODO moar asserts
+    // The withdrawal shrinks the user's Kamino collateral position by ~the withdrawn amount, and
+    // the position stays active (partial withdrawal).
+    const sharesAfter = await kaminoAssetShares(0);
+    assert.isBelow(sharesAfter, sharesBefore);
```
IliaZyrin: This seems redundant, the below check is enough.

tests/k10_kaminoLiquidate.spec.ts
```diff
@@ -358,7 +362,30 @@ describe("k10: Kamino Liquidation", () => {
...
+      (b) => b.bankPk.equals(banks[0]) && b.active === 1
+    )!;
+    const debtAfter = accAfter.lendingAccount.balances.find(
+      (b) => b.bankPk.equals(banks[0]) && b.active === 1
+    )!;
+    assert.isTrue(
+      wrappedI80F48toBigNumber(debtAfter.liabilityShares).lt(
+        wrappedI80F48toBigNumber(debtBefore.liabilityShares)
+      )
+    );
```
IliaZyrin: If possible, would be nice to have more precise assertions here.

tests/k12_BorrowPostInterest.spec.ts
```diff
@@ -113,7 +124,7 @@ describe("k12: Borrow Tests (Recycles mrgn banks from k10)", () => {
       mrgnID,
       throwawayGroup.publicKey,
       ecosystem.tokenAMint.publicKey,
-      seed,
+      seed
```
IliaZyrin: Format the file, please.

tests/s02_addBank.spec.ts
```diff
@@ -94,7 +94,85 @@ describe("Init group and add banks with asset category flags", () => {
     }
   });
 
-  // TODO add bank permissionless fails prior to opting in
+  it("(permissionless) add staked bank before opting in - should fail", async () => {
```
IliaZyrin: There is gonna be a conflict on this one, as I also [implemented it](https://github.com/0dotxyz/marginfi-v2/pull/598/changes#diff-5985e12b7606937f8bfaa0d167d109540e7e9d7a9c29ea5d401047aa6cbecc1dR112) =)

tests/s05_solAppreciates.spec.ts
```diff
@@ -275,41 +275,53 @@ describe("Borrow power grows as v0 Staked SOL gains value from appreciation", ()
       console.log("It is now epoch: " + epoch + " slot " + slot);
     }
 
-    // Next, the replenish crank cycles free SOL into the "on ramp" pool
+    // Next, the replenish crank cycles free SOL into the "on ramp" pool.
```
IliaZyrin: This file is drastically updated in [SVSP PR](https://github.com/0dotxyz/marginfi-v2/pull/598/changes#diff-789cd03e5aa697602973cb64f00e7dd00688886bb76e9dc4044d9bc1511cc048) as well, probably easier to just revert these.

tests/s07_withdraw_repay.spec.ts
```diff
@@ -131,7 +133,29 @@ describe("Withdraw staked asset", () => {
...
+      b.bankPk.equals(validators[0].bank)
+    )!;
+    const lstBalAfter = balances.find((b) =>
+      b.bankPk.equals(validators[0].bank)
+    )!;
+    assert.isTrue(
+      wrappedI80F48toBigNumber(lstBalAfter.assetShares).lt(
+        wrappedI80F48toBigNumber(lstBalBefore.assetShares)
+      )
+    );
```
IliaZyrin: More precise assertions please.

tests/s07_withdraw_repay.spec.ts
```diff
@@ -172,7 +198,29 @@ describe("Withdraw staked asset", () => {
...
+      b.bankPk.equals(bankKeypairSol.publicKey)
+    )!;
+    const solBalAfter = balances.find((b) =>
+      b.bankPk.equals(bankKeypairSol.publicKey)
+    )!;
+    assert.isTrue(
+      wrappedI80F48toBigNumber(solBalAfter.liabilityShares).lt(
+        wrappedI80F48toBigNumber(solBalBefore.liabilityShares)
+      )
+    );
```
IliaZyrin: More precision please.

IliaZyrin (changes requested):
Amazing work, a LOT of improvements!

Only a few points from my side.

programs/marginfi/src/state/price.rs
IMEF-FEMI: just made a PR for this [here](https://github.com/switchboard-xyz/switchboard-sdk/pull/18)

tests/utils/juplend/admin-instructions.ts
```diff
@@ -20,7 +20,7 @@ export type InitJuplendLiquidityArgs = {
 
 export const initJuplendLiquidityIx = (
   programs: JuplendPrograms,
-  args: InitJuplendLiquidityArgs,
+  args: InitJuplendLiquidityArgs
```
IliaZyrin: Nitpick but I would format this script (plus all the others with the same "comma" changes), it's just easier to review the whole PR this way.

IliaZyrin: Actually, it's not really a nitpick, given the amount of changed files. Please format/revert all of them, otherwise the very next PR someone will just get the commas back again and it will bloat the PR, and then it repeats again...💀

IliaZyrin (approved):

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
jgur-psyops: I think we added this so that PDA accounts can sign for liquidations as long as they have a fee payer. It was a similar reason to why the transfer account needed a payer. This enables people to create a liquidation "program" managed by a PDA (not that anyone has done that, afaik).

programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
```diff
@@ -1,76 +0,0 @@
...
-
-use crate::{
-    prelude::*,
-    state::{
-        bank::BankImpl,
-        marginfi_account::{BalanceImpl, LendingAccountImpl, MarginfiAccountImpl},
-    },
-};
-
-pub fn lending_account_purge_delev_balance(
```
jgur-psyops, line 16: I think we still need this ix to fully sunset banks in super-bankruptcy mode? 

cc @IliaZyrin

type-crate/src/types/bank.rs
```diff
@@ -150,8 +150,16 @@ pub struct Bank {
     ///   the bank may safely be closed if this is zero. Will never go negative.
     pub borrowing_position_count: i32,
 
+    /// Fee (in basis points) the liquidator earns when liquidating against this bank's liability.
+    /// * 0 falls back to the default (`DEFAULT_LIQUIDATION_FEE_BPS` = 250 bps = 2.5%).
+    pub liquidation_liquidator_fee_bps: u16,
```
jgur-psyops: Note: had to move these to Bank because Config ran out of space in 1.9

There is a funny footgun with these if you set them such that the sum is ~100%, but that's probably fine, we're not gna do that. 

I would prefer to use the whole u16 here like we do with the u32 0-100% values and do a u16_to_centi here to get a little more precision rather than expressing these in bps as that would give something like 6x more precision (since bps only uses up to 10k which is like 1/6 of the u16 space). You're probably right we don't need the whole u32 precision.

programs/marginfi/src/instructions/marginfi_account/purge_delev_balance.rs
IliaZyrin: I have rechecked and I think Jon is right, we should not delete this one.
The alternative which I thought would work as a replacement was `start_deleverage->withdraw_all->end_deleverage`, but it actually won't work: `withdraw_all` is prohibited during the receivership if the asset weight is zero, and if it's not zero, then the check in `end_receivership` would fail.

So we should either revert this one, or make the changes to unblock this deleverage-withdrawal explicitly.

type-crate/src/types/bank.rs
IMEF-FEMI: upping to u32 for consistency... was previously trying to save some space

jgur-psyops:
Note for the update: `LendingAccountLiquidateEvent` changes are breaking because `LiquidationBalances` was updated, old consumers will fail to parse pre_balances/post_balances properly: this is fine for our purposes.

PR #603, opened 2026-06-22
Persistent same-mint auto-rebalance orders

IMEF-FEMI:
#  Summary
  
This PR adds a permissionless auto-rebalance flow that keeps a user's deposits in the highest-yield banks for a given asset. A user places a  one-time, persistent order naming an asset, a set of allowed banks, a per-execution budget, and a keeper tip; thereafter any keeper can relocate that position across higher-yielding same-mint banks whenever the spread justifies it. One setup keeps capital chasing the best supply rate across the protocol's lending venues with no further user action, until the user cancels.
 
## How a user sets it up
  A rebalance order specifies:
  - Asset: a single mint; every referenced bank must use it.
  - Allowed banks: an explicit on-chain list of 2 to 8 banks, not venues. One venue can run several same-mint banks at different yields, and the 
  user may want only some eligible.
  - Amount: a per-execution total-value budget. Each run may relocate at most this much value across all banks; 0 means uncapped.
  - Keeper tip: lamports paid to the executing keeper, proportional to the value moved, from a per-account SOL fee pool the user funds  (default 0). Top-ups are permissionless; the authority can withdraw. The pool always retains its rent-exempt reserve.
  - Policy: minimum APR improvement required to move (default +5%) and a cooldown between moves (default 24h).
 
  Policy, allowlist, amount, and keeper tip are all editable in place; the order persists across executions and is closed only by the user (or permissionlessly once it can no longer act).
 
## How a keeper executes
  The keeper is unpermissioned. Off-chain it scans the order's allowed banks and picks a set of moves, each from a lower-rate source bank into a
  higher-rate destination, then submits one atomic transaction:
 
  start_rebalance(moves) → withdraw(sources) → deposit(destinations) → end_rebalance
  
  - start_rebalance ingests the declared moves, validates each (every referenced bank uses the order's mint and is allowlisted; each move's destination APR beats its source by the order's minimum improvement; cooldown elapsed), and snapshots pre-move state.
  - The keeper withdraws from the sources into its own token account and deposits into the destinations, interleaving each venue's refresh crank instructions as needed.
  - end_rebalance re-checks that no move's rate advantage inverted after its own market impact, reconciles the declared moves against the observed per-bank value deltas (strict conservation, within dust), enforces the order's value budget, verifies the account stays healthy at the maintenance requirement and no untouched balance changed, pays the keeper tip from the SOL pool, then closes the sandwich.
 
  The on-chain contract guarantees value is preserved and every destination beats its source by the required margin; selecting the best venues (and how to fan the position across them) is the keeper's job. Because orders persist and keepers compete for the tip, a position not yet optimally placed is a standing opportunity for the next keeper, so it converges over time
  
  ## Additional fixes
- **MFI-16 (Med) — foreign-account sandwich:** rebalance legs are now bound to
  the order's account and the sandwich shape is validated (`RebalanceMalformedSandwich`), with an allowlist-position guard (`RebalanceNoAllowlistPosition`). An attacker's own deposit/withdraw can no longer ride the victim's sandwich to pulse the rate.
- **MFI-15 (Med) — unused reference bank sets the tip denominator:** every referenced bank must be touched by a move, and the tip/amount denominator is derived from the order mint's decimals and the referenced sources, not a keeper-selected positional `banks[0]`.
- **MFI-03 (High) — principal retention across divergent same-mint oracles:** `reconcile` rejects net underlying-token loss (`total_actual >= -dust`) with a relative conservation-dust bound, so under-depositing at a higher-priced destination fails with `RebalanceValueLeak`. Conservation is enforced in underlying-token units, so it holds even when two same-mint banks price from different oracles (the divergent-oracle root cause also addressed, from the e-mode side, by the feed-family fix in PR #622).
- **MFI-04 (Med) — cross-transaction (Jito bundle) rate pulse:** the keeper tip is escrowed at `end_rebalance` and released by a separate permissionless `settle_rebalance_tip` after a per-order delay (10m–1h), and only if each move's destination bank realized more supply yield than its source over that window (destination vs source `current_yield_index / move_time_index`). A transient bundle-induced rate spike realizes no yield, so the tip is refunded to the fee pool and the attack is profitless. Residual (accepted): a bundle can still *induce* the move itself, but it earns nothing (tip refunded), the user's principal is conserved (see MFI-03), and repetition is bounded by the per-order cooldown and the conservation dust.


programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,677 @@
...
+        REBALANCE_FLAT_FEE_USD, REBALANCE_ORDER_SEED, REBALANCE_RECORD_SEED,
+    },
+    types::{
+        BalanceSide, Bank, HealthCache, MarginfiAccount, MarginfiGroup, OraclePriceType,
+        RebalanceOrder, RebalanceRecord, WrappedI80F48, ACCOUNT_IN_ORDER_EXECUTION,
+        ACCOUNT_IN_REBALANCE, ORDER_BLOCKING_FLAGS,
+    },
+};
+
+/// Equity (weight-1) USD value of a raw native token amount in `bank`, priced from its oracle.
```
IliaZyrin: Nitpick: `(weight-1)` reads confusing, maybe change to `(weight = 1)` instead?

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,677 @@
...
+        check_eq!(
+            disc,
+            MarginfiAccount::DISCRIMINATOR,
+            MarginfiError::InternalLogicError
+        );
+        let marginfi_account: &mut MarginfiAccount =
+            bytemuck::from_bytes_mut(&mut data[8..8 + std::mem::size_of::<MarginfiAccount>()]);
+
+        let allowed = &order.allowed_banks[..order.allowed_bank_count as usize];
+        let can_close = !marginfi_account
```
IliaZyrin: I think we should also enable closing if there is only **one** active position (of the order's mint).

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,677 @@
...
+        let pre_total = pre_src_value
+            .checked_add(pre_dst_value)
+            .ok_or_else(math_error!())?;
+        let post_total = post_src_value
+            .checked_add(post_dst_value)
+            .ok_or_else(math_error!())?;
+        let floor = pre_total
+            .checked_sub(REBALANCE_FLAT_FEE_USD)
+            .ok_or_else(math_error!())?;
+        check!(post_total >= floor, MarginfiError::RebalanceValueLeak);
```
IliaZyrin: I think these checks are redundant, since they operate with the same set of values and the same threshold as the previous set of checks above.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,677 @@
...
+    );
+    require_keys_eq!(
+        ctx.accounts.dst_bank.key(),
+        dst_key,
+        MarginfiError::InvalidBankAccount
+    );
+
+    // Remaining layout: [src venue/oracle accounts][dst venue/oracle accounts][post-move health
+    // observation set]. The first two segments price the {src,dst} rate gate; the tail is the full
+    // observation set (bank+oracle per active balance, after the move) for the init-health recheck.
```
IliaZyrin:
```suggestion
    // observation set (bank+oracle per active balance, after the move) for the maint-health recheck.
```

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,677 @@
...
+        ],
+        bump,
+    )]
+    pub rebalance_order: AccountLoader<'info, RebalanceOrder>,
+    #[account(mut)]
+    pub fee_payer: Signer<'info>,
+    pub system_program: Program<'info, System>,
+}
+
+pub fn close_rebalance_order(ctx: Context<CloseRebalanceOrder>) -> MarginfiResult {
```
IliaZyrin, line 286: Can probably merge it with the keeper's one, just by checking that `fee_recipient` is `marginfi_account`'s `authority` and allowing to close in this case. Not sure it's worth having a separate ix for this.

test-utils/src/rebalance.rs
```diff
@@ -0,0 +1,515 @@
...
+        let blockhash = self.test_f.get_latest_blockhash().await;
+        let ctx = self.test_f.context.borrow_mut();
+        let tx = Transaction::new_signed_with_payer(
+            ixs,
+            Some(&self.keeper.pubkey()),
+            &[&self.keeper],
+            blockhash,
+        );
+        ctx.banks_client.process_transaction(tx).await
+    }
```
IliaZyrin, line 1013: Seems to be a duplicate of the [RebalanceFixture's one](https://github.com/0dotxyz/marginfi-v2/pull/603/changes#diff-7e6b481f6acbde71d152b86e7a222089b44e579f9becda23fa1f116778d7dafbR235-R248).

IliaZyrin (commented):
LGTM, just a few minor things!

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
IMEF-FEMI: Actually, we can still call rebalance on a single active position: the allowlist is always ≥2 banks, and dst isn't required to be a pre-existing balance (the deposit ix call would create it).

IliaZyrin: yeah, makes sense!

test-utils/src/rebalance.rs
```diff
@@ -0,0 +1,670 @@
...
+
+impl RebalanceFixture {
+    /// A referenced native-USDC bank (one oracle) for the moves stream.
+    pub fn bank_meta(&self, bank: Pubkey) -> RebalanceBankMeta {
+        RebalanceBankMeta::new(bank, vec![self.oracle_metas[0].clone()])
+    }
+
+    /// A move of `ui_value` USD (== UI USDC amount at the $1 test oracle) from `src_index` to
+    /// `dst_index` (indices into the referenced-bank list).
+    pub fn usdc_move(&self, src_index: u8, dst_index: u8, ui_value: f64) -> RebalanceMove {
```
IliaZyrin: can probably just remove this wrapper

type-crate/src/types/rebalance.rs
```diff
@@ -0,0 +1,133 @@
...
+pub const MAX_REBALANCE_BANKS: usize = MAX_ALLOWED_BANKS;
+
+/// Declared moves one atomic rebalance may carry. Each move relocates value from one referenced bank
+/// to another; `end_rebalance` reconciles the declared amounts against real per-bank value deltas.
+pub const MAX_REBALANCE_MOVES: usize = 8;
+
+/// Balances a `RebalanceRecord` snapshots for the untouched-balance proof: every active balance not in
+/// the referenced set. At least one source is always an active balance, so at most
+/// `MAX_LENDING_ACCOUNT_BALANCES - 1` others remain.
+pub const MAX_REBALANCE_RECORD_BALANCES: usize = MAX_LENDING_ACCOUNT_BALANCES - 1;
```
IliaZyrin, line 28: Just to clarify, this max is only reachable when there is only one active (src) balance which we would **fully** withdraw and replace with a new active (dst) balance, right?

IliaZyrin (dismissed):

type-crate/src/types/rebalance.rs
IMEF-FEMI: yes that correct

programs/marginfi/tests/user_actions/rebalance.rs
```diff
@@ -0,0 +1,2754 @@
...
+        )
+        .await;
+    let res = f
+        .process(&[start_ix, withdraw_ix, deposit_ix, end_ix])
+        .await;
+    assert_custom_error!(res.unwrap_err(), MarginfiError::RebalanceValueLeak);
+    Ok(())
+}
+
+/// A shortfall within the dust tolerance passes (absorbs sub-unit venue rounding).
```
jgur-psyops, line 2479: For USDC and SOL this probably isn't much of an issue, but for e.g. BTC where a Satoshi is approaching a penny, the dust could be significant. The goal is probably for dust to be at most a few Satoshi, I don't see why it would be more under normal operation since there's no swap fees.

Overall, I think this is acceptable mechanically, but the dust tolerance shouldn't be proportional to the `declared_total` but rather admin-configured per mint. There is no swap or slippage fee to absorb so the amount actually doesn't matter, any loss here would normally be due to rounding issues.

For USDC I would probably set it to 10 * 10^-6 (10 atoms), for SOL I might do 10 lamports. So perhaps 10 atoms is (10 * 10 ^-mint_decimals) is a sensible default if you wanted to just back into a constant. Since we support multiple "moves" per tx you could also grant the tolerance per "move", or we could just set the limit considering up to 8 moves.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,1246 @@
...
+                .map(|w| I80F48::from(*w))
+                .collect::<Vec<_>>(),
+            record.settle_delay,
+        )
+    };
+
+    check!(
+        (clock.unix_timestamp as u64)
+            >= move_ts
+                .checked_add(settle_delay)
```
jgur-psyops, line 1265: Since `start_rebalance` must init this PDA again, but it cannot until this PDA closes, and the PDA cannot close until the tip settles, the effective minimum cooldown is `settle_delay` (I guess 600 seconds by default). I think that's fine, not sure if it's been documented as intentional tho.

The less intentional element here is that the authority can no longer cancel during this tip settlement period. I assume normally this isn't a big deal, but could see some edge cases where a user visits the page after an Order lands and is frustrated by their inability to close the Order.

If pending_tip == 0 the record could immediately close at `end_rebalance`

Overkill territory perhaps: the cleanest way to fix this whole thing would be to create a "RebalanceTipClaim" PDA that settles separately from the record. Then an account can have as many `RebalanceTipClaim` as needed, one per execution/Keeper that has served the account, and Keepers can redeem on their own schedule. The seed can be something like `["rebalance_tip_claim", rebalance_order, execution_seq]` where `execution_seq` monotonically increases with each execution. The order also could track `pending_rebalance_tip_claims: u8` and validate non-zero when closing or ignore it and allow "orphaned" `RebalanceTipClaim` that close even when the Order is gone (the latter resolves the authority-can't-close thing described above)

cc @IliaZyrin @dawid-sw

programs/marginfi/src/state/rebalance.rs
```diff
@@ -0,0 +1,272 @@
...
+            if actual > I80F48::ZERO {
+                total_moved = total_moved.checked_add(actual).ok_or_else(math_error!())?;
+            }
+        }
+
+        check!(total_actual >= -dust, MarginfiError::RebalanceValueLeak);
+        Ok((total_moved, total_ref_pre, dust))
+    }
+
+    fn verify_others_unchanged(&self, marginfi_account: &MarginfiAccount) -> MarginfiResult {
```
jgur-psyops, line 266: I believe you can still 
```
start_rebalance(ref banks = [src, dst])
withdraw src
deposit dst
deposit dust amount into unrelated_bank
end_rebalance
```
Because:
1) `start_rebalance` snapshots active balances outside the referenced banks (here none)
2) `end_rebalance` reconciles just the referenced banks (here src/dst)
3) `verify_others_unchanged` only checked non-referenced (here none) that existed at the start

This doesn't do any damage but lets the Keeper troll an account by giving it dust.

I suppose the most simple fix is to have `balance_states` record all balances, including inactive (which it does already, incidentally, by recording Default).  So perhaps all we need to do is add a loop like this to `verify_others_unchanged`:

```
for balance in marginfi_account.lending_account.balances.iter().filter(|b| b.is_active()) {
    let is_ref = self.ref_banks[..self.ref_bank_count as usize]
        .iter()
        .any(|r| r.bank == balance.bank_pk);

    if is_ref {
        continue;
    }

    let was_snapshotted = self.balance_states[..self.active_balance_count as usize]
        .iter()
        .any(|rec| rec.bank == balance.bank_pk);

    check!(was_snapshotted, MarginfiError::IllegalBalanceState);
}
```

There's probably a more clever solution that's more O(N) and avoids the two .any() or fits in the same loop `verify_others_unchanged` is already doing but this is probably cheap enough already.

tests/specs/orders/01_rebalanceOrders.spec.ts
```diff
@@ -0,0 +1,2328 @@
...
+      mintedShares.toString(),
+      "dst shares equal the fTokens minted by the move",
+    );
+
+    const orderAcc = await program.account.rebalanceOrder.fetch(order);
+    assert.ok(orderAcc.marginfiAccount.equals(ownerAcc), "order persists");
+
+    await closeOrder(order);
+  });
+});
```
jgur-psyops, line 3109: We might create a test that documents the greedy-keeper/attacker's profitability options to acknowledge them. As far as I can think, the best path for a Keeper with infinite funding that can manipulate interest rates is:

1. Momentarily make dst APR exceed src APR by min_improvement.
   This passes start_rebalance.
2. Rebalance executes and records move-time yield indices.
3. Before settlement, ensure dst's yield has grown even slightly more than src's. (This can be much less than min_improvement, it just must be >0.)
4. Settle and collect the full pending tip.


If this is the worst attack we can think of, then it's a good threat model I think, because all they get here is a lousy tip for massive amounts of fund movement and complexity. And most importantly, the user still benefited, even if trivially, so the attacker would have to continue improving APR to repeat indefinitely.

tests/specs/orders/01_rebalanceOrders.spec.ts
```diff
@@ -0,0 +1,2328 @@
...
+      ),
+    );
+  };
+
+  /**
+   * Build and execute the keeper sandwich: cranks -> start -> src withdraw -> dst deposit -> end.
+   * Packed into a v0 transaction + lookup table because the two full venue account sets exceed the
+   * legacy 1232-byte limit.
+   */
+  const runSandwich = async (opts: {
```
jgur-psyops, line 2120: I'd like to see if this still able to execute in the worst-case, which is probably either 16 integration positions, or 1 lending positions and 15 borrows.

tests/specs/orders/01_rebalanceOrders.spec.ts
```diff
@@ -0,0 +1,2328 @@
...
+      ).toFixed(0),
+      mintedShares.toString(),
+      "dst shares equal the fTokens minted by the move",
+    );
+
+    const orderAcc = await program.account.rebalanceOrder.fetch(order);
+    assert.ok(orderAcc.marginfiAccount.equals(ownerAcc), "order persists");
+
+    await closeOrder(order);
+  });
```
jgur-psyops, line 2392: Another scenario I'd like to test:
```
Allowed Rebalance banks: [A, B, C]
User has:
  deposit in A
  borrow in B
Rates:
  B > C > A
```
The Keeper will want to go A->B but cannot because of the Borrow in B. The Keeper can still complete the Order with A->C. If user repays B, then A->B is viable again. FE should strictly warn users when they have a liability in any bank in the Allowed Rebalance list but from our PoV this is valid/fine.

jgur-psyops: A minor troll users can do here is to borrow both B and C, rendering the Order unfillable.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,1246 @@
...
+    ctx: Context<PlaceRebalanceOrder>,
+    allowed_banks: Vec<Pubkey>,
+    min_improvement: Option<WrappedI80F48>,
+    cooldown_seconds: Option<u64>,
+    amount: Option<u64>,
+    keeper_tip: Option<u64>,
+) -> MarginfiResult {
+    // User-owned policy with sensible defaults: 5% min improvement, 24h cooldown, no budget cap,
+    // no keeper tip.
+    let min_improvement =
```
jgur-psyops, line 207: Consider a scenario like so:

```
User is lending A, min_improve = 5%
A = 0%
B = 6%
C = 20%
min_improvement = 5%
cooldown = 6h
```

Let's say a nasty Keeper moves you A->B. Valid, but clearly inferior here.

Slightly worse:
```
A = 0%
B = 15.1%
C = 20%
min_improvement = 5%
```

Now you are stuck in B and can't move to C on the next tick because C-B<min_improv.

Most likely we punt on this, because fixing it for N transfers is quite complex, but it is a tricky thing I wonder if there is some improvement to be made here that forces Keepers to allocate to the local maxima rather than just a better bank. Maybe there is not...

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -0,0 +1,1246 @@
...
+        #[cfg(not(feature = "client"))]
+        parsed.key,
+    )?;
+    bank.update_bank_cache(&group)?;
+    Ok(())
+}
+
+pub fn start_rebalance<'info>(
+    ctx: Context<'info, StartRebalance<'info>>,
+    moves: Vec<RebalanceMove>,
```
jgur-psyops, line 736: One troll here is just to move the minimum required to trip the user's cooldown. We are already managing this with tip proportion so it's strictly not profitable. For now, we will probably have no mitigation, and assume that keepers aren't going to throw money away to troll user cooldowns. We should remember to monitor for this in prod so we can add mitigation if somebody decides to be annoying.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -522,11 +595,8 @@ pub struct LendingAccountLiquidate<'info> {
             let a = liquidator_marginfi_account.load()?;
             account_not_frozen_for_authority(&a, authority.key())
         } @ MarginfiError::AccountFrozen,
-        constraint = {
-            let a = liquidator_marginfi_account.load()?;
-            let g = group.load()?;
-            is_signer_authorized(&a, g.admin, authority.key(), false, false, false)
-        } @ MarginfiError::Unauthorized
+        // Signer authorization moved to the handler so the CB-halt admin path can accept
+        // group.admin / risk_admin without requiring them to own the liquidator account.
```
IliaZyrin: In which cases do we want to liquidate using an account we don't own?

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -117,6 +157,19 @@ pub fn place_rebalance_order(
...
+    // balance in an allowed bank. This keeps the order tied to a real position (a full move still
+    // leaves the balance at the destination), so a later "no allowed position" state is a genuine
+    // exit rather than a pre-deposit gap, and permissionless close on that state cannot grief a user
+    // who simply has not deposited yet.
+    check!(
+        account
+            .lending_account
+            .balances
+            .iter()
+            .any(|b| b.is_active() && allowed_banks.contains(&b.bank_pk)),
```
IliaZyrin: Let's also check that it's a **deposit** position.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -307,6 +370,16 @@ pub fn update_rebalance_order(
...
         let mut order = ctx.accounts.rebalance_order.load_mut()?;
         if let Some(banks) = allowed_banks {
+            // Same invariant as placement: the new allowlist must contain a bank the account holds a
+            // position in, so an update can't leave the order pointing at venues with no position.
+            check!(
+                account
+                    .lending_account
+                    .balances
+                    .iter()
+                    .any(|b| b.is_active() && banks.contains(&b.bank_pk)),
```
IliaZyrin: Same: let's check that it's a deposit position.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -959,3 +1083,164 @@ impl<'info> Hashable for EndRebalance<'info> {
...
+    // and staleness-checked in `yield_index_of`).
+    for parsed in banks.iter() {
+        accrue_native_bank(parsed, &ctx.accounts.group, &clock)?;
+    }
+
+    let mut current_indices: Vec<I80F48> = Vec::with_capacity(banks.len());
+    for parsed in banks.iter() {
+        let bank = parsed.loader.load()?;
+        current_indices.push(yield_index_of(&bank, parsed.oracles, &clock)?);
+    }
```
IliaZyrin: I would merge these two loops.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
IliaZyrin: I like the overkill suggestion actually, in its "ignore the orphaned tips" form. Seems right to decouple orders landing from tips claiming.

programs/marginfi/src/errors.rs
```diff
@@ -430,31 +430,60 @@ pub enum MarginfiError {
     #[msg("Invalid Juplend withdraw intermediary ATA")]
     InvalidJuplendWithdrawIntermediaryAta, // 6512
     // **************END JUPLEND ERRORS
+    #[msg("Account is already at (or above) the target size")]
```
IliaZyrin, line 432: It looks like this file was incorrectly merged, please clean up errors categorization.

tests/specs/orders/01_rebalanceOrders.spec.ts
IliaZyrin, line 1: Nitpick: either put it to the basic set, or rename to 01

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
IMEF-FEMI: The intent was to let the risk admin run a quick liquidation in an emergency using an in-house pre-funded account. But I've tightened it since it could be somewhat dangerous if the risk admin hot wallet ever got compromised: the liquidator ownership check now runs on every path.

Thanks for the catch.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
IMEF-FEMI: Updated this. now the rate check prices every bank in the order's allowlist rather than only the ones the moves touch, so a keeper can't hide a better destination by leaving it out of the account set. Both examples are now handled, A->B is rejected while C is higher and still has capacity.

Also, rates are now compared after the move's own deposit, with every candidate priced at the same amount, so a thin pool can't win on a headline rate that collapses once the tokens land.

programs/marginfi/tests/user_actions/rebalance.rs
IMEF-FEMI: you're right.. although i think the admin config per mint would be an overkill for this. I'd lean towards using a constant instead

IMEF-FEMI: updated it to use 3 atoms per declared move, scaled by the venue multiplier. native-only sets get the raw 3, integration sets get 3 x multiplier since a leg rounds to one venue token and that's worth `multiplier` atoms at the current rate. measured a Kamino -> Drift and a Drift -> JupLend round trip at exactly 1 atom each, so 3 per move is a 3x margin.

tests/specs/orders/01_rebalanceOrders.spec.ts
IMEF-FEMI: i think it'll just be best to just not allow a user to include a bank they have a liability in when placing the order.

IliaZyrin (approved):

PR #604, opened 2026-06-22
Drift claims process

jgur-psyops:
Adds an instruction to claim from Drift's bad debt claims portal.

For any given bank, the claimant is the `liquidity_vault_authority`, the PDA who owns the corresponding Drift account. To get the `proof` and `amount`, one should query Drift’s distributor API: `GET <api-url>/eligibility/<liquidity_vault_authority>`

We would like to include this update in 1.9, since it shouldn't affect anything except the defunct Drift banks, and we want this resolved asap.

I intended to make this permissionless: `DriftClaimBadDebt` goes to the global fee wallet's ATA (currently our MS), and from there we will figure out how to distribute the funds to users. Drift requires the withdraw to go to the owner first, which is the `liquidity_vault_authority`'s ATA, so we withdraw there and then forward to the global fee wallet (much like e.g. jup withdraws work). Payer just pays to init ATA like the `claim_status`.

There is at least one T22 bank to consider (PyUSD), I believe I have supported it already.

### DIFF
--- a/patch-note-drafts/README_PATCH_NOTE_FORMAT.MD
+++ b/patch-note-drafts/README_PATCH_NOTE_FORMAT.MD
@@ -1,0 +1,11 @@
+# Porting to GH
+
+Line spacing in the IDE is probably configured to better fit your IDE window, use pandoc or similar
+to convert before pasting to GH:
+
+### Example
+```
+pandoc patch-note-drafts/patch-notes-0.1.9.md \
+  -f gfm -t gfm --wrap=none \
+  -o patch-note-drafts/patch-notes-0.1.9.github.md
+```
--- a/programs/marginfi/src/instructions/drift/claim_bad_debt.rs
+++ b/programs/marginfi/src/instructions/drift/claim_bad_debt.rs
@@ -271,9 +271,7 @@ impl<'info> DriftClaimBadDebt<'info> {
     }
 
     fn claimant_token_balance(&self) -> MarginfiResult<u64> {
-        Ok(accessor::amount(
-            &self.claimant_token_account.to_account_info(),
-        )?)
+        accessor::amount(&self.claimant_token_account.to_account_info())
     }
 
     fn emit_claim_event(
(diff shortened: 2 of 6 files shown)


patch-note-drafts/README_PATCH_NOTE_FORMAT.MD
```diff
@@ -0,0 +1,11 @@
+# Porting to GH
```
dawid-sw, line 1: not sure what this is about

programs/marginfi/src/instructions/drift/claim_bad_debt.rs
```diff
@@ -0,0 +1,261 @@
...
+            from: self.claimant_token_account.to_account_info(),
+            to: self.destination_token_account.to_account_info(),
+            authority: self.liquidity_vault_authority.to_account_info(),
+        };
+        let bump = self.bank.load()?.liquidity_vault_authority_bump;
+        let signer_seeds: &[&[&[u8]]] =
+            bank_signer!(BankVaultType::Liquidity, self.bank.key(), bump);
+        let cpi_ctx = CpiContext::new_with_signer(self.token_program.key(), accounts, signer_seeds);
+
+        token::transfer(cpi_ctx, amount)?;
```
dawid-sw, line 269: we can add some event, so it's easier to index these claim actions

dawid-sw (approved):
looks good! some comments

patch-note-drafts/README_PATCH_NOTE_FORMAT.MD
jgur-psyops: I realized when updating the patch notes that the local VS code .md has newlines that don't paste well into GH so the easiest solution is just run this command before pasting to GH after an update.

dawid-sw: ok SG

IliaZyrin (approved):
LGTM

PR #605, opened 2026-06-23
ReduceOnlyWithBorrowingPower

IliaZyrin:
This adds `ReduceOnlyWithBorrowingPower`, a new bank operational state for cases where new deposits/borrows should remain blocked like `ReduceOnly`, but existing assets should continue contributing to initial health and borrowing power.

### DIFF
--- a/tests/14_reduceOnlyBanks.spec.ts
+++ b/tests/14_reduceOnlyBanks.spec.ts
@@ -125,6 +125,149 @@ describe("Reduce-Only Bank Tests", () => {
     }
   });
 
+  it("(user 0) ReduceOnlyWithBorrowingPower collateral counts for new loans", async () => {
+    const user = users[0];
+    const userAccountKeypair = Keypair.generate();
+    const userAccount = userAccountKeypair.publicKey;
+
+    await user.mrgnProgram.provider.sendAndConfirm(
+      new Transaction().add(
+        await accountInit(program, {
+          marginfiGroup: marginfiGroup.publicKey,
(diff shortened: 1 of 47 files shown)


tests/14_reduceOnlyBanks.spec.ts
```diff
@@ -125,6 +125,149 @@ describe("Reduce-Only Bank Tests", () => {
...
+    const depositAmountTokenA = 0.5;
+    const depositAmountTokenA_native = new BN(
+      depositAmountTokenA * 10 ** ecosystem.tokenADecimals,
+    );
+    const borrowAmountUsdc = 1;
+    const borrowAmountUsdc_native = new BN(
+      borrowAmountUsdc * 10 ** ecosystem.usdcDecimals,
+    );
+
+    try {
```
jgur-psyops, line 154: nit: I don't hate the try/finally model but idk but AI is so obsessed with including it! I guess it's cool that just one test fails if something goes wrong instead of all the ones after this one but the CI wouldn't pass either way!

jgur-psyops (approved):
LGTM, great implementation!

dawid-sw (approved):
LGTM

tests/14_reduceOnlyBanks.spec.ts
IliaZyrin: `finally` also caught my eye in the beginning, but I left it anyway for consistency with the other existing tests (e.g. the very next test uses exactly the same block to "restore" the bank).

PR #608, opened 2026-06-29
add timelocked governance for high-risk admin operations

### LINKED ISSUE #562: Timelocked-admin for most general admin functions

## Summary

In light of the Drift exploit, we want to move some operations to a timelocked admin.

We want to remove the group admin's ability to do the following, while enabling a new  "timelocked_admin" to do them. The new timelocked admin is also the only one who can change the timelocked admin. These tasks, considered "high risk", 

should include:
* adding banks (including all integrations)
* configuring bank oracles (including fixed price)
* removing a bank from reduce/paused mode to operational
* switching a bank from isolated to collateral
* anything else where an attacker might edit financial info in such a way that could cause a serious loss

should not include:
* Pausing the protocol
* Adjusting deposit/withdraw caps
* Adjusting interest curves
* Setting a bank to reduce mode

This is NOT a timelock feature built in-program, just a new admin for timelocked features, and the timelock itself will be implemented in Squads.

ashleychandy:
## Description

Implements timelocked governance for high-risk admin operations by introducing a dedicated timelocked admin role, delayed execution workflows, and gating immediate execution of sensitive protocol changes.

## Changes

- Add timelocked admin configuration with a configurable execution delay.
- Introduce `TimelockedOperation` accounts to track scheduled governance actions.
- Add schedule/execute workflows for:
  - Bank creation (including integrations)
  - Bank activation (Reduce/Paused → Operational)
  - Bank oracle configuration
  - Fixed oracle price updates
  - Bank risk tier transitions
- Gate legacy admin instructions once a timelocked admin is configured.
- Add support for cancelling pending timelocked operations.
- Add migration for `MarginfiGroup` account reallocation to support new governance fields.
- Add shared timelock validation and execution utilities.
- Update CLI authorization and account handling to align with the new governance model.

Fixes #562


jgur-psyops:
Thank you for the PR!

This somewhat misunderstands the issue: we actually just need to move certain functions that should be timelocked to a new admin, there is no need to implement timelocking in the program itself, we would just pick a Squads account that has timelock already for that admin. 

So the goal here is actually much more trivial than this implementation, which also includes the timelock functionality. All we want to do is move adding/configuring banks to a different admin than high-urgency tasks like pausing the protocol, putting banks into reduce-only mode, or adjusting low-risk params like caps and interest curves.

jgur-psyops:
See comment above

ashleychandy:
Hi @jgur-psyops,
Thanks for the clarification, that makes sense.

I just want to confirm the intended scope before I put together a new PR.

My understanding is:

Add a separate bank_admin authority to MarginfiGroup (this should fit in the existing _padding_1, so no account reallocation should be necessary).
Configure it through the existing configure instruction (adding a new_bank_admin parameter, similar to new_admin).
Keep emergency / operational actions on admin, including panic_pause, panic_unpause, operational state changes (reduce-only / paused), and low-risk configuration such as caps and interest curves.
Move bank-management operations to bank_admin, such as lending_pool_add_bank, lending_pool_add_bank_with_seed, configure_bank_oracle, and set_fixed_oracle_price.

I had two questions:

Should lending_pool_add_bank_permissionless remain permissionless, since it's intended to be callable by any validator?
For lending_pool_configure_bank, should the entire instruction require bank_admin, or only the bank-management fields (e.g. risk_tier), while operational state and other fields remain on admin?

Does that match the intended scope, or are there other instructions that should also move to the new admin?

PR #610, opened 2026-06-29
feat(oracle): add Scope price adapter (+ Kamino/Drift/Juplend wrappers)

k0beLeenders:
## Summary (generated)

Adds **Kamino's Scope** oracle aggregator as a price source option for marginfi banks, alongside the existing oracle types.

Scope is an on-chain aggregator: it cranks many sources into a single `OraclePrices` account, with source-divergence validation at write time. This PR teaches marginfi to read a price out of that account — and to apply the same lending-reserve exchange-rate wrappers we already support for the other oracle types.

**This is a draft / PoC** to validate feasibility and effort, not a production-ready change. See "Out of scope" below.

## What this adds

Four new `OracleSetup` variants, all sourcing their base price from Scope:

| Variant | Accounts | Exchange rate |
|---|---|---|
| `Scope` (18) | `[scope_prices]` | none (raw price) |
| `KaminoScope` (19) | `[scope_prices, kamino_reserve]` | `total_liq / total_col` |
| `DriftScope` (20) | `[scope_prices, drift_spot_market]` | `cumulative_deposit_interest / PRECISION` |
| `JuplendScope` (21) | `[scope_prices, juplend_lending]` | `token_exchange_price / PRECISION` |

The `ScopePriceFeed` reads a `DatedPrice` from the Scope `OraclePrices` account by index and converts `{value, exp}` → `I80F48`, with owner / discriminator / staleness / index checks. The wrappers reuse the **existing** reserve-loading helpers, freshness checks, and exchange-rate formulas verbatim, scaling the Scope base price via a single `apply_multiplier`.

## Design decisions

- **Confidence-less by design.** Scope validates source divergence at write time (e.g. `CappedMostRecentOf`), so price bias and the `oracle_max_confidence` gate are no-ops here, and time-weighted == real-time. Trust shifts from per-feed confidence to Scope's divergence guard.
- **Feed index storage.** Scope addresses prices by a 0–512 index. It's packed into the low 2 bytes of `oracle_keys[4]` (unused by every Scope setup — the wrappers use `[0]` for the Scope account and `[1]` for the reserve). **No account migration.**
- **Program id.** Points at Kamino's mainnet Scope (`HFn8…2fWJ`) for the PoC so the test reads real live data; a self-hosted instance would swap this constant.

## Testing

Host unit tests (no validator), runnable via:

```
cargo test -p marginfi --lib state::price
```

- `scope_get_price_sol_usd` — decodes the **live mainnet SOL/USD price ($72.81)** from real `OraclePrices` bytes (account `3NJYftD5sjVfxSnUdZ1wVML8f3aC6mp1CXCL6L7TnU8C`, entry 0); asserts owner/discriminator/staleness/out-of-range-index handling and that bias is a no-op.
- `scope_apply_multiplier` — the kToken/LST rate scaling used by all three wrappers.

Existing price-feed tests are unchanged and still pass. The full marginfi library type-checks with all four new variants (existing `OracleSetup` matches already had catch-alls — no collateral breakage).

## Diff size

```
 errors.rs   |  13 +    (4 Scope error variants)
 price.rs    | 366 +    (adapter + 4 dispatch arms + 4 validate arms + 2 tests + docs)
 bank.rs     |   8 +    (4 OracleSetup variants + from_u8)
 3 files, 386 insertions, 1 deletion
```

A full Scope integration — including kToken / Drift / JupLend exchange-rate collateral — in under 400 lines, reusing existing wrapper code paths.

## Out of scope (intentionally)

- The price-pusher path for off-chain sources.
- Self-hosted Scope deployment (own program id, dep trim, mappings/config).
- Integration tests exercising a live reserve account alongside Scope.
- Security audit, IDL/client plumbing, config instruction support, staged rollout.

### DIFF
--- a/programs/marginfi/src/state/price.rs
+++ b/programs/marginfi/src/state/price.rs
@@ -81,6 +81,7 @@ pub enum OraclePriceFeedAdapter {
     PythPushOracle(PythPushOraclePriceFeed),
     SwitchboardPull(SwitchboardPullPriceFeed),
     Fixed(FixedPriceFeed),
+    Scope(ScopePriceFeed),
 }
 
 /// Checks oracles[0], which is typically the Pyth or Switchboard key.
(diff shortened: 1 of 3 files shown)


programs/marginfi/src/state/price.rs
```diff
@@ -1045,6 +1170,148 @@ impl PriceAdapter for FixedPriceFeed {
...
+            MarginfiError::ScopeInvalidPriceIndex
+        );
+
+        let off = SCOPE_HEADER_LEN + entry_id * SCOPE_ENTRY_LEN;
+        check!(
+            data.len() >= off + SCOPE_ENTRY_LEN,
+            MarginfiError::ScopeInvalidAccount
+        );
+
+        let value = u64::from_le_bytes(data[off..off + 8].try_into().unwrap());
```
jgur-psyops, line 1248: we avoid manual slicing outside of last resort, the scope model would be added to the type crate or appropriate mock program and deserialized with an anchor wrapper (like eg kamino reserves)

jgur-psyops:
To be resumed in an internal branch if we go in this direction.

PR #612, opened 2026-07-05
CB:  audit fixes

IMEF-FEMI:
## Summary

- **M-01 / M-02 (Hashlock H-01, High):** run `run_cb_price_gate` before oracle prices are consumed in `start_liquidation` and `lending_account_liquidate` (liquidatee at start, liquidator after taking on the liability). Admin CB-halt liquidation stays exempt.
- **M-03:** `cb_price_gate` now also enforces the long-window anchor (`cb_window_reference_price`), catching slow-walk manipulation the per-pulse tier-0 check misses.
- **L-01:** permissionless bankruptcy is gated across the whole portfolio;  admin/risk-admin settle in any state.
- **L-02 (Hashlock M-01):** interest accrual excludes the exact halt span via `cb_frozen_seconds`, independent of when accrual runs. Admin clear and configure-disable paths accrue before dropping the halt span.
- **I-01:** deposits into a halted bank are allowed only for accounts that  already hold a balance there, closing the dust-deposit liquidation-griefing path.


IliaZyrin (approved):
LGTM!

programs/marginfi/src/instructions/marginfi_group/clear_circuit_breaker.rs
```diff
@@ -25,7 +25,16 @@ pub fn lending_pool_clear_circuit_breaker(
         );
     }
 
+    let now = Clock::get()?.unix_timestamp;
     let mut bank = ctx.accounts.bank.load_mut()?;
+    // Consume the interest freeze while the halt span is still recorded: accruing here
+    // excludes the halted time (up to now) and restarts accrual from the clear.
+    bank.accrue_interest(
```
dawid-sw, line 32: there are two paths to reset CB. one is here.

but there is second in `lending_pool_configure_bank` with `circuit_breaker_enabled: Some(false)`. that path is also calling `reset_cb_runtime_state` and can have same consequences as described in L02

dawid-sw (changes requested):
looks good just one comment

PR #613, opened 2026-07-07
Account resize and FeeState V2 removal

dawid-sw:
## Summary

We need more space on `MarginfiGroup` and `FeeState` for the upcoming variable borrow-rate premium.

This PR shows that we can accomplish this with account resizing. It also removes `FeeStateV2` and the instruction used to copy settings.

## Account Size Changes

| Account | Old Size | New Size | Purpose |
|---|---:|---:|---|
| `MarginfiGroup` | `1056 B` | `9248 B` | Adds reserved padding for variable-borrow premium |
| `FeeState` | `256 B` | `512 B` | Adds reserved padding for variable-borrow premium |

## New Instructions

This PR adds two new permissionless resize instructions:

- `lending_pool_resize_group_account`
- `resize_global_fee_state`

These instructions:

- Check the account owner and discriminator on the raw account data
- Top up rent from the payer
- Resize the account to the new target size
- Zero-fill the newly added bytes

The v1 account data is a byte-identical prefix of the new layout, so all existing state carries over unchanged.

### Option A: Final-size structs, resize immediately after deploy

This is the approach used in this PR.

Flow:

1. Deploy the upgraded program
2. In the next slot, send one pre-staged transaction with both resize instructions
3. Accounts are resized to match the new struct sizes

Tradeoffs:

- Between upgrade and resize, the two affected accounts cannot be loaded
- Transactions touching those accounts will fail during that short window
- After resize, account size matches struct size
- All clients work with the new crate / IDL unchanged
- Only one deploy is needed
- The variable-borrow premium can later use the reserved padding with no further migration

### Option B: Resize first, bump structs in a later deploy

Alternative approach:

1. Keep structs at their v1 size
2. Resize accounts ahead of time
3. Extra bytes are ignored
4. Later, deploy a second upgrade that enlarges the structs

Tradeoffs:

- No outage during the initial resize
- Requires two deploys
- Until the second deploy, account size does not match struct size
- This breaks Anchor's exact-size deserialization in Rust clients, including `p0-cli`
- Clients would need patched reads during the interim period
- Client reads may need to be patched twice

**Option A** is picked


tests/specs/basic/02a_groupResize.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  it("a v1-sized group is bricked until the permissionless resize; state survives", async () => {
+    const adminBefore = (
+      await bankrunProgram.account.marginfiGroup.fetch(marginfiGroup.publicKey)
+    ).admin;
+
+    // Simulate the mainnet group as it exists BEFORE this deploy: truncate to the v1 size
+    const account = await banksClient.getAccount(marginfiGroup.publicKey);
+    const v1Data = Buffer.from(account.data).subarray(0, GROUP_V1_ACCOUNT_LEN);
+    bankrunContext.setAccount(marginfiGroup.publicKey, {
```
jgur-psyops, line 46: nit: ideally would dump a v1 group to json so we have it in fixtures, but honestly can't see why it would be necessary, so this is fine.

jgur-psyops (approved):
LGTM!

PR #615, opened 2026-07-09
Variable Borrow Premium

dawid-sw:
## Variable Borrow Premium

Adds a per-user surcharge on top of base borrow interest.

The premium is priced pairwise: the group holds a `(collateral premium_tag, liability premium_tag) → APR` matrix, and each liability accrues at the USD-weighted average rate across the account’s collateral.

Missing pairs and untagged banks (`tag 0`) pay `0%` — the matrix is sparse opt-in.

## Design Decisions

* **No account resizes, no migration.**
  All state is carved from existing padding:

  * The matrix (`64` entries) + settings live in `MarginfiGroup` padding.
  * `Bank` gains `premium_tag`, `premium_activated_at`, and `collected_premium_outstanding` in its two 16-byte reserves.
  * `Balance` reuses the zeroed ex-emissions bytes for `premium_rate_snapshot` and `premium_outstanding`.

  Layout-pinning unit tests and mainnet-fixture regression tests prove pre-existing accounts read as “no premium”.

* **Never retroactive.**
  Every snapshot rewrite claims at the old rate first, so a `0 → nonzero` rate transition cannot charge the period before the rate existed. Accrual is additionally clamped to `bank.premium_activated_at`.

* **Realized-only accounting.**
  Accrued premium is a receivable on the user’s balance. `bank.collected_premium_outstanding` is credited only when repay tokens actually enter the liquidity vault.

  Bankruptcy, tokenless repay, and side flips write the receivable off. The permissionless sweep ix therefore can never take lender liquidity.

* **Repay settles premium first, then principal.**
  `repay_all` settles everything.

* **`pulse_health` is the permissionless crank.**
  It claims pending premium and refreshes snapshots for all liabilities in one pass; there is no separate refresh ix.

  Snapshots also recompute on:

  * borrow / withdraw, including integration withdraws
  * flashloan end
  * liquidation end

  Deposits do not recompute because there are no oracles in remaining accounts — frontends should append `pulse_health`.

* **Health projection includes unpaid premium.**
  Dormant accounts degrade, and liquidation is the backstop for non-repayment.

* **Admin split:**

  * `emode_admin` configures the matrix and bank tags.
  * `global_fee_admin` sets the premium wallet on `FeeStateV2`.

## Relation to #613

If #613, account resize, merges, this feature switches from `FeeStateV2` to the resized `FeeState`, keeping the old PDA and IDL account name.

The resized group also makes room for a much bigger premium matrix. The current `64`-entry cap exists only because the matrix must fit in group padding.

All matrix reads go through a single accessor, `find_premium_rate`, so raising the capacity is contained.

## Breaking Changes

* `lending_account_pulse_health`, `lending_account_end_flashloan`, and `end_liquidation` now require the group account. Integrators calling these must add it as readonly.
* New IDL fields on `MarginfiGroup`, `Bank`, and `Balance`, using renamed padding / ex-emissions fields.

## New Instructions

* `lending_pool_configure_group_premium`
  Emode admin; one pair per call.

* `lending_pool_configure_bank_premium`
  Emode admin.

* `edit_fee_state_v2_premium`
  Global fee admin. But this is going to be removed as in the end we are going to use FeeState struct which already has updating instruction. So this is phantom and will be removed when #613 merges.

* `lending_pool_collect_bank_premium_fees`
  Permissionless sweep.


dawid-sw:
I added a refresh_unavailable gate that blocks a premium refresh when the health pass couldn't price every collateral leg and the account has premium-active debt.

The premium rate is a collateral-weighted average, and the health pass soft-skips stale oracles (values that leg $0). Refreshing off an incomplete pass drops the skipped leg from the weighting - so
you could zero your own premium by acting while your risky collateral's oracle is stale, keep a new borrow's 0% starting rate through an outage, or grief a victim's rate upward.

Before this, withdraw/borrow (and a liquidator's own leg) succeeded through a stale collateral oracle as long as the rest covered health; now those revert for premium-active accounts until oracles recover. Bounded - inactive debt still passes, and it recovers on the next clean pass/pulse.

dawid-sw:
> I added a refresh_unavailable gate that blocks a premium refresh when the health pass couldn't price every collateral leg and the account has premium-active debt.
> 
> The premium rate is a collateral-weighted average, and the health pass soft-skips stale oracles (values that leg $0). Refreshing off an incomplete pass drops the skipped leg from the weighting - so you could zero your own premium by acting while your risky collateral's oracle is stale, keep a new borrow's 0% starting rate through an outage, or grief a victim's rate upward.
> 
> Before this, withdraw/borrow (and a liquidator's own leg) succeeded through a stale collateral oracle as long as the rest covered health; now those revert for premium-active accounts until oracles recover. Bounded - inactive debt still passes, and it recovers on the next clean pass/pulse.


I losen this a bit, so we allow to not refreshing new rate on withdraw and liquidations - so its not blocked

programs/marginfi/src/instructions/marginfi_account/flashloan.rs
```diff
@@ -143,6 +166,10 @@ pub struct LendingAccountEndFlashloan<'info> {
     )]
     pub marginfi_account: AccountLoader<'info, MarginfiAccount>,
 
+    /// Needed for the same-asset emode checks and the premium snapshot recompute
+    #[account(
+        constraint = marginfi_account.load()?.group == group.key() @ MarginfiError::InvalidGroup
```
IliaZyrin: This is already checked with `has_one` above.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
jgur-psyops, line 434: Here .repay() primitive calls `increase_balance_internal` which in turn calls `claim_premium` on that liability banks, but it does not settle the premium into `bank.collected_premium_outstanding`. If the liability is fully closed, then the premium is not debited and is lost forever.

Here the punt is we could simply prevent the liability from fully closing (keep it empty, but live). The full solution is to settle it if possible (like receivership does, since it routes through `repay_all` rather than `repay`), but that is probably somewhat challenging. 

Story 5 is the relevant one here.

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
```diff
@@ -0,0 +1,68 @@
...
+    if active {
+        bank.flags |= PREMIUM_ACTIVE;
+        // Stamp only the inactive->active TRANSITION (an idempotent re-config of an active
+        // bank must not forgive pending accrual). Accrual is clamped to start here, so the
+        // deactivated window can never be charged or health-projected.
+        if !was_active {
+            bank.premium_activated_at = Clock::get()?.unix_timestamp;
+        }
+    } else {
+        bank.flags &= !PREMIUM_ACTIVE;
```
jgur-psyops, line 46: Deactivation is pretty messy. Currently, deactivation will cause `claim_premium` to write off all the premium cranked after that point, so anything accrued during the active time is gone.

We could just prohibit deactivation outright when borrows are non-zero (or borrow count non-zero, assuming we can backfill it to be canonical).

We could also have a third state where unrealized premium still claims, so e.g. in the "PremiumWindingDown" state `claim_premium` might materialize premium as in the "active" state but otherwise all places where accrual currently happens are skipped.

It's actually a "forgive all unpaid premium" instruction as implemented and should be documented as such if we keep it that way.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -1913,6 +2037,63 @@ impl<'a> BankAccountWrapper<'a> {
         }
     }
 
+    /// Materialize pending premium at the current snapshot rate. On premium-inactive banks the
+    /// receivable is written off instead: health no longer projects it, so repay must not
+    /// collect it.
+    pub fn claim_premium(&mut self) -> MarginfiResult {
```
jgur-psyops, line 2127: nit: there are two functions named `claim_premium` (one in account, one in premium), and `claim_premium` calls `claim_premium`. I suppose once could be `claim_premium_internal` or `claim_premium_on_balance`. Some of the documentation also uses the term `materialize` which I think works better than `claim` in some contexts, tho having both claim and materialize would be even more confusing so if we've picked claim might as well stick with it.

Organizationally it's a bit weird that the other `claim_premium` takes a `&mut Balance` but isn't an impl of Balance and the `update_premium_snapshots` takes a `&mut MarginfiAccount` but isn't an impl of MarginfiAccount and so forth.

type-crate/src/types/user_account.rs
```diff
@@ -290,16 +290,23 @@ pub struct Balance {
     /// Tag used by orders to reference this balance (0 means unused/unassigned).
     /// A tag may also have a non-zero value while having no orders.
     pub tag: u16,
-    pub _pad0: [u8; 4],
+    /// Collateral-weighted variable-borrow premium APR snapshot for this liability position,
+    /// encoded like interest-curve points via `milli_to_u32` (0-1000%). Recomputed by every
+    /// oracle-carrying ix (borrow, withdraw, liquidation, order end, `pulse_health`) but not
```
jgur-psyops, line 314: I didn't see prior versions of this PR, but I assume this was updated to fix the breaking changes to deposit that would require risk accounts.

On one hand, fair. On the other hand, it does break Story 3 in https://app.notion.com/p/mrgnlabs/Variable-Borrow-Premium-30cff20a24da802cb776f634bf72dac8: a borrower who deposits better collateral and improves their premium on paper does not realize the improvement until the next risk snapshot. You could be lending 1 BONK borrowing 1 WIF and paying 10% premium, deposit 100 USDC, and end up still paying the 10% premium (instead of near 0) until your next risk action, even if that's a year from now.

So here we have two choices: (1) force deposit to take risk accounts for any balances that have premiums enabled, which could satisfy Story 3 at the program level but be annoying for obvious reasons, or (2) keep as-is, and front end can send a refresh after any deposit where premium was theoretically improved, which satisfies Story 3 in practice. Notably, this would probably have to be in a bundle, and integrators would either also have to be aware of this, or potentially eat the increased premium until their next refresh.

cc @k0beLeenders

programs/marginfi/src/instructions/marginfi_account/repay.rs
```diff
@@ -76,17 +76,45 @@ pub fn lending_account_repay<'info>(
...
+    // Materialize pending premium up front (idempotent — later claims in the same ix see
+    // elapsed 0) so the receivable is measurable for the settlement event.
+    bank_account.claim_premium()?;
+    let premium_outstanding_before: I80F48 = bank_account.balance.premium_outstanding.into();
     let (repay_amount_post_fee, share_amount) = if repay_all {
-        bank_account.repay_all(in_receivership)?
+        bank_account.repay_all(in_receivership, !tokenless_repayment)?
     } else {
-        let share_amount = bank_account.repay(I80F48::from_num(amount))?;
+        // Premium settles before principal: take it out of the repaid amount and repay the
```
jgur-psyops, line 96: Technically Story 6 defines the opposite behavior: the principle settles first, then the premium. I think both behaviors would be justified actually, so this is fine as-is. Fee-first is more economically favorable to whoever is getting the fees (I guess the lenders, ultimately), and principle-first favors the borrower, so here we are explicitly favoring the fee recipient.

programs/marginfi/src/state/premium.rs
```diff
@@ -0,0 +1,733 @@
...
+        if !premium_active {
+            balance.premium_outstanding = I80F48::ZERO.into();
+            balance.premium_rate_snapshot = 0;
+            balance.last_update = now;
+            continue;
+        }
+
+        // Materialize at the OLD rate before overwriting it. This also bumps `last_update`,
+        // making a 0 -> nonzero rate transition safe (no retroactive projection).
+        claim_premium(balance, *liability_amount, *activated_at, now)?;
```
jgur-psyops: This is cool: if I'm understanding it correctly, the limitation we described in 3.2 and 3.6 (also see Story 3) does not fully apply: as long as the user's premium is realized at least once between the old and new rate, they get the appropriate rate applied to the appropriate duration for the old and new premium. 

If there is a "new new" before the user realizes the "new" premium, then the user will realize the entire balance at "old" and "new new" instead of "new", but that's easily avoidable by cranking users after a change.

As described in the above comment,, there is a bad interplay that causes this to fail with non-risk updates like deposits (I think):
```
t0: snapshot = "old"
t1: deposit technically changes the  rate to "new", but no refresh
t2: a crank refreshes the snapshot
```
Here, at t2, the claim charges "old" for the entire t0 -> t2 interval, then stores "new"

Consider also the slightly more elaborate case with a premium update in between:
```
t0: stored snapshot = old
t1: admin changes rate to "new", the user’s stored snapshot is still "old"
t2: safer collateral is deposited; user rate becomes "new new", while the stored snapshot remains "old"
t3: the crank claims the entire t0 -> t3 interval at "old", then computes and stores "new new"
```

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -2161,6 +2371,21 @@ impl<'a> BankAccountWrapper<'a> {
...
+        // Claim premium at the pre-mutation debt; write off the receivable on premium-inactive
+        // banks (see `BankAccountWrapper::claim_premium`).
+        if bank.get_flag(PREMIUM_ACTIVE) && had_liabs {
+            claim_premium(
+                balance,
+                current_liability_amount,
+                bank.premium_activated_at,
+                Clock::get()?.unix_timestamp as u64,
+            )?;
+        } else if !bank.get_flag(PREMIUM_ACTIVE)
```
jgur-psyops: Unfortunately, we still have places in code (liquidation) where an asset can flip into liability. In that case, I guess this branch doesn't fire, and could later create a premium. 

`snapshot_pass_writes_off_inactive_liability_receivable` covers the case where the premium is inactive (representative of the field this one is recycled from having something in it before the turn the premium feature on), but actually the field can be populated even after the feature is activated through this (admittedly moronic) liab -> asset flip (or vice-versa, maybe).

tl;dr I think the fix is straightforward: we can remove the !active condition and just have:
```
if bank.get_flag(PREMIUM_ACTIVE) && had_liabs {
    claim_premium(....)?;
} else if I80F48::from(balance.premium_outstanding) != I80F48::ZERO {
    // premium-inactive bank OR asset-side balance: no receivable should legitimately exist
    balance.premium_outstanding = I80F48::ZERO.into();
}
```

The alternative (which I tried once, pre-AI, and was able to do, but it was ugly) is disable ALL flipping of Balances (the last one being the one in liquidate, iirc). 

Same comment for `decrease_balance_internal` and `BankAccountWrapper::claim_premium`

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -212,17 +213,35 @@ pub fn lending_account_borrow<'info>(
...
+        !premium_scratch.refresh_unavailable(),
+        MarginfiError::PremiumSnapshotUnavailable
+    );
+
     // Revert if any involved bank's oracle price has jumped past the breach threshold.
     run_cb_price_gate(&marginfi_account, ctx.remaining_accounts)?;
 
+    // Claim premium at the old rates and refresh every liability's premium rate snapshot with
+    // the post-borrow collateral mix.
+    update_premium_snapshots(
```
jgur-psyops: If I'm interpreting this correctly, I guess here every premium-impacting oracle (if any liability implements premium) must refresh, so borrow must now supply all-non-stale oracles, which gets rid of the old trick where you could have:

```
lending $1 A
lending $100 B
borrowing $2 C
```

and supply

```
stale/junk oracle A
non-stale oracle B
non-stale oracle C
```

They must now be all non-stale (excluding Isolated, reduceOnly, etc). I'm not sure if anyone was explicitly relying on that "feature", but we must document that as a breaking change if we keep it as-is. 

We can loosen this logic a little for increasing an existing borrow, but probably can't loosen it more than this for new borrows. We could do a "force borrow" where we do something silly like assign the max premium in your pair table if any oracle is stale, but I doubt anyone would want that.

cc @k0beLeenders

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -569,6 +575,52 @@ pub fn lending_account_liquidate<'info>(
...
+/// [`check_liquidatee_health_and_refresh_premium`] for the `#[inline(never)]` rationale.
+#[inline(never)]
+fn check_liquidator_health_and_refresh_premium<'info>(
+    liquidator_marginfi_account: &mut MarginfiAccount,
+    group: &MarginfiGroup,
+    liquidator_remaining_accounts: &'info [AccountInfo<'info>],
+    now: u64,
+) -> MarginfiResult {
+    let mut premium_scratch = PremiumScratch::default();
+    check_account_init_health(
```
jgur-psyops, line 633: Here there is a convoluted path to dodge some premium.

init health runs the Initial health pass, where stale oracles are just skipped. Consider: 

```
liquidatee holds 
   $100 SOL collateral 
   $90 USDC debt
liquidator holds 
   $1000 SOL collateral, 
   newly-assumed $1 USDC debt, 
   stale-oracle $1 LST collateral
```

The SOL/USDC liquidation checks succeed because the stale LST is encountered only in the liquidator’s Initial health check, which is fine because their other collateral (SOL) is sufficient.

Subsequently `premium_scratch.complete = false` since the LST was marked "Skip", then `update_premium_snapshots` is a no-op. The newly created USDC liablity now has `premium_rate_snapshot` = 0, which will not update until the next successful risk activity.

programs/marginfi/src/instructions/marginfi_group/config_group_premium.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::MarginfiResult;
+use anchor_lang::prelude::*;
+use bytemuck::Zeroable;
+use marginfi_type_crate::types::{
+    MarginfiGroup, PremiumEntry, MAX_PREMIUM_ENTRIES, PREMIUM_TAG_EMPTY,
+};
+
+/// (emode admin only) Set one pair of the group's variable-borrow premium matrix.
+///
+/// `rate > 0` inserts or updates the (collateral_tag, liability_tag) pair; `rate == 0` removes
```
jgur-psyops, line 13: Here some reasonable upper limit is probably a good idea (say 100%). Or we could set the premium to the more-precise u32_centi kind that's naturally capped to 100% (up to 1000% seems excessive, though perhaps there are some limited use-cases).

programs/marginfi/src/state/premium.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// A collateral leg: contributes `usd_value` of weight at `premium_tag`.
+    /// * `usd_value` is UNWEIGHTED collateral USD; zero when the risk engine prices this
+    ///   collateral at zero (isolated banks, skipped stale oracles).
+    Asset { premium_tag: u16, usd_value: I80F48 },
+    /// A liability leg: gets claimed at its old rate and receives a recomputed snapshot.
+    /// Recorded for premium-INACTIVE banks too (`premium_active: false`) so the snapshot pass
+    /// can WRITE OFF their receivable instead of accruing.
+    Liability {
+        /// Matches `Balance.bank_pk` — how the entry finds its balance again post-health-loop.
+        bank_pk: Pubkey,
```
jgur-psyops: The memory usage of `PremiumScratch` is dominated by the `PremiumScratchEntry:Liability` which is in turn half `bank_pk`. That memory forces you to e.g. `inline` at Liquidation (since the struct currently costs like 1/4 of the stack budget)

Ideally we could narrow this down to use just `balance_index: u8`. The scratch is invalid if the Balances change anyways, so this seems like it would be doable.

Alternatively, a janky and easy win would be to store just the first N bytes of the Bank's PK here: the first 4-8 are probably more than sufficient to be unique in practice. AI hates this for obvious reasons: an attacker could prefix-farm for a matching bank key in permless Staked Collateral bank creation, but with the PDA limitations it's probably very unlikely to actually find a collision, and even if they pull off this astronomically unlikely task, all they achieve in the best case is a tiny discount on premium. It probably costs marginally less CU to compare the first N bytes as well.

imo we should also remove the enum and just have this to save the enum overhead:
```
#[repr(C)]
struct PremiumScratchEntry {
    value: I80F48,       // asset USD or liability amount or zero
    activated_at: i64,   // zero for asset entries
    premium_tag: u16,
    // bank_pk or whatever replaces it
    flags: u8,           // empty / asset / liability + premium-active
}
```

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -583,6 +587,84 @@ fn calc_weighted_liab_value_cached_standalone(
...
+    calc_value(
+        total_premium,
+        price,
+        bank.get_balance_decimals(),
+        Some(liability_weight),
+    )
+}
+
+/// Record a balance into the premium scratch during the health loop. Asset entries carry the
+/// UNWEIGHTED collateral USD (zero-priced collateral contributes zero weight).
```
jgur-psyops: Doesn't this use the "low bias" price when it gets fed by `lending_account_pulse_health`?

Fine imo, but notable that a "greedy" keeper can wait until the good collateral has poor confidence (or pricing, generally), raising the victim's rate.

programs/marginfi/src/state/premium.rs
```diff
@@ -0,0 +1,733 @@
...
+///   variant carries only the fields that are meaningful for it.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum PremiumScratchEntry {
+    /// Slot not in use (or a balance that contributes nothing to premium).
+    #[default]
+    Skip,
+    /// A collateral leg: contributes `usd_value` of weight at `premium_tag`.
+    /// * `usd_value` is UNWEIGHTED collateral USD; zero when the risk engine prices this
+    ///   collateral at zero (isolated banks, skipped stale oracles).
+    Asset { premium_tag: u16, usd_value: I80F48 },
```
jgur-psyops: it is a bit challenging to understand how this works when different valuations apply (Initial/Maint/Equity). My understanding is:

Initial: borrow, most withdraws, pulse, flashloan end, liquidator
Maint: liquidatee, orders end
Equity: receivership end

Notably, Initial sets ReduceOnly banks to 0 while the others include them.

That means the premium slightly differs depending on which ix computes it, or am I interpreting incorrectly?

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -1064,9 +1152,17 @@ pub fn get_health_components<'info>(
             }
         }
 
+        // Project premium into the liability side; collect snapshot-recompute data if requested.
+        let now = clock.as_ref().map(|c| c.unix_timestamp).unwrap_or(0) as u64;
```
jgur-psyops: Probably just want to unwrap here to panic if the clock doesn't load

jgur-psyops:
Excellent overall! Only big-picture comment would be that covering the stories as explicit tests would be a nice-to-have.

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
IliaZyrin, line 46: Shouldn't we also introduce `premium_deactivated_at` to not lose the premium accrued during the premium-active time?

programs/marginfi/src/instructions/marginfi_group/edit_fee_state_premium.rs
```diff
@@ -0,0 +1,36 @@
+// Global fee admin calls this to edit the variable-borrow premium fields on the fee state.
+use crate::MarginfiError;
+use anchor_lang::prelude::*;
+use marginfi_type_crate::{constants::FEE_STATE_SEED, types::FeeState};
+
+pub fn edit_fee_state_premium(
+    ctx: Context<EditFeeStatePremium>,
+    premium_wallet: Option<Pubkey>,
```
IliaZyrin: Should probably be changed to non-optional given that the `None` case is currently no-op.

IliaZyrin (commented):
Solid work!

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
dawid-sw: The loss path can't actually commit: the dust write-off only fires when the repay leaves `liability_shares < EMPTY_BALANCE_THRESHOLD`, and the post-liquidation `ExhaustedLiability` check rejects exactly that state (same threshold via `Balance::is_empty`) - so full close reverts, and partial liquidation keeps the receivable collectible on the next real repay. 

So we prevent the full-closing scenario, just as a revert. Since this rests on an implicit two-site threshold coupling, I'll add a regression test pinning it and try to think about a way to add extra layer of protection

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
dawid-sw: Hmm, I was thinking about this more like a hard break, so if anything regarding the premium feature goes off, we can just disable it. it should not be treated like typical operations. So I'm voting for just documenting it. If we like to remove premium just from one bank, then it should work well...

type-crate/src/types/user_account.rs
dawid-sw: hmm yea. we can make these accounts required in deposits, its better. I was just afraid about potential breaking change. but 2) option is not ideal either. so lmk @k0beLeenders what you prefer

programs/marginfi/src/instructions/marginfi_account/repay.rs
dawid-sw: yes its deviation from story 6, but the reason is that a full repay would strand `premium_outstanding` on a sub-`EMPTY_BALANCE_THRESHOLD` balance (never health-projected, wiped by the dust-clear), so premium-first is what guarantees the receivable settles before the balance can reach the dust zone

programs/marginfi/src/state/premium.rs
dawid-sw: Agree, we can actually make the admin-change case exact: `PremiumSettings.timestamp` is already stamped on every matrix change but never read - at snapshot refresh we'll split the claim interval there, charging the stored rate for [last_update, change] and the freshly computed rate for [change, now]. Your first timeline then charges "old" for t0→t1 and "new" for t1→t2 with no crank race; the mix-change component (t2 in your second timeline) stays approximated since we don't store collateral history - so still we need to crank

programs/marginfi/src/state/marginfi_account.rs
dawid-sw: agree

programs/marginfi/src/instructions/marginfi_account/borrow.rs
dawid-sw: yea. waiting for Kobe opinion here. for now leaving it as it is

programs/marginfi/src/instructions/marginfi_group/config_group_premium.rs
dawid-sw: good point. adding 100%

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
jgur-psyops: I agree just documenting is fine, as long as the caller is clear that the deactivation is destructive and irreversibly forgives any unpaid premiums cranked while deactivated. 

I guess if we are disabling it for one bank, our workflow would be to find all the borrowers for that bank and settle them up before we disable? A test that demonstrates the expected workflow to sunset an existing premium might help us in case we forget how to do this later.

programs/marginfi/src/instructions/marginfi_account/repay.rs
jgur-psyops: that makes sense technically, and is more consistent with what borrowers probably expect. It's already a benefit to borrowers that we don't charge the premium on the premium balance itself (it only compounds when realized), so favoring the fee-earner first here is a fair tradeoff.

programs/marginfi/src/state/premium.rs
jgur-psyops: I guess we can reduce the need to crank but it would require us to be able to write to the snapshot on deposit, which would require oracle accounts. 

With the current approach, we're just obligated to make an attempt to refresh all premium payers after any rate change, which shouldn't be too hard at our current scale. We also get into the economic disincentive in the spec: technically if we lower a premium, the incentive on our end is *not* to crank it, but the "friendly" keeper will ignore those incentives I guess and crank regardless if it sees a rate update.

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
dawid-sw: disabling means setting premium to `0`

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
dawid-sw: instead of gating the liquidator (blocks liquidations during outages) I'll seed the newly-assumed debt's snapshot with the max configured pair rate for its tag when the pass is incomplete - repriced at the next clean refresh, so keeping a stale leg becomes money-losing instead of free. sg?

programs/marginfi/src/state/marginfi_account.rs
dawid-sw: yes. I was biased by CU optimizations. we can do a no-bias version

programs/marginfi/src/state/premium.rs
dawid-sw: good point, I agree and will unify

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
dawid-sw: explained [here](https://github.com/0dotxyz/marginfi-v2/pull/615#discussion_r3670902255)

dawid-sw:
changes since last review:

- Stale premium on non-liability balances is now cleared at every claim site (pre-mutation), so an asset→liability flip can't resurrect it as debt
- `edit_fee_state_premium`: `Option<Pubkey>` → `Pubkey`
- Two `claim_premium` fns resolved via extension traits (`BalancePremiumImpl`, `MarginfiAccountPremiumImpl`); shared `write_off_premium()` helper
- Dropped redundant `group` constraint in end-flashloan (covered by `has_one`)
- Removed dead clock fallback in health components (no `Option`, no `unwrap_or(0)`)
- `PremiumScratch` shrunk ~1.2KB → ~530B: flat 32-byte entry, raw `balance_index` instead of `bank_pk`, write-time revalidation kept
- Premium rate cap: 100% APR (`MAX_PREMIUM_RATE`) enforced in `configure_group_premium`
- Deactivation stays forgive-all, documented as a lazy amnesty (+ `msg!` warning); retag + `pulse_health` crank is the non-forgiving wind-down, pinned by a sunset-workflow test
- Liquidator stale-oracle 0% dodge: incomplete pass now seeds the assumed debt at the max pair rate for its tag (claims at old rate first; raises anything below max, never lowers); reprices on next clean pass
- Scratch valuation made requirement-type independent: unbiased RealTime prices (kills the confidence-band crank-timing lever), ReduceOnly collateral always counts
- Story-3 deposit caveat documented (`[deposit, pulse_health]` bundle)
- Stale ReduceOnly oracle no longer falsely completes the premium pass (`unpriceable_leg`)
- Full-close liquidation coupling made structural (`is_empty` shared by dust write-off and `ExhaustedLiability`) + regression tests
- All spec stories now have explicit tests, deviations-by-design noted in comments

**Still TBD**
- Borrow all-oracles-fresh breaking-change (awaiting frontend team input)
- Interval-split accrual at `PremiumSettings.timestamp` - if we are ok with cranking or we should try to have more complicated design but with less cranking needed

dawid-sw:
ok there is one thing that I overlooked.  The unbiased premium pricing adds a per-collateral-leg oracle derivation on every health pass - I've gated it behind an empty-matrix check so non-premium groups pay nothing, but once premium is live, big accounts pay real CU (and the 16-position liquidation is already touching the 1.4M hard cap) for what is honestly a second-order accuracy gain in a fee rate. 

I'd revert to reusing the health pass's biased price for the premium weights and accept the small confidence-band drift - it's bounded by the max pair rate, self-corrects on the next refresh, and costs zero extra CU. WDYT?

programs/marginfi/src/instructions/marginfi_group/config_bank_premium.rs
```diff
@@ -0,0 +1,83 @@
...
+///
+/// To stop accrual WITHOUT forgiving earned premium, retag to an unused `premium_tag` and
+/// crank `pulse_health` instead — receivables stay collectible and the flag stays on.
+/// Tag changes never forgive anything; only the flag transition does.
+pub fn lending_pool_configure_bank_premium(
+    ctx: Context<LendingPoolConfigureBankPremium>,
+    premium_tag: u16,
+    active: bool,
+) -> MarginfiResult {
+    let mut bank = ctx.accounts.bank.load_mut()?;
```
IMEF-FEMI, line 28: I think we should enforce FREEZE_SETTINGS here

programs/marginfi/src/instructions/marginfi_account/repay.rs
```diff
@@ -76,17 +76,45 @@ pub fn lending_account_repay<'info>(
     )?;
 
     let in_receivership = marginfi_account.get_flag(ACCOUNT_IN_RECEIVERSHIP);
+
+    // Tokenless repayments skip the transfer entirely: the premium receivable is written off
+    // instead of settled, since no tokens back it.
+    let tokenless_repayment = authority.key() == group.risk_admin
```
IMEF-FEMI, line 80: Can we also reuse this variable [here](https://github.com/mrgnlabs/marginfi-v2/blob/1dc4ccb7e85a0e1cbbe832ce6f773d0f8a67cdb4/programs/marginfi/src/instructions/marginfi_account/repay.rs#L103) so we don’t need to keep both in sync later

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -1517,7 +1688,11 @@ pub fn check_post_liquidation_condition_and_get_account_health<'info>(
 
 /// Helper function to calculate weighted value for a single balance.
 ///
-/// Calculates asset or liability value with appropriate weights and price biases.
+/// Calculates asset or liability value with appropriate weights and price biases. When
+/// `need_premium_price` is set (asset legs of premium-relevant accounts only), the returned
+/// 5th element is the UNBIASED RealTime price for the premium scratch — derived from the same
+/// normalization pass as the health price (see `PriceAdapter::get_health_and_premium_prices`),
+/// so it costs ~nothing extra.
```
IMEF-FEMI:
```suggestion
/// 5th element is the requirement-typed price for the premium scratch — derived from the same
/// normalization pass as the health price, so it costs ~nothing extra.
```

IMEF-FEMI (approved):
LGTM.. just a few nits

PR #616, opened 2026-07-10
remove swb std_dev, configurable static confidence

### LINKED ISSUE #582: remove switchboard stdev
switchboard prices no longer use the stdev field (it's always zero), usage of this can be removed and/or replaced by some fixed or per-bank configurable confidence interval.

jgur-psyops:
Thank you to https://github.com/ashleychandy for starting the feature

Closes #582 

Removes swb standard dev, which is now defunct and always zero.

SWB banks can now configure a "static" confidence, we recycled the `bank_config.oracle_max_confidence` for this (which makes it somewhat a misnomer, but good enough). Most non-stables should be configured to 1-2% here.

This update assumes `bank_config.oracle_max_confidence` is 0 for all prod banks.


IMEF-FEMI (approved):
LGTM!

programs/marginfi/src/state/price.rs
```diff
@@ -1404,49 +1393,29 @@ impl SwitchboardPullPriceFeed {
...
 
-        // Fail the price fetch if confidence > price * oracle_max_confidence
-        let oracle_max_confidence = if oracle_max_confidence > 0 {
-            I80F48::from_num(oracle_max_confidence)
-        } else {
-            // The default max confidence is 10%
-            U32_MAX_DIV_10
-        };
-        let max_conf = price
+        // Note: negative prices also create negative confidence intervals (though we not anticipate
```
IliaZyrin, line 1398:
```suggestion
        // Note: negative prices also create negative confidence intervals (though we do not anticipate
```

IliaZyrin (approved):
LGTM

PR #622, opened 2026-07-18
Same-Asset E-Mode Pricing Fix

IMEF-FEMI:
## Summay

Addresses **L-01** (same-asset e-mode incomplete pricing equivalence). Same-asset treatment now requires banks to share mint, `oracle_keys[0]`, **and** oracle feed family, and integration-priced setups can no longer establish the
*liability* side. Oracle changes that alter the feed family or primary key are blocked while eligibility is enabled; fixed/deprecated/unset setups cannot participate.

**Deviation from the literal recommendation:** L-01 suggested requiring identical `oracle_setup` and equivalent multiplier behaviour. We deliberately do not go that far, because it would break the feature's main use case. Integration banks (Kamino/Drift/Solend/JupLend) share the native bank's mint and feed and
differ only by an exchange-rate multiplier; pairing that collateral against a native liability is the whole point. The shared feed's price risk cancels; the residual multiplier is the collateral's redemption value, only moves against you
in a venue credit event (already accepted by listing the bank), and is bounded by the leverage caps. It also cannot touch the debt side: integration banks can never be a liability (enforced by the borrow ix `is_marginfi_asset_tag`
constraint). Matching feed family gives the pricing-equivalence guarantee L-01 asks for without disabling native-liability × integration-collateral e-mode.

### DIFF
--- a/programs/marginfi/tests/admin_actions/setup_bank.rs
+++ b/programs/marginfi/tests/admin_actions/setup_bank.rs
@@ -765,6 +765,9 @@ async fn set_same_asset_emode_eligibility_success_and_fixed_rejects() -> anyhow:
     let set_fixed_ix = test_f
         .marginfi_group
         .make_lending_pool_set_fixed_oracle_price_ix(usdc_bank, I80F48!(1).into());
+    // Warp a slot first: this transaction is byte-identical to the earlier rejected set-fixed, and
+    // with the same blockhash it would be signature-deduped by BanksClient into that cached failure.
+    test_f.context.borrow_mut().warp_to_slot(100).unwrap();
     {
         let ctx = test_f.context.borrow_mut();
         let tx = Transaction::new_signed_with_payer(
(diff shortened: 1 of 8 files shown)


programs/marginfi/tests/admin_actions/setup_bank.rs
```diff
@@ -765,6 +765,9 @@ async fn set_same_asset_emode_eligibility_success_and_fixed_rejects() -> anyhow:
     let set_fixed_ix = test_f
         .marginfi_group
         .make_lending_pool_set_fixed_oracle_price_ix(usdc_bank, I80F48!(1).into());
+    // Warp a slot first: this transaction is byte-identical to the earlier rejected set-fixed, and
```
IliaZyrin, line 768: Nitpick: we usually use `dummyIx` to cope with it.

IliaZyrin (approved):
LGTM

programs/marginfi/tests/admin_actions/setup_bank.rs
IliaZyrin: Ah, just realized it's not in TS-tests, nevermind

IliaZyrin: Although may make sense to introduce a similar concept in Rust tests as well.

dawid-sw (approved):
lgtm

PR #623, opened 2026-07-19
Add liquidation record tagging with time-based premium growth

### LINKED ISSUE #419: Liquidate fee grows over time
This only applies to the receivership liquidation approach.

A liquidator should be able to permissionlessly "tag" an unhealthy account (more likely, its LiquidationRecord) with some instruction. After being tagged, as time elapses, the maximum profit a liquidator can earn from that account should grow. It should take some time, on the order of ~1 hour, for any growth to begin in this way after an account is tagged.

If the account is liquidated (even before the "tag" matures), the tag should reset. Since liquidation is highly competitive, whoever is first to be able to pick up the account at a perceived profit will (theoretically) claim it. 

In the extreme case after e.g. one week if nobody has claimed the account, the premium might have increased to e.g. 100%, indicating that all the collateral left on the account is undesirable.

IMEF-FEMI:
## Summary
Closes #419.

- Adds a permissionless liquidation-record tag for unhealthy receivership accounts.
- After tagging, the premium stays at the existing base for 1 hour, then grows linearly, reaching the 100% cap one week after the tag.
- The tag instruction clears an existing tag when called on an account that is healthy again, and any liquidation or deleverage that repays debt
resets the tag.
- Prevents tagged liquidation records from being closed, so the premium-growth clock cannot be bypassed.
- Updates legacy liquidation to require the liquidatee's liquidation record when one exists, allowing it to apply the same reset rules.


programs/marginfi/src/errors.rs
```diff
@@ -446,7 +446,9 @@ pub enum MarginfiError {
     CircuitBreakerRequiresWarmCache, // 6603
     #[msg("Oracle price deviates too far from the circuit breaker reference; action rejected")]
     CircuitBreakerPriceJump, // 6604
-                             // **************END CIRCUIT BREAKER ERRORS
+    // **************END CIRCUIT BREAKER ERRORS
+    #[msg("Liquidation record is already tagged")]
+    LiquidationRecordAlreadyTagged, // 6605
```
IliaZyrin: Nitpick but I would re-use one of the [vacated](https://github.com/0dotxyz/marginfi-v2/pull/623/changes#diff-da9844aec337b91d2ac85f4d202703b6060ce8a74de01b1dcb9a7774230dbaf5L114) ones instead.

programs/marginfi/tests/user_actions/tag_liquid_record.rs
```diff
@@ -0,0 +1,558 @@
...
+    assert!(res.is_err());
+    assert_custom_error!(res.err().unwrap(), MarginfiError::LiquidationPremiumTooHigh);
+
+    // ...while $2.80 is allowed, and the 20% repayment resets the tag
+    let liquidator_sol_acc = run_receivership_liquidation(
+        &test_f,
+        &liquidatee,
+        record_pk,
+        &liquidator_usdc_acc,
+        0.28,
```
IliaZyrin: Can change to `0.305` to check the exact boundary condition.

tests/specs/limits/m06_tagLiquidationRecord.spec.ts
```diff
@@ -0,0 +1,506 @@
...
+    config.depositLimit = BANK_LIMIT;
+    config.borrowLimit = BANK_LIMIT;
+    return config;
+  };
+
+  const sendTag = async (expectFail: boolean = false) => {
+    const liquidator = users[1];
+    const tx = new Transaction().add(
+      // The varied CU limit makes each tag tx unique so results are never replayed
+      ComputeBudgetProgram.setComputeUnitLimit({ units: 400_000 + tagNonce++ }),
```
IliaZyrin: Can use `dummyIx()` here.

tests/specs/limits/m06_tagLiquidationRecord.spec.ts
```diff
@@ -0,0 +1,506 @@
...
+
+    // The liquidation resets the tag and records the entry
+    const now = await getBankrunTime(bankrunContext);
+    const record = await fetchRecord();
+    assert.equal(record.taggedAt.toNumber(), 0);
+    assertKeyDefault(record.liquidationReceiver);
+    assert.equal(record.entries[3].timestamp.toNumber(), now);
+  });
+
+  it("(permissionless) re-tag after reset while still unhealthy", async () => {
```
IliaZyrin: Maybe if the account is still significantly unhealthy (more than ~0 underwater) after liquidation, it would be better to **not** reset it completely (to 0) on liquidation, but rather just bump `taggedAt` to the current time, so that the premium is reset but the account is still tagged and is signalling "I require more liquidations".

Don't have a hard opinion on this though, just feels weird to require manual re-tagging of eligible accs after each small liquidation.

IliaZyrin (approved):
Neat implementation!

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -162,9 +168,13 @@ pub fn end_receivership<'info>(
     let seized: I80F48 = pre_assets_equity - post_assets_equity;
     let repaid: I80F48 = pre_liabs_equity - post_liabilities_equity;
 
-    // clear receivership
+    // Clear receivership. Any completed liquidation that repays debt resets the tag, matching
+    // the tag's one-shot incentive model.
     marginfi_account.unset_flag(ACCOUNT_IN_RECEIVERSHIP, false);
     liq_record.liquidation_receiver = Pubkey::default();
+    if repaid > I80F48::ZERO {
```
jgur-psyops: I guess this means if you are a really big whale with a bad debt, what you want to do is liquidate yourself for 1 lamport repeatedly, then you keep resetting the tag. We could only reset the tag if you repay at least some portion of the user's negative health, e.g.:

```
let pre_deficit = I80F48::max(I80F48::ZERO, -pre_health);
let post_deficit = I80F48::max(I80F48::ZERO, -post_health);
let deficit_reduction = pre_deficit - post_deficit;
```

then let's say we want to make sure you paid at least 10% off (just an example, imo 25% or even 50% might make more sense there):

```
onst TAG_RESET_DEFICIT_FRACTION: I80F48 = I80F48!(0.10); 
let reset_threshold = pre_deficit * TAG_RESET_DEFICIT_FRACTION;
let should_reset_tag =
    post_deficit == I80F48::ZERO || deficit_reduction >= reset_threshold;
```

And a similar thing in liquidate.rs can use `pre_liquidation_health` and `post_liquidation_health` so it should be fairly consistent.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
```diff
@@ -644,4 +674,10 @@ pub struct LendingAccountLiquidate<'info> {
     pub bank_insurance_vault: UncheckedAccount<'info>,
 
     pub token_program: Interface<'info, TokenInterface>,
+
+    /// The liquidatee's liquidation record. Required whenever the liquidatee has one registered
+    /// (`liquidatee_marginfi_account.liquidation_record` is set); the completed liquidation may
+    /// reset its premium-growth tag.
+    #[account(mut)]
+    pub liquidatee_liquidation_record: Option<AccountLoader<'info, LiquidationRecord>>,
```
jgur-psyops: I am a little concerned about making this account required since it could break liquidation for classic liquidators, but at the same time I don't want to use marginfi_account space since we are running low. I think we just need one i64 here (`liquidation_tagged_at`) so maybe just using the account is preferable.

The premium growth doesn't affect classic anyways so I guess a third option is classic isn't even involved in this feature and doesn't reset the growth, though that feels weird to me.

cc @IliaZyrin @dawid-sw thoughts?

IliaZyrin: I slightly more prefer moving the `tagged_at` field to marginfi_account to not introduce a breaking change, although it does semantically belong to the record.

programs/marginfi/src/instructions/marginfi_account/tag_liquid_record.rs
```diff
@@ -0,0 +1,95 @@
...
+    // and any stale tag on them can be cleared.
+    if health > I80F48::ZERO || liabs == I80F48::ZERO {
+        check!(liq_record.tagged_at != 0, MarginfiError::HealthyAccount);
+        liq_record.tagged_at = 0;
+    } else {
+        check!(
+            liq_record.tagged_at == 0,
+            MarginfiError::LiquidationRecordAlreadyTagged
+        );
+        // While any balance bank is CB-halted, liquidation is admin-only, so the premium-growth
```
jgur-psyops: This is somewhat inconsistent since existing tags get to keep going. I suppose for consistency the best option is just to allow tagging even during CB halt. This seems reasonable imo because
* This is consistent with pre-CB tagging
* Unhealthy accounts are still unhealthy regardless of CB issues
* The incentive reflects the time spent in CB halt, which is useful if the CB was due to a genuine big price move and now liquidators must be encouraged to move rapidly.

programs/marginfi/src/instructions/marginfi_account/tag_liquid_record.rs
```diff
@@ -0,0 +1,95 @@
...
+use marginfi_type_crate::types::{
+    HealthPriceMode, LiquidationRecord, MarginfiAccount, MarginfiGroup, ACCOUNT_DISABLED,
+    ACCOUNT_IN_DELEVERAGE, ACCOUNT_IN_FLASHLOAN, ACCOUNT_IN_ORDER_EXECUTION,
+    ACCOUNT_IN_RECEIVERSHIP,
+};
+
+/// (Permissionless) Tags an unhealthy account, letting the allowed liquidation premium grow over
+/// time (see `tag_adjusted_premium`). Calling this instruction while the account is healthy again
+/// or has no liabilities clears any existing tag instead.
+/// * Fails if unhealthy and already tagged, or healthy and not tagged.
```
jgur-psyops, line 18: I think the worst case outcome is like this:

```
T0: account unhealthy -> tagged_at = T0
T1: account becomes healthy -> tag remains unless someone calls tag_liquidation_record
T2 (much later): account becomes unhealthy again -> old tagged_at still applies
T3: liquidation uses elapsed = T3 - T0
```

so we must remember to configure our friendly Keepers to aggressively purge tags on healthy accounts to avoid the above scenario.

programs/marginfi/src/instructions/marginfi_account/liquidate.rs
dawid-sw: I'm more for introducing breaking change. but its not a very strong opinion. losing space in marginfi_account is bad as its not easy to resize ( over 500k accounts)

programs/marginfi/src/instructions/marginfi_account/close_liquid_record.rs
```diff
@@ -22,13 +22,22 @@ const INACTIVITY_PERIOD_SECS: i64 = 60 * 24 * 60 * 60;
...
 /// - The record must match the account's `liquidation_record` field
 /// - The record must be inactive for at least 60 days (derived from the most
 ///   recent `LiquidationEntry.timestamp`), OR never have been liquidated at all
 ///   (all entry timestamps are zero).
 pub fn close_liquidation_record(ctx: Context<CloseLiquidationRecord>) -> MarginfiResult {
     let record = ctx.accounts.liquidation_record.load()?;
 
+    // A tagged record cannot be closed: closing would erase the premium-growth tag and let an
+    // unhealthy account restart the growth clock at will.
+    check!(
```
dawid-sw: I found some bypass. users can transfer their account using `transfer_to_new_account` ix and do bypass. So it can be a little argument to keep that on account rather than tag and move on account transfer.

dawid-sw (changes requested):
one comment

programs/marginfi/src/instructions/marginfi_account/close_liquid_record.rs
jgur-psyops: Good point yes, here we could prevent transfers when your liquidation record is still live, but that's pretty annoying, and moving the field to the account is probably the most straightforward solution.

programs/marginfi/src/instructions/marginfi_account/tag_liquid_record.rs
IMEF-FEMI: Agree... also made small updates so user actions can self-clear tags when healthy... so fewer tag purges on the keeper side

programs/marginfi/src/instructions/marginfi_account/liquidate_end.rs
```diff
@@ -38,27 +43,39 @@ pub fn end_liquidation<'info>(ctx: Context<'info, EndLiquidation<'info>>) -> Mar
 
     let pre_assets_equity: I80F48 = liq_record.cache.asset_value_equity.into();
 
-    // Note: We guarantee that liquidation improves health to at most 0, unless the account's net value is
-    // below the threshold, then we can clear it regardless (or not).
-    let ignore_healthy = pre_assets_equity < LIQUIDATION_CLOSEOUT_DOLLAR_THRESHOLD;
+    // Note: We guarantee that liquidation improves health to at most 0, unless the account's net
+    // value is below the threshold: such dust accounts may be fully cleared (ending healthy) and
+    // skip the premium cap, since a partial liquidation isn't worth the fees.
+    let is_dust_closeout = pre_assets_equity < LIQUIDATION_CLOSEOUT_DOLLAR_THRESHOLD;
```
jgur-psyops, line 49: Consider the following case:

User has
* $4 Kamino USDC collateral at 80% asset weight
* $10 P0 USDC debt at 100% liab weight
* net health = 4 * .8 - 10 * 1 = -6.8

This is a "dust closeout" by definition here.

The liquidator tries to:
* withdraw $4
* repay nothing

Here `is_dust_closeout` = true, as the user equity is <$5. `in_bad_debt` = true as $4 < $10

The new check in `if premium_capped` allows health to worsen by `extracted = max(0, seized - repaid)`, here that's $4, and maintenance health declined by -3.2, so it passes. The account was already insolvent so the `if pre_assets_equity >= pre_liabs_equity && post_assets_equity < post_liabilities_equity` check also passes. We don't do the `seized <= repaid * max_fee` because it's a dust closeout. 

So tl;dr this is a regression from main in that accounts below that low-value threshold can be drained of assets without repaying any debt, wheras in main right now you would have to repay at least some debt (~3.2 in the example above). Likely we don't need the `is_dust_closeout` skip for `seized <= repaid * max_fee`, the check must remain mandatory (update the doc accordingly). Skipping the healthy state check (with `ignore_healthy` via is_dust_closeout) is probably still fine though.

Let's fix and write a test for this <$5 case to validate there's no regression between main/here.

programs/marginfi/src/instructions/marginfi_account/borrow.rs
```diff
@@ -214,8 +215,8 @@ pub fn lending_account_borrow<'info>(
     // Check account health, if below threshold fail transaction
     // Assuming `ctx.remaining_accounts` holds only oracle accounts
     let mut premium_scratch = PremiumScratch::default();
-    check_account_init_health(
-        &marginfi_account,
+    check_account_init_health_and_clear_tag(
```
jgur-psyops, line 218: if all risk-bearing instructions are getting this modified tag-clearing variant then I think `end_execute_order` also needs it.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -336,6 +336,10 @@ impl MarginfiAccountImpl for MarginfiAccount {
         self.indexer_flags
             .sync_balance_derived(&self.lending_account.balances);
         self.indexer_flags.mark_active_now();
+        // A liability-free account is never taggable (see `tag_liquidation_record`)
```
jgur-psyops, line 339: I see this interesting edge cases causes a user repaying their last debt to clear the tag, and I suppose this is the only circumstance under which repay can clear the tag (since otherwise it doesn't do a full risk pass). 

`repaying_all_debt_clears_the_tag` documents this nicely but I wonder if it's worth mentioning either here or in repay somewhere since it's a rather odd flow to reach (though definitely clever and helpful).

programs/marginfi/src/instructions/marginfi_account/tag_liquid_record.rs
```diff
@@ -0,0 +1,78 @@
...
+        ctx.remaining_accounts,
+        None,
+        &mut None,
+        HealthPriceMode::Live { liq_cache: None },
+        true,
+    )?;
+
+    // Accounts with no liabilities cannot be meaningfully liquidated: they are never taggable,
+    // and any stale tag on them can be cleared.
+    if health > I80F48::ZERO || liabs == I80F48::ZERO {
```
jgur-psyops, line 38: nit: the usual maintenance checks use `assets >= liabilities` which therefore includes health = 0. Zero health should be considered consistently so probably health >= I80F48::ZERO here as well unless I'm missing something.

IMEF-FEMI: It's supposed to match `check_pre_liquidation_condition_and_get_account_health`, which start_liquidation uses: zero health is currently  liquidatable there, so this is meant to preserve the existing liquidation threshold for tagging. The `assets >= liabilities` form is used by the risk engine's user-action checks

programs/marginfi/src/constants.rs
```diff
@@ -85,10 +85,24 @@ pub const LIQUIDATION_FLAT_FEE_DEFAULT: u32 = 5000;
...
-/// of liquidation bonus, etc) if it has net assets worth less than this amount in dollars. This
-/// roughly covers the fee to open a liquidation record plus a little extra.
+/// Liquidators may leave the account healthy, or close it out entirely, if it has net assets worth
+/// less than this amount in dollars. The liquidation bonus cap still applies. This roughly covers
+/// the fee to open a liquidation record plus a little extra.
 pub const LIQUIDATION_CLOSEOUT_DOLLAR_THRESHOLD: I80F48 = I80F48!(5);
+/// Margin above the health-neutral discount (`asset weight / liability weight`) that
+/// `lending_account_liquidate` credits the liquidatee, so a fee-capped liquidation still improves
+/// health by more than rounding.
+pub const LIQUIDATION_HEALTH_GAIN_MARGIN: I80F48 = I80F48!(0.0001);
```
dawid-sw, line 94: Not used

dawid-sw (approved):
lgtm

programs/marginfi/src/constants.rs
IMEF-FEMI: used now 😁

PR #625, opened 2026-07-23
Fix: repay/withdraw_all can leave share dust

jgur-psyops:
Fixes an issue where balances with opposite-side dust (e.g. a Deposit with some dust liability shares), which occasionally exist due to e.g. liquidation, are properly credited when positions close to avoid orphaning those dust-shares. 

Previously orphaned dust-shares are generally benign, but in edge cases could prevent the withdraw of the last few satoshi in a bank.


IliaZyrin (approved):
LGTM

dawid-sw (approved):
lgtm

PR #631, opened 2026-07-24
New oracles

IliaZyrin:
This introduces a few new oracle setups:
- `PythMSOL` (+ its Kamino and Juplend variations) - for pricing mSOL specifically
- `PythLST` (+ its Kamino and Juplend variations) - for pricing all the other LSTs
- `PTPyth` - for pricing PT-bulkSOL and PT-hyloSOL
- `PTFixed` - for pricing PT-hyUSD

For `MSOL` and `LST` oracles, `lending_pool_configure_bank_oracle` sets both the main feed (Pyth SOL/USD one) and the pricing account (Marinade State / Stake Pool, respectively). Integration (Kamino, Juplend) oracle stays untouched **as before**.

**Breaking changes**: the `set_fixed_oracle_price` ix is renamed to `set_oracle_price`.
It now takes an explicit oracle `setup` argument, plus up to **two remaining accounts** for that setup:

- No extra accounts → plain `Fixed`, as before (price ≥ 0).

- One account:
an Exponent-owned vault → `PTFixed`. `price` is the PT linear-pricing start price, in (0, 1].
otherwise the venue reserve/lending state → `FixedKamino` / `FixedDrift` / `FixedJuplend`, chosen by the bank's `asset_tag` (price ≥ 0).

- Two accounts ([Pyth SOL/USD feed, Exponent vault]) → `PTPyth`. `price` is the start price, in (0, 1].


programs/marginfi/src/state/price.rs
```diff
@@ -1131,13 +1611,23 @@ impl OraclePriceFeedAdapter {
                 }
             }
             OracleSetup::Fixed => {
+                check_eq!(
+                    bank_config.asset_tag,
+                    ASSET_TAG_DEFAULT,
+                    MarginfiError::InvalidOracleSetup
+                );
```
IliaZyrin, line 1444: Not sure if we should also allow `ASSET_TAG_STAKED` here. My opinion - we shouldn't but I may be not seeing some use case.

Thoughts? @jgur-psyops @dawid-sw @IMEF-FEMI

programs/marginfi/src/state/price.rs
```diff
@@ -277,6 +284,183 @@ fn juplend_price_multiplier(lending: &JuplendLending) -> MarginfiResult<I80F48>
...
+        .checked_div(I80F48::from_num(MSOL_PRICE_PRECISION))
+        .ok_or_else(math_error!())?)
+}
+
+/// Minimal borsh view of an SPL `StakePool`: the fixed-size prefix up to `total_lamports` /
+/// `pool_token_supply`. Trailing bytes are ignored; `account_type` (1 = StakePool) sits at byte 0
+/// rather than an 8-byte discriminator, so it deserializes from the start.
+#[derive(AnchorDeserialize)]
+#[allow(dead_code)] // fields before the two we read only advance the borsh cursor
+struct MinimalStakePool {
```
jgur-psyops: I would put all the stake-related state and functions in some other file.

programs/marginfi/src/instructions/marginfi_group/set_oracle_price.rs
```diff
@@ -36,7 +36,33 @@ pub fn lending_pool_set_fixed_oracle_price(
         return err!(MarginfiError::Unauthorized);
     }
 
-    bank.config.oracle_setup = if bank.config.asset_tag == ASSET_TAG_KAMINO {
+    let price_i80: I80F48 = price.into();
+
+    // Two accounts ([Pyth SOL/USD, vault]) signal PTSOL; a single Exponent-owned vault signals
+    // PT-hyUSD (hyUSD ~= $1, no base feed). A plain fixed price passes no accounts; the Fixed*
```
jgur-psyops: I think you probably want to generalize this to "A PT token that has a Pyth oracle" (maybe PTPyth or PT_Pyth) and "A PT token that has a fixed price" (maybe PYFixed or PT_Fixed), but I am guessing if we can't find a Pyth oracle for Hyusd we would actually prefer a swb oracle vs a fixed price. Setting a fixed price just for hyusd when we have yet to do so for the other stables is probably not the best move.

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
```diff
@@ -39,6 +41,30 @@ pub fn lending_pool_configure_bank_oracle(
         bank.config.oracle_setup = setup_type;
         bank.config.oracle_keys[0] = oracle;
```
jgur-psyops, line 45: one additional check we might add is for banks migrating away from Pyth. In `get_oracle_max_age` we have a bypass for Pyth only where they ignored a 0 oracle_max_age and used the `MAX_PYTH_ORACLE_AGE` instead, the new Pyth-based banks should inherit this behavior or check for zero-ages here (I don't recall if we have any left in prod offhand, but probably not).

tests/specs/basic/20_lstMsolOracle.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    );
+  });
+
+  it("(admin) tries to configure PythMSOL - fails with a non-marinade account", async () => {
+    await expectFailedTxWithError(
+      async () => {
+        await setOracle(ORACLE_SETUP_PYTH_MSOL, [BSOL_POOL]);
+      },
+      "MarinadeStateValidationFailed",
+      6600,
```
jgur-psyops: This is a different error in our program, does this one come from marinade? Likewise for 6601 and such in 21_. Or maybe this is just a bug in `expectFailedTxWithError`

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
```diff
@@ -25,8 +25,10 @@ pub fn lending_pool_configure_bank_oracle(
...
                 | OracleSetup::FixedDrift
                 | OracleSetup::FixedJuplend
+                | OracleSetup::PTSOL
+                | OracleSetup::PTHYUSD
         ) {
-            return err!(MarginfiError::UseSetFixedOraclePrice);
+            return err!(MarginfiError::UseSetOraclePrice);
         }
         check!(
             !bank.get_flag(BANK_SAME_ASSET_EMODE_ELIGIBLE)
```
jgur-psyops, line 35: nit: The new PythLST banks (and whatever we settle on for PT-with-oracle) might explicitly excluded from SAME_ASSET_EMODE eligibility, since e.g. LST will have SOL in keys[0] but are not actually SOL. since they have a different mint, they will not participate in the feature automatically anyways, so I think there's no risk to the current implementation.

We do have to be extra careful changing a bank with emode enabled since if the exchange rate is not aligned with what our prior oracle reported it could liquidate people.

programs/marginfi/src/state/price.rs
```diff
@@ -277,6 +284,183 @@ fn juplend_price_multiplier(lending: &JuplendLending) -> MarginfiResult<I80F48>
...
+
+    if now <= start_ts {
+        return Ok(start_price);
+    }
+    if now >= maturity {
+        return Ok(I80F48::ONE);
+    }
+
+    // start_price + (1 - start_price) * (now - start_ts) / (maturity - start_ts)
+    let progress = I80F48::from_num(now - start_ts)
```
jgur-psyops: we have a `lerp` function in `interest_rate` you might be able to use, slightly CU optimized by skipping certain edge case checking.

programs/marginfi/src/state/price.rs
```diff
@@ -277,6 +284,183 @@ fn juplend_price_multiplier(lending: &JuplendLending) -> MarginfiResult<I80F48>
...
+#[allow(dead_code)] // fields before the two we read only advance the borsh cursor
+struct MinimalStakePool {
+    account_type: u8,
+    manager: Pubkey,
+    staker: Pubkey,
+    stake_deposit_authority: Pubkey,
+    stake_withdraw_bump_seed: u8,
+    validator_list: Pubkey,
+    reserve_stake: Pubkey,
+    pool_mint: Pubkey,
```
jgur-psyops: it looks like you're able to get the mint for both the msol and "vanilla" stake variant: let's check that mint = bank.mint during oracle validation to reduce the risk of that footgun.

programs/marginfi/src/instructions/marginfi_group/set_oracle_price.rs
```diff
@@ -36,7 +36,33 @@ pub fn lending_pool_set_fixed_oracle_price(
...
+    // integrations pass a single non-vault (reserve/lending) account.
+    let is_ptsol = ctx.remaining_accounts.len() == 2;
```
jgur-psyops: Switching the instruction logic based on the accounts length is probably a bad idea, if there are different operating modes have the caller pass a flag of some sort to indicate their desired operating mode.

programs/marginfi/src/instructions/marginfi_group/set_oracle_price.rs
jgur-psyops: I think ASSET_TAG_SOL would default to "Fixed" here and then fail validation later. Doesn't matter really, just an observation, could make it explicit that we reject fixed prices for ASSET_TAG_SOL or test that it is expected to fail. Probably don't need to support it since I don't see why we would ever do so.

tests/specs/juplend/jlr12_lstMsolOracle.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  it("(admin) configures JuplendLST - happy path", async () => {
+    await setOracle(ORACLE_SETUP_JUPLEND_LST, oracles.wsolOracle.publicKey, [
+      lending,
+      BSOL_POOL,
+    ]);
+    const { config } = await program.account.bank.fetch(bank);
+    assert.deepEqual(config.oracleSetup, { juplendLst: {} });
+    assertKeysEqual(config.oracleKeys[0], oracles.wsolOracle.publicKey);
+    assertKeysEqual(config.oracleKeys[1], lending); // lending preserved, not rewritten
+    assertKeysEqual(config.oracleKeys[2], BSOL_POOL);
```
jgur-psyops: As a broad note, it's a little confusing that the tests actually purposefully configure banks with a stake account for a different mint, which would be a pretty serious mistake in prod.

It could be difficult to do this because the setup is "wrong" in that sense but it would be nice to validate that the price before/after is the same for this kind of bank when configured in the real production-way. Since our placeholder "price" from the LST oracle was made-up to begin with I guess that's probably not possible unless the whole E2E uses real fixtures and a real Pyth oracle instead of our made-up one.

programs/marginfi/src/state/price.rs
```diff
@@ -277,6 +284,183 @@ fn juplend_price_multiplier(lending: &JuplendLending) -> MarginfiResult<I80F48>
         .ok_or_else(math_error!())?)
 }
 
+fn load_marinade_state<'info>(
```
jgur-psyops: Byte-slicing is pretty safe on a program like msol that's been stable for a long time, but I wonder if we can add some minor sanity checks at price-load time, for example price < (200 x SOL). That should catch any unlucky slicing issues and leaves plenty of room for us to be SOL maxis. Same general theme for vanilla stake and likewise we can check the bound on the timestamps in the PT account to the next five years or something.

programs/marginfi/src/state/price.rs
jgur-psyops: Not sure what the question is: do you mean if we should allow Staked assets to use a multiplier instead of reading from svsp/staked pool accounts? I think not. And Staked Assets should never be able to switch oracle types, in general.

IliaZyrin: Yeah the question was about if we ever might need to set a fixed price for a staked bank (we have [a comment](https://github.com/0dotxyz/marginfi-v2/pull/631/changes#diff-9dda96f29e54e4db5b4ca3a37b50c195619e99bf262cf21035deef9180e93c54R83) suggesting it is possible). But I also think not, so I will update the comment now.

programs/marginfi/src/instructions/marginfi_group/set_oracle_price.rs
IliaZyrin: All good points!

tests/specs/basic/20_lstMsolOracle.spec.ts
IliaZyrin: Interesting, yeah, I've updated the error codes while merging with the other branches, and this didn't fail because `expectFailedTxWithError` only checks for the name of the error if the logs exist.

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
IliaZyrin: I checked and confirmed there are currently no banks in prod with zero `oracle_max_age`. And there is no way to set it to zero at the moment. All possible setters (except for [clone_bank()](https://github.com/0dotxyz/marginfi-v2/blob/main/programs/marginfi/src/instructions/marginfi_group/clone_bank.rs) but that's okay) go through [the validation](https://github.com/0dotxyz/marginfi-v2/blob/main/programs/marginfi/src/state/bank_config.rs#L78-L81) ensuring it's at least 10 seconds. **So this bypass is now a dead code and I removed it entirely.**

tests/specs/juplend/jlr12_lstMsolOracle.spec.ts
IliaZyrin: Agreed, updated the tests to be less "hacky" and closer to reality. We now validate that `pricing_account.mint == bank.mint` for all LST/mSOL setups, so a wrong-mint config is rejected.

I've also added a price-parity test for a bSOL bank. It uses real fixtures to confirm that current SWB pricing [agrees with the new PythLST pricing](https://github.com/0dotxyz/marginfi-v2/pull/631/changes#diff-61f8ddceeae7a8d79a5e55f92c23d9871dab0720518010ec7349b67677a61929R227-R235).

programs/marginfi/src/state/interest_rate.rs
```diff
@@ -271,37 +271,31 @@ impl InterestRateCalc {
...
-    /// * returns start_y if end_x <= start_x or target < start_x
-    /// * returns end_y if target > end_x
+    /// * returns start_y if end_x <= start_x or target <= start_x
+    /// * returns end_y if target >= end_x
     /// * None if end_y < start_y. Note: this means curves where the rate decreases as the
     ///   utilization goes up are unsupported, though there's no reason you would generally want to
     ///   do that anyways.
     #[inline]
-    fn lerp(
+    pub(crate) fn lerp(
```
IliaZyrin, line 280: As part of re-using `lerp` in [pt_linear_multiplier](https://github.com/0dotxyz/marginfi-v2/pull/631/changes#diff-b76a862a933c5c2d647c69bf9d650bcdc4376d5f19e2842daca211db84c8cf81R248-R254), **I've updated the behavior of this function** to actually clamp on edge values, instead of returning `None`. Now it matches the comment and I've confirmed that the only place where it was called is unaffected by the change (it was anyway [already clamping](https://github.com/0dotxyz/marginfi-v2/pull/631/changes#diff-0da14b77cca97906c7d9908234afbad1dc817465bfbd3352df9c6a335be2a740R253-R254)).

cc @jgur-psyops

programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs
IliaZyrin: I think it makes sense to potentially allow them to participate in same-asset-emode. As you pointed out the mint-equality check will prevent different LSTs from intermingling with each other and with plain SOL.

But we can always make them "orphans" (i.e. return `None` as their feed family), if we really want to hard-disable it.

programs/marginfi/src/state/lst_stake_price.rs
```diff
@@ -0,0 +1,256 @@
...
+    // start and no more than ~5 years out, which comfortably bounds any real Exponent vault.
+    let horizon = now
+        .checked_add(PT_MAX_MATURITY_HORIZON)
+        .ok_or_else(math_error!())?;
+    check!(
+        maturity > start_ts && maturity <= horizon,
+        MarginfiError::ExponentVaultValidationFailed
+    );
+
+    Ok(InterestRateCalc::lerp(
```
IMEF-FEMI: I think we should also consider the underlying PT redemption rate here. this function assumes the PT reaches 1.0 at maturity, but that does not account for a loss/depeg in the underlying or the vault being under-backed. In that case we could end up pricing the PT above what it can actually redeem for.

Could we cap the expected linear rate with the converted redemption value? Something like:

```rust
 let expected_rate = InterestRateCalc::lerp(
      I80F48::from_num(start_ts),
      start_price,
      I80F48::from_num(maturity),
      I80F48::ONE,
      I80F48::from_num(now),
  )
  .ok_or_else(math_error!())?;

  // SY per PT × asset per SY = asset per PT.
  // This is 1.0 while fully backed, and falls below 1.0 only if the
  // vault's escrow is short relative to PT supply.
  let redemption_rate = pt_redemption_rate(vault)?
      .checked_mul(vault.last_seen_sy_exchange_rate())
      .ok_or_else(math_error!())?;

  Ok(expected_rate.min(redemption_rate))
  ```

  This would preserve the existing linear model for healthy vaults, while preventing the PT from being marked above its actual redemption backing if the vault becomes under-backed.

  Note that sy_for_pt / pt_supply is denominated in SY per PT, while this function is denominated in the underlying asset per PT, so we would need to convert before comparing them. While fully backed, sy_for_pt == pt_supply / sy_exchange_rate 
  ([source](https://github.com/exponent-finance/exponent-core/blob/HEAD/programs/exponent_core/src/state/vault.rs#L569)).
  so the bare ratio is 1 / sy_exchange_rate.

File: programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs (hunk 6 above)
Comment (IMEF-FEMI) [reply programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs:35]: I agree with @IliaZyrin. We could allow them to participate in same-asset e-mode, but we should tighten the checks.

The multiplier setups need their own feed family rather than being grouped with PythPush, since they can report a different effective price even when oracle_keys[0] is the same. For example, with two e-mode-eligible mSOL banks sharing a SOL/USD feed, switching the collateral bank from PythPushOracle to PythMSOL passes the migration guard [here](https://github.com/mrgnlabs/marginfi-v2/blob/new-oracles/programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs#L33-L39), since neither oracle_keys[0] nor feed_family() changes. validate_oracle_setup does not catch it either: load_price_update_v2_checked only proves that keys[0] is a Pyth price account, not that it is the expected base-asset feed. The collateral is then marked at SOL/USD × mSOL/SOL, while the liability side remains SOL/USD.

Comment (IMEF-FEMI) [reply programs/marginfi/src/instructions/marginfi_group/config_bank_oracle.rs:35]: We'd also need to update the e-mode liability allowlist [here](https://github.com/mrgnlabs/marginfi-v2/blob/new-oracles/programs/marginfi/src/state/marginfi_account.rs#L864-L872).

File: programs/marginfi/src/state/lst_stake_price.rs [hunk 14]
```diff
@@ -0,0 +1,256 @@
...
+    Ok(())
+}
+
+/// Pricing: LST/SOL exchange rate = `total_lamports / pool_token_supply`, with sanity bounds.
+pub(crate) fn stake_pool_price_multiplier(
+    bank_config: &BankConfig,
+    stake_pool_info: &AccountInfo,
+    key_index: usize,
+) -> MarginfiResult<I80F48> {
+    let pool = load_stake_pool(bank_config, stake_pool_info, key_index)?;
```
IMEF-FEMI, line 190: Could we also decode and check that `last_update_epoch` is current here? So we don't use a stale exchange rate if the pool’s backing has changed

programs/marginfi/src/state/lst_stake_price.rs
```diff
@@ -0,0 +1,256 @@
...
+    bank_config: &BankConfig,
+    vault_info: &'info AccountInfo<'info>,
+    key_index: usize,
+) -> MarginfiResult<AccountLoader<'info, MinimalExponentVault>> {
+    require_keys_eq!(
+        *vault_info.key,
+        bank_config.oracle_keys[key_index],
+        MarginfiError::ExponentVaultValidationFailed
+    );
+    let loader: AccountLoader<MinimalExponentVault> = AccountLoader::try_from(vault_info)
```
IMEF-FEMI, line 239: Could we also validate that the Exponent vault's `mint_pt` matches the bank mint here? similar to what we have [here](https://github.com/0dotxyz/marginfi-v2/blob/78c2fff89510116fc12f4be79d13ea4d1d4502c2/programs/marginfi/src/state/lst_stake_price.rs#L203)  so we don't end up using a wrong vault with a different PT token / maturity / redemption fields.

programs/marginfi/src/state/lst_stake_price.rs
```diff
@@ -0,0 +1,256 @@
...
+    check!(
+        pool.pool_token_supply > 0,
+        MarginfiError::ZeroSupplyInStakePool
+    );
+
+    let rate = I80F48::from_num(pool.total_lamports)
+        .checked_div(I80F48::from_num(pool.pool_token_supply))
+        .ok_or_else(math_error!())?;
+    check!(
+        rate < I80F48::from_num(MAX_LST_SOL_RATE),
```
IMEF-FEMI: Nit: We need to reject zero rates here too 
```suggestion
        rate > I80F48::ZERO && rate < I80F48::from_num(MAX_LST_SOL_RATE),
```

IMEF-FEMI (commented):
Looks good to me... just a few suggestions.

IMEF-FEMI (commented):
can we also drop the [`temp/`](https://github.com/mrgnlabs/marginfi-v2/blob/new-oracles/temp) files seems they were mistakenly pushed

programs/marginfi/src/state/lst_stake_price.rs
IliaZyrin: Very good catch, thank you! @k0beLeenders you might want to also include it on the swb/frontend side.

IMEF-FEMI (approved):
LGTM

dawid-sw (commented):
hardening idea @IliaZyrin 

configure_bank_oracle never resets fixed_price while the hardening now compares it in the same-asset identity, a bank that was ever set Fixed/PT keeps a leftover value and silently fails to pair with otherwise identical banks after switching back to Pyth. We can fix it by zeroing the field in that instruction.

dawid-sw (approved):

PR #633, opened 2026-07-27
SVSP phantom tokens accounting

IliaZyrin:
The on-ramp path divided pool NAV by raw LST supply, overvaluing low-supply staked collateral. SVSP prices every deposit/withdraw against a notional supply of [raw + PHANTOM_TOKEN_AMOUNT (fixed 1 SOL)](https://github.com/solana-program/single-pool/blob/main/program/src/processor.rs#L291-L301). We now divide by the same, so the multiplier equals the redeemable rate.


jgur-psyops (approved):
LGTM, will merge to include this in 1.10

dawid-sw (approved):
lgtm

PR #636, opened 2026-07-31
Preserve the elapsed frozen span when a circuit breaker reset clears the halt record

IMEF-FEMI:
## Summary 
reset_cb_runtime_state now carries a halt's elapsed frozen seconds into cb_frozen_seconds_pending instead of dropping them, so the L-02 interest freeze holds even if a caller resets without accruing first.

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_group/clear_circuit_breaker.rs
+++ b/programs/marginfi/src/instructions/marginfi_group/clear_circuit_breaker.rs
@@ -36,7 +36,7 @@ pub fn lending_pool_clear_circuit_breaker(
         ctx.accounts.bank.key(),
     )?;
     let prior_tier = bank.cb_tier;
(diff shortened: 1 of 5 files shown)


jgur-psyops:
One thing to be aware of with this change is there is economic tradeoff now. Consider asset A, which is in price free-fall because it got hacked or something

* Borrowers of A don't want to repay because they know it's going down rapidly.
* Lenders of A want to exit but can't, because borrowers won't repay to free up liquidity.

The usual market incentive is:
* Lenders all withdraw up to max utilization
* Borrowers pay massive interest, which encourages some of them to unwind

But here:
* Breaker trips
* Borrowers pay nothing until breaker resets

I think this is a worthwhile tradeoff, the breaker reset time is relatively brief and even very high interest is trivial over those timeframes.

PR #638, opened 2026-08-01
Feat/position transfer

### LINKED ISSUE #628: Position Transfer Instruction

## Summary

It’s annoying to have compose a transfer of some position between two accounts as a withdraw/deposit or repay/borrow. If risk allows (e.g. health check passes in both accounts), permit a user to transfer an asset position from one account to another. We probably will not check that the user owns both accounts, which would enable you to directly send positions to people (this would be especially useful for our emissions). You should be able to send 

Moving debts between accounts with the same authority (or both authorities sign?) would also be nice, but can be reserved for a separate PR.

Some caveats:
* We will charge a small (probably configurable on global fee state) flat fee in SOL to participate to discourage spam
* The minimum transfer amount should be around $1 (probably configurable on global fee state), again to discourage spam
* User accounts should be able to opt out of this feature entirely (for integrators that have delicate account management), thought opted-in will likely be the default (including for legacy accounts)

ashleychandy:
## Description

Implement position transfer support, allowing users to transfer asset positions between Marginfi accounts while enforcing health checks, configurable protocol fees, minimum transfer value, and per-account opt-out controls.

## Changes

- Add `lending_account_transfer_position` instruction for transferring asset positions between accounts.
- Enforce source and destination health checks, bank state validation, and transfer amount validation.
- Add configurable position transfer fee and minimum USD transfer value with default fallbacks.
- Add account-level send/receive opt-out flags and instruction to manage them.
- Add position transfer events and error types.
- Extend fee state initialization, editing, and propagation to include position transfer configuration.
- Add comprehensive tests covering successful transfers, failure cases, flags, limits, and invariants.

Fixes #628


IMEF-FEMI:
Thanks @ashleychandy for kicking this off. Carried forward as #683, so closing this one. Further updates will be done there.

PR #639, opened 2026-08-02
prohibit durable nonce in admin instructions

### LINKED ISSUE #637: Prohibit durable nonce in all admin ixes

## Summary

Anything that changes bank settings, adds banks, etc should prohibit durable nonce. We have no genuine use-case for this feature and it only opens surface area for attacks.

User-only ixes (like deposit, withdraw, etc) can continue to allow durable nonce: we are agonistic to workflows from integrators that may or may not depend on that feature.

ashleychandy:
## Description

Prevent durable nonce transactions from being used with privileged/admin instructions to reduce attack surface. User-facing instructions such as deposit, withdraw, borrow, and repay continue to allow durable nonce transactions.

## Changes

- Added `check_no_durable_nonce` helper in `ix_utils`.
- Added `DurableNonceNotAllowed` error.
- Added the Instructions sysvar account to protected admin instructions.
- Invoked durable nonce validation before executing admin operations.
- Applied the restriction across admin/configuration, bank management, fee management, risk admin, and protocol administration instructions.

Fixes #637

### DIFF
--- a/programs/marginfi/fuzz/Cargo.lock
+++ b/programs/marginfi/fuzz/Cargo.lock
@@ -430,7 +430,7 @@ dependencies = [
  "solana-define-syscall 3.0.0",
  "solana-feature-gate-interface",
  "solana-instruction",
- "solana-instructions-sysvar",
+ "solana-instructions-sysvar 3.0.1",
  "solana-invoke",
  "solana-loader-v3-interface",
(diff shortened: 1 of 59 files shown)


dawid-sw:
https://github.com/0dotxyz/marginfi-v2/pull/670 cloned

PR #640, opened 2026-08-02
add max borrow withdraw query

### LINKED ISSUE #627: Ping max available to borrow/withdraw

## Summary

One of the greatest frontend/integrator annoyances is that it’s actually quite involved to determine the amount available to borrow or withdraw at your current health level, essentially forcing the caller to fully reproduce on-chain health state including prices. There should be no-op instruction (or one each for borrow/withdraw) that takes a bank as an argument and performs a full risk engine check, then returns the max amount you would be able to borrow or withdraw. We have to provide this output in such a way that rpc simulate can easily grab it, most likely an event. When withdrawing, but mindful that if the allowable withdraw amount exceeds what the user actually has in that asset, it should say "all" in some way.

ashleychandy:
## Description

Adds a no-op instruction to query the maximum amount a marginfi account can borrow or withdraw from a given bank without reproducing the on-chain risk engine off-chain.

The instruction simulates borrow/withdraw operations using the existing risk checks and emits the result as an event, making it easy to consume via RPC simulation.

## Changes

- Add `lending_account_query_max_borrow_withdraw` instruction
- Add `QueryMaxBorrowWithdrawEvent` for RPC simulation output
- Implement binary search to determine the maximum borrow amount
- Implement binary search to determine the maximum withdraw amount
- Return `is_all` when the entire user balance can be withdrawn
- Reuse existing borrow/withdraw logic and health checks without mutating state
- Export the new instruction through the program entrypoint

Fixes #627


dawid-sw:
cloned as #671

PR #641, opened 2026-08-06
chore: update security file to include deprecated copy

borcherd:
Add supported assets and deprecated sections & copy


dawid-sw (dismissed):
lgtm

PR #644, opened 2026-08-08
Reject leverage and fee configs that would leave a position unliquidable

IMEF-FEMI:
## Summary

This PR prevents the configuring of  positions whose liquidator fee cannot be paid.

A liquidation only clears while the liquidatee's credit outweighs the seized collateral, which leaves `1 / leverage` of room for fees. A bank whose liquidator fee exceeds that room has its realized fee is silently degraded below the configured rate: at a 0.99 collateral weight a bank configured for 2.5% pays the liquidator 1.0%, and the insurance cut reaches zero. Nothing errors and nothing is logged. This PR makes that config rejected at write time instead.

The relationship is checked whenever either side moves: when e-mode entries are set, when a bank's liquidator fee changes, and when a bank opts into same-asset e-mode, whose weights are synthesized from the group's leverage rather than stored as entries.


type-crate/src/types/interest_rate.rs
```diff
@@ -137,6 +137,8 @@ pub fn basis_to_u32(value: I80F48) -> u32 {
     let max_value: I80F48 = I80F48::from_num(100.0); // 0-100 range
     let clamped: I80F48 = value.min(max_value).max(I80F48::ZERO);
     let ratio: I80F48 = clamped / max_value;
+    // Truncates, so an encoded value decodes a hair below itself. Callers comparing against a
+    // decoded bound must allow for that; `1` must keep decoding below 1 to read as disabled.
```
IliaZyrin: Nitpick: this reads a bit confusing, would be nice to simplify the language.

IliaZyrin (approved):

dawid-sw (approved):
looks good. raising one thing to remember is that `same_asset_emode_maint_leverage` via group configure doesn't re-validate opted-in banks against this check. but not realistic to do it on program level, so we need to remember to validate it off-chain before this param mod

PR #649, opened 2026-08-18
Rounding for minting and burning shares

dawid-sw:
Defense in depth for mint & burn operations.

### DIFF
--- a/programs/marginfi/src/state/marginfi_account.rs
+++ b/programs/marginfi/src/state/marginfi_account.rs
@@ -2200,7 +2200,16 @@ impl<'a> BankAccountWrapper<'a> {
         // liability to repay, a pure repay adds no assets). The amounts are `max(_, 0)`, so `> 0`
         // captures exactly the cases where `change_*_shares(0)` would have been a no-op.
         let asset_shares_increase = if asset_amount_increase > I80F48::ZERO {
+            // `get_asset_shares` truncates toward zero. At an inflated asset share value a
+            // positive token transfer can round down to zero raw share bits, recording no
+            // claim while taking the caller's tokens. Reject that the same way we reject a
+            // zero-burn withdrawal.
             let shares = bank.get_asset_shares(asset_amount_increase)?;
+            check!(
+                shares > I80F48::ZERO,
+                MarginfiError::IllegalBalanceState,
+                "Deposit would transfer assets without minting shares"
+            );
             balance.change_asset_shares(shares)?;
             bank.change_asset_shares(
                 shares,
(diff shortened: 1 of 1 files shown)


IliaZyrin (approved):
LGTM!

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -2306,7 +2315,19 @@ impl<'a> BankAccountWrapper<'a> {
...
         let asset_shares_decrease = if asset_amount_decrease > I80F48::ZERO {
-            let shares = bank.get_asset_shares(asset_amount_decrease)?;
+            // Round the burn UP. `get_asset_shares` truncates, which would surrender fewer shares
+            // than the tokens leaving the vault are worth; at a high `asset_share_value` that gap
+            // is a material quantity, not dust. Clamp to the balance so withdrawing the position's
+            // full value cannot over-burn by the added ULP.
+            let truncated = bank.get_asset_shares(asset_amount_decrease)?;
+            let shares = if bank.get_asset_amount(truncated)? < asset_amount_decrease {
+                min(
+                    I80F48::from_bits(truncated.to_bits() + 1),
```
IMEF-FEMI, line 2325: We should consider rounding up liability_shares_increase, since the truncating conversion can record liability shares worth less than the tokens transferred

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -2200,7 +2200,16 @@ impl<'a> BankAccountWrapper<'a> {
         // liability to repay, a pure repay adds no assets). The amounts are `max(_, 0)`, so `> 0`
         // captures exactly the cases where `change_*_shares(0)` would have been a no-op.
         let asset_shares_increase = if asset_amount_increase > I80F48::ZERO {
+            // `get_asset_shares` truncates toward zero. At an inflated asset share value a
```
jgur-psyops, line 2203: nit: I think we know the exact value where this starts to occur (2^48ish?) could be specific.

IMEF-FEMI (commented):
I also think it would be good to reject same-mint emissions when total_asset_shares < EMPTY_BALANCE_THRESHOLD as an additional hardening measure

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -2306,7 +2315,19 @@ impl<'a> BankAccountWrapper<'a> {
         // liability, a pure borrow removes no assets). The amounts are `max(_, 0)`, so `> 0`
         // captures exactly the cases where `change_*_shares(0)` would have been a no-op.
         let asset_shares_decrease = if asset_amount_decrease > I80F48::ZERO {
-            let shares = bank.get_asset_shares(asset_amount_decrease)?;
+            // Round the burn UP. `get_asset_shares` truncates, which would surrender fewer shares
+            // than the tokens leaving the vault are worth; at a high `asset_share_value` that gap
```
jgur-psyops, line 2319: likewise here, the doc implies this can happen at "high" asset share values which is spookier than the actual implication

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -2306,7 +2315,19 @@ impl<'a> BankAccountWrapper<'a> {
...
+            // is a material quantity, not dust. Clamp to the balance so withdrawing the position's
+            // full value cannot over-burn by the added ULP.
+            let truncated = bank.get_asset_shares(asset_amount_decrease)?;
+            let shares = if bank.get_asset_amount(truncated)? < asset_amount_decrease {
```
jgur-psyops, line 2323: unfortunate to use this many CU on this defense/edge case but I guess it is what it is. get_asset_amount is pretty expensive iirc.

programs/marginfi/src/state/marginfi_account.rs
```diff
@@ -3368,4 +3405,136 @@ mod test {
             );
         }
     }
+
+    /// A withdrawal must never move more token value out of the vault than the
+    /// value of the shares it burns. `get_asset_shares` truncates toward zero,
+    /// so the burn leg under-counts by up to one raw share bit. That bit is
+    /// worth `asset_share_value * 2^-48`, which is harmless at a normal share
+    /// value but becomes a material quantity once `asset_share_value` is
+    /// inflated (permissionless `lending_pool_emissions_deposit` can do this on
```
jgur-psyops, line 3414: I wonder if we should just disable permissionless emissions deposit under some condition. the easiest would probably be to disable if the asset share value is above something like 10, I think that would easily cover all production banks.

dawid-sw: yes might be good idea, I need to dig more. the downside is if we introduce another trick to increase shares besides emissions deposit, we need to remember to patch it as well

PR #654, opened 2026-08-25
Kamino emergency state handling

IliaZyrin:
This updates Kamino IDLs and introduces handling of the [emergency state](https://github.com/Kamino-Finance/klend/blob/release/v1.25.0/programs/klend/src/lending_market/lending_operations.rs#L67-L70) Kamino introduced for their banks.

If a bank is in this state, its collateral is worth **zero** in terms of (**initial**) borrowing power.


dawid-sw (approved):
LGTM

programs/kamino-mocks/src/state.rs
```diff
@@ -296,6 +298,13 @@ impl MinimalReserve {
         // in the same slot brings it current. Keepers reading a venue rate must refresh in-tx.
         self.slot < current_slot
     }
+
+    /// True if Kamino put this reserve into emergency mode (`ReserveConfig.emergency_mode`). When
+    /// set, `refresh_reserve` clears the price status:
+    /// https://github.com/Kamino-Finance/klend/blob/release/v1.25.0/programs/klend/src/lending_market/lending_operations.rs#L67-L70
+    pub fn is_emergency_mode(&self) -> bool {
+        self.config.emergency_mode != 0
```
jgur-psyops, line 306: I suppose we also care about the (apparently new?)  LendingMarket.emergency_mode as well?

My intuition is yes: that is their equivalent of a "pause" on the entire market and all of our txes involving will fail regardless, however they permit refreshing to continue, so users will continue to have borrowing power against paused collateral on our platform.

We likely don't want to have to pass the Market in every Kamino ix, so we would need some sort of propagation mechanism, which is probably beyond the scope of this PR, so I'll open an issue for it.

scripts/update_idls.sh
```diff
@@ -10,16 +10,20 @@ ANCHOR_PROVIDER_URL="${ANCHOR_PROVIDER_URL:-https://api.mainnet-beta.solana.com}
...
-  kamino_lending
-  kamino_farms
+# One row per program: name | program id | IDL source path. Every artifact is named after the
+# program, so `<name>.json`, `<name>.ts` and `<name>.so` all follow from the first field.
+#
+# Sources are vendor repos, not `anchor idl fetch`: Kamino and JupLend publish no IDL account, and
+# Drift's is a 3-instruction stub that would clobber the real file.
+RAW="https://raw.githubusercontent.com"
+
+PROGRAMS=(
```
jgur-psyops, line 20: Here we are assuming that Kamino/Drift/Jup are cool dudes like us where main branch = mainnet and not weird like svsp where main branch = coming to mainnet in 4 months. 

We fetch the .so from rpc but then assume the idl at master/main matches that, which is probably true, but not necessarily. 

Will note in the README that it's incumbent on the re-generator of the idl to actually check.

PR #667, opened 2026-09-02
Bank resize Instruction

IMEF-FEMI:
## Summary 
Adds an instruction to resize bank accounts with 2 KB of reserve space. 
### Why?
Banks currently have only about 8 reserved bytes remaining, making future per-bank features difficult to implement safely. Resizing them ahead of layout changes would enable us implement new features without downtime or breaking existing bank operations.


programs/marginfi/tests/admin_actions/bank_resize.rs
```diff
@@ -0,0 +1,136 @@
...
+            .iter()
+            .filter(|b| b.is_active())
+            .count(),
+        2
+    );
+    Ok(())
+}
+
+/// Grow-only: a bank already at the target is rejected. Raising `BANK_ACCOUNT_LEN` in a later
+/// release deliberately makes this callable again, so a reserve top-up needs no new instruction.
```
IliaZyrin, line 100: Not sure I got this one - do you plan to keep this instruction in the next release as well? I thought it's one-time only.

IliaZyrin (approved):

programs/marginfi/tests/admin_actions/bank_resize.rs
IMEF-FEMI: we could keep it so if we every have need to grow the account size again" we just update the BANK_ACCOUNT_LEN and call again. the test is so nobody can spam it and grow accounts when there's no need

type-crate/src/constants.rs
```diff
@@ -107,6 +108,20 @@ pub const REBALANCE_DEFAULT_COOLDOWN_SECONDS: u64 = 86_400;
...
+pub const BANK_RESERVED_BYTES: usize = 1024;
+
+/// Byte length `lending_pool_resize_bank_account` grows a bank account to, discriminator included.
+///
+/// This is deliberately NOT `8 + Bank::LEN`. Banks are resized ahead of any struct growth so that
+/// growth needs no migration and causes no downtime: a program whose `Bank` is smaller than the
+/// account simply ignores the tail, whereas a program whose `Bank` is LARGER than the account
+/// fails every instruction that loads a bank. Growing the struct and resizing in one release would
+/// take the protocol down for as long as the migration takes.
+pub const BANK_ACCOUNT_LEN: usize = 8 + Bank::V1_LEN + BANK_RESERVED_BYTES;
```
dawid-sw, line 123: Making struct smaller than account size would break a lot of integrators. Some deserialization processes are checking the struct size and account size and these need to match. I'd just put reserved on account struct. also its going to be much less confusing

dawid-sw (changes requested):
some comments

type-crate/src/constants.rs
jgur-psyops: I think Dawid is probably right here as we discussed: we were able to do the V1->V2 upgrade for group/fee state within 5 minutes of the program update, I think we can do the same with the Bank upgrade and not worry about bricking Banks for a minute or so.

Here currently we do avoid any bricking but have that awkward situation of V1 banks that are actually V2 banks which larger than the V1 struct, and although anchor ignores the extra space it is confusing for anyone loading the raw bytes or relying on length to fetch accounts (which really they shouldn't be, but hey people do it).

programs/marginfi/src/state/marginfi_group.rs
```diff
@@ -377,6 +377,7 @@ mod tests {
         assert_eq!(offset_of!(Bank, premium_tag), 1840);
         assert_eq!(offset_of!(Bank, _pad3), 1842);
         assert_eq!(offset_of!(Bank, premium_activated_at), 1848);
+        assert_eq!(offset_of!(Bank, _padding_1), Bank::V1_LEN);
```
dawid-sw, line 380: this is kinda random, we doubled size of bank, so this accidentally pass, but its would be more clear to me to just pass number here. but its just very very small nit

dawid-sw (changes requested):
[BankImpl::new](https://github.com/0dotxyz/marginfi-v2/blob/bank-account-resize/programs/marginfi/src/state/bank.rs#L163) now breaks on-chain because it builds the 3904-byte Bank by value on the stack, pushing its frame to 4928 bytes against the SBF limit of 4096 (Stack offset of 4880 exceeded max offset of 4096 by 784 bytes), so any call would trap with an access violation - it has no callers since #567 replaced it with in-place init, so delete it rather than leave the trap. Or as alternative do  `#[cfg(not(target_os = "solana"))]` or refactor to allow usage of this function, similar as Group was handled

programs/marginfi/src/state/marginfi_group.rs
IMEF-FEMI: This is actually intentional. `_padding_1` represents where the extra space starts and `V1_LEN` represents where the old size stopped, following the same pattern we have in the [group](https://github.com/0dotxyz/marginfi-v2/blob/e6ad58ffb3f39aebd8bf4c7677e58c716de1e721/programs/marginfi/src/state/marginfi_group.rs#L333) and [fee state](https://github.com/0dotxyz/marginfi-v2/blob/e6ad58ffb3f39aebd8bf4c7677e58c716de1e721/programs/marginfi/src/state/fee_state.rs#L109-L110) tests (just for consistency)

jgur-psyops:
> [BankImpl::new](https://github.com/0dotxyz/marginfi-v2/blob/bank-account-resize/programs/marginfi/src/state/bank.rs#L163) now breaks on-chain because it builds the 3904-byte Bank by value on the stack, pushing its frame to 4928 bytes against the SBF limit of 4096 (Stack offset of 4880 exceeded max offset of 4096 by 784 bytes), so any call would trap with an access violation - it has no callers since #567 replaced it with in-place init, so delete it rather than leave the trap. Or as alternative do `#[cfg(not(target_os = "solana"))]` or refactor to allow usage of this function, similar as Group was handled

Removed the Bank's new() impl since we no longer call it. 

Separately but related, the `ArchiveRecord` parse is too large for SBF. We also don't call it, so type-crate now builds a stub if `#[cfg(target_os = "solana")]`, this can be our alternate approach when we do need these kinds of functions for some off-chain reason (though I agree that for Bank's new(), we simply didn't need it).

dawid-sw (approved):

PR #673, opened 2026-09-09
Same asset auto-rebalance fixes

IMEF-FEMI:
Fxes:

  1. Native bank rates are computed from live assets and liabilities at the post-deposit utilization, replacing the cached pre-deposit rate in the start-side improvement and best-venue checks
  2. Rates are evaluated in aggregate: each destination at its total declared inflow, and each competing bank at its own inflow plus the move, so chunked or consolidated moves cannot pass a best-venue check their total would fail
  3. A bank may not be both a source and a destination within one execution, rejected with the new `RebalanceBankSourceAndDestination` error
  4. `admin_close_account` requires the account's rebalance fee pool to be empty, the same guard as `close_account`, so keeper-tip SOL cannot be orphaned by a permissionless close
  5. `settle_rebalance_tip` accepts a closed marginfi account, reading it only while it still exists, so an escrowed tip settles after the account is closed
  6. Every withdraw/deposit leg in a rebalance sandwich must act on a bank in the order's allowlist, rejected with the new `RebalanceForeignBankLeg` error, so a keeper cannot touch the account's other balances inside the sandwich
  7. The rebalance and order-execution records compare each untouched balance's order tag at end alongside its side and shares, so a balance closed and reopened with its tag cleared is rejected
  8. An order-tagged balance moves whole and alone into an empty bank and its tag is carried to the destination, rejected otherwise with the new `RebalanceTaggedBalanceSplit` error, so a stop-loss keeps guarding collateral that a rebalance relocates
  9. Both execution records skip an active balance whose shares fall below `EMPTY_BALANCE_THRESHOLD`, so a leftover empty slot no longer blocks `start_rebalance` or `start_execute_order`
  10. `end_rebalance` and `end_execute_order` stamp the health cache they write with the current clock, and the order path also records the program version and engine flag, so the stored maintenance health reads as fresh instead of uninitialized
  11. JupLend deposits are previewed at the borrow rate the mint's `RateModel` pays at the post-deposit utilization, read from the model account now passed with each JupLend bank, so the best-venue ranking no longer relies on the reserve's stale stored rate

### DIFF
--- a/programs/marginfi/src/instructions/marginfi_account/rebalance.rs
+++ b/programs/marginfi/src/instructions/marginfi_account/rebalance.rs
@@ -5,13 +5,15 @@
 //! consumed on execution; it persists until cancelled.
 //!
 //! On-chain guarantees: every referenced bank holds the order's mint and is in the allowed set; each
-//! declared move goes from a lower-rate bank to one beating it by `min_improvement` (pre-move) and
-//! not inverted after the move's own market impact (post-move); no move passes over a higher-rate
-//! referenced bank that still has deposit capacity, measured as the tighter of the bank's own limit
-//! and its venue's; the total tokens moved are capped by the order's `amount` budget (uncapped when
-//! the order is unlimited); token principal is conserved per bank up to a small dust tolerance; the
-//! non-referenced balance set is unchanged, neither altered nor added to; the account stays healthy
-//! at the maintenance requirement if it borrows; and a per-order cooldown.
+//! move's destination, priced with every declared deposit into it counted, beats the move's source
+//! by `min_improvement` before the legs run and still does after they land; no other referenced
+//! bank with deposit capacity, measured as the tighter of the bank's own limit and its venue's,
+//! would pay the move's tokens more, counting that bank's own declared inflow; a bank is either a
+//! source or a destination within one execution; the total tokens moved are capped by the order's
+//! `amount` budget (uncapped when the order is unlimited); token principal is conserved per bank
+//! up to a small dust tolerance; the non-referenced balance set is unchanged, neither altered nor
+//! added to; the account stays healthy at the maintenance requirement if it borrows; and a
+//! per-order cooldown.
 //!
 //! Supports native, Kamino, Drift, and JupLend legs; Solend banks are rate-visible but have no move
 //! legs and are rejected up front. Referenced banks arrive as a deduped, indexed stream in the
@@ -53,7 +55,7 @@ use crate::{
         marginfi_group::MarginfiGroupImpl,
         premium::{MarginfiAccountPremiumImpl, PremiumScratch},
         price::OraclePriceFeedAdapter,
-        rate::{self, rate_at, rate_of, RewardsAccounts},
+        rate::{self, rate_at, rate_of, NativeRateModel, RewardsAccounts},
         rebalance::{RebalanceOrderImpl, RebalanceRecordImpl},
     },
     utils::is_integration_asset_tag,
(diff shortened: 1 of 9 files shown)


programs/marginfi/src/state/interest_rate.rs
```diff
@@ -325,6 +312,75 @@ impl InterestRateCalc {
...
+        }
+        prev_util = point_util;
+        prev_rate = point_rate;
+    }
+
+    InterestRateCalc::lerp(prev_util, prev_rate, I80F48::ONE, hundred_rate, ur)
+}
+
+/// A bank's seven-point interest curve with its points converted to I80F48 once, for repeated
+/// lookups at different utilizations.
```
IliaZyrin, line 341: Just wondering how heavily this affected a typical rebalance CUs? If we measured the performance somewhere.
Starts to look very math-intense.

IliaZyrin (approved):
LGTM!

programs/juplend-mocks/src/state.rs
```diff
@@ -321,6 +326,83 @@ impl TokenReserve {
     }
 }
 
+const _: () = assert!(core::mem::size_of::<RateModel>() == 45);
+
+/// Minimal mirror of JupLend's liquidity-layer `RateModel`, the per-mint borrow-rate curve that
+/// `operate` re-prices a reserve on after every supply or borrow change.
+/// https://github.com/Instadapp/fluid-solana-programs/blob/master/programs/liquidity/src/state/rate_model.rs
+#[zero_copy]
+#[repr(C, packed)]
```
jgur-psyops, line 335: nit: Not a fan of the repr packed here, I'd prefer if we follow the system other structs use where we use plain repr c and manually demonstrate where the padding is placed.  in loading, `try_pod_read_unaligned` should not be necessary in that case.

But, we've used packed a lot lately, so it's fine for now. At some point should go back and clean those up, but is what it is.

programs/marginfi/src/state/rate.rs
```diff
@@ -359,28 +405,61 @@ fn juplend_supply_apr<'info>(
     Ok(base.checked_add(rewards_apr).ok_or_else(math_error!())?)
 }
 
+/// JupLend's `RateModel` for `mint`: the liquidity program keys one per mint, so an account it owns
```
jgur-psyops, line 408: I think this invariant is currently true, but I guess it may not always be true. Can derive the RateModel PDA and compare it, or is that too cumbersome? If not, nbd, checking the mint is likely sufficient.

programs/marginfi/src/instructions/marginfi_account/rebalance.rs
```diff
@@ -821,60 +843,112 @@ pub fn start_rebalance<'info>(
...
             MarginfiError::RebalanceNotImproving
         );
-        // Banks this execution has already filled to their deposit capacity are skipped; no other
-        // bank may beat the destination at the same deposit amount.
+        let amount_native = to_native(mint_decimals, I80F48::from(m.amount))?;
+        // Skip banks this execution already fills to capacity, and native banks already at or below
+        // the destination's rate with their own inflow (more deposits only lower a native rate).
         for i in 0..banks.len() {
-            if i == d || inflow[i] >= capacity[i] {
+            if i == d || inflow[i] >= capacity[i] || (models[i].is_some() && landed[i] <= landed[d])
```
jgur-psyops, line 943: re: SA-01, I think this is closer, but consider: a candidate is priced at inflow[i] + amount_native, so if inflow[i] < capacity[i], where inflow[i] + current_move > capacity[i]....

then the candidate cannot accept the whole move and should be skipped. It is nonetheless compared, which leads to this scenario:

* 1 token currently in A pending move
* 1 capacity in C, which has already received 0.5 elsewhere in the batch
* An attempt to move into 1 token into D being evaluated

The hypothetical rate of C is evaluated at 1.5 tokens, which C cannot support, and if hypothetical C > D then the move is rejected.

tl;dr we need some more aggressive candidate skipping here, but this is somewhat a nit since it only applies at edge cases and generally leads only to more conservative results (as far as I can tell).

PR #677, opened 2026-09-16
Add Scope oracles for Juplend and Kamino

dawid-sw:
# Add ScopeKamino and ScopeJuplend oracle setups

New `OracleSetup::ScopeKamino` (27) and `ScopeJuplend` (28): a Scope feed entry priced through the Kamino reserve or JupLend lending exchange rate in `oracle_keys[1]`.

`lending_pool_configure_bank_oracle_scope` picks the setup from the bank's asset tag and validates the venue account passed after the feed; the venue key stays as set at bank creation. Bank creation and the generic configure instruction reject these setups.

Tests: unit tests for validation and feed family, `k21_scopeOracle` and `jlr13_scopeOracle` specs, shared Scope feed mock in `tests/utils/scope-utils.ts`.

The goal is to have failover for Pyth Core feeds, so we are able to use Pyth Lazer or even Chainlink powered feeds.


IliaZyrin (approved):
LGTM

jgur-psyops (approved):
LGTM. 

We did not implement the admin-configurable confidence for Scope like we recently did with SWB. I would call it very low priority since we have yet to use the feature at all for SWB, but I will add it to issues as low-priority for later.

PR #683, opened 2026-09-19
Position transfer internal

IMEF-FEMI:
Closes #628. thanks @ashleychandy for starting this in #638. Rebased onto 0.1.12-main and made a few updates


programs/marginfi/src/instructions/marginfi_account/set_position_transfer_flags.rs
```diff
@@ -0,0 +1,57 @@
...
+};
+use anchor_lang::prelude::*;
+use marginfi_type_crate::types::{
+    MarginfiAccount, ACCOUNT_POSITION_TRANSFER_RECEIVE_DISABLED,
+    ACCOUNT_POSITION_TRANSFER_SEND_DISABLED,
+};
+
+pub fn set_position_transfer_flags(
+    ctx: Context<SetPositionTransferFlags>,
+    disable_send: Option<bool>,
```
IliaZyrin: This is probably over-engineered. In which case would someone want to disable send for themselves? The transfer ix anyway requires them to sign.

Also, if the account is frozen, the transfer ix can be run by the admin, but if this flag is set (`ACCOUNT_POSITION_TRANSFER_SEND_DISABLED`), admin cannot unset it, and it blocks the transfer. So if we still wanna keep these flags, let's allow admin to sign this too (at least if the acc is frozen).

Regarding disable receive, I can see that someone wants to be protected from some spammy deposits occupying all remaining balances and thus griefing the acc, but it's easily recoverable and also not worth a whole ix, imho.

tests/specs/basic/25_transferPosition.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+    assert.equal(await lamports(users[1].wallet.publicKey), receiverBefore);
+  });
+
+  it("(user 0 -> user 0) rejects a transfer to the same account", async () => {
+    const result = await transfer(users[0], users[0], collateralBank, lst);
+    // PositionTransferIdenticalAccounts
+    assertBankrunTxFailed(result, 6904);
+  });
+
+  it("(user 0 -> user 2) rejects a frozen destination", async () => {
```
IliaZyrin, line 213: Can also add a case for frozen source.

programs/marginfi/src/instructions/marginfi_account/set_position_transfer_flags.rs
IMEF-FEMI: fixed

